PR #53, opened 2014-10-29
moved string.Format job into the formatter for WriteWarning extension method

tugberkugurlu:
I believe that `string.Format` shouldn't be called directly as `WriteWarning` method receives the message and args parameters. The warning logging could be disabled and `WriteWarning` would end up calling `string.Format` for no good.

I'm also not sure about my implementation here :smile: I used `Tuple` and it felt dirty :smile:

### DIFF
--- a/src/Microsoft.Framework.Logging.Interfaces/LoggerExtensions.cs
+++ b/src/Microsoft.Framework.Logging.Interfaces/LoggerExtensions.cs
@@ -13,9 +13,7 @@ namespace Microsoft.Framework.Logging
     /// </summary>
     public static class LoggerExtensions
     {
-        private static readonly Func<object, Exception, string> TheMessage = (message, error) => (string)message;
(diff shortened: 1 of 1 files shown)


muratg:
@tugberkugurlu Thanks for the PR. Could you rebase this one and send an update?

tugberkugurlu:
@muratg rebased but as you can guess from my commit message :smile: I am not sure if these changes make sense any more as the intention of this PR was already handled by other changes.

However, the changes still made the code consistent if you as me :smile:

muratg:
@kichalla

kichalla:
Thanks @tugberkugurlu The logger extensions class has gone through several code changes since this PR was submitted and i think your concern mentioned here has been addressed. Please take a look at the following again and submit  a new PR if required: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs

tugberkugurlu:
:+1:

Issue #89, opened 2015-01-12
ILogger.Write to support taking in all types


Praburaj:
Today when I want to write an `enum` or an `int` to the logger stream using `ILogger.WriteInformation` etc, I have to do a `ToString()` which seems to be painful for most common scenarios. It will be nice to have it something like `Console.WriteLine()` where we can pass in anytype. By default can we do a ToString() internally?

Praburaj:
Fixed.

PR #92, opened 2015-01-13
Fixing the logger extensions to take in params object[] instead of strin...

Praburaj:
...g[]

Fixes bug # https://github.com/aspnet/Logging/issues/89

Right now the logger extensions take in params string[] as args for substitution values. This makes it painful to send basic CLR types and user code has to do a .ToString() on even basic CLR types.

With this change it will be possible to possible the object as it is.

/CC @Tratcher


Eilon:
Looks reasonable to me, :shipit: I wonder why it wasn't done that way to begin with...

Eilon:
Looks like it started that way, perhaps Sonja wasn't aware of the `object` way of doing it.

Praburaj:
Probably I guess the same. In fact the XML docs for this param seems to be right, but just the `params` being taken in is `string[]`.

PR #98, opened 2015-01-29
Feature structuredformatstring

Praburaj:
This adds ability to specify named placeholders while writing logs. 

For example: 
`logger.WriteInformation("Tests started at '{StartTime}'", DateTime.Now)` (instead of `'{0}'` => `'{StartTime}'`). 

@lodejard

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -38,8 +38,8 @@ namespace SampleApp
                 .WriteTo.Sink(new RollingFileSink("file-{Date}.json", new JsonFormatter(), null, null))
                 .WriteTo.Sink(new FileSink("dump.txt", new RawFormatter(), null)));
 #endif
-            //factory.AddConsole();
-            //factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
+            factory.AddConsole();
+            factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
         }
 
         public void Main(string[] args)
@@ -77,8 +77,8 @@ namespace SampleApp
             _logger.WriteInformation("Stopping");
 
             _logger.WriteInformation(Environment.NewLine);
-            _logger.WriteInformation("{2,-10}{0,15}{1,15}{3,15}", "RESULT", "START TIME", "END TIME", "DURATION(ms)");
-            _logger.WriteInformation("{2,-10}{0,15}{1,15}{3,15}", "------", "----- ----", "--- ----", "------------");
+            _logger.WriteInformation("{Result,-10}{StartTime,15}{EndTime,15}{Duration,15}", "RESULT", "START TIME", "END TIME", "DURATION(ms)");
+            _logger.WriteInformation("{Result,-10}{StartTime,15}{EndTime,15}{Duration,15}", "------", "----- ----", "--- ----", "------------");
             _logger.WriteInformation("{Result,-10}{StartTime,15:mm:s tt}{EndTime,15:mm:s tt}{Duration,15}", "SUCCESS", startTime, endTime, (endTime - startTime).TotalMilliseconds);
         }
     }
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -41,12 +41,13 @@ namespace SampleApp
             //factory.AddConsole();
             //factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
         }
-        
+
         public void Main(string[] args)
         {
             _logger.WriteInformation("Starting");
 
-            _logger.WriteInformation(1, "Started at '{StartTime}' and 0x{Hello:X} is hex of 42", DateTimeOffset.UtcNow, 42);
+            var startTime = DateTimeOffset.UtcNow;
+            _logger.WriteInformation(1, "Started at '{StartTime}' and 0x{Hello:X} is hex of 42", startTime, 42);
 
             try
             {
@@ -70,10 +71,15 @@ namespace SampleApp
                 Console.ReadLine();
             }
 
-            _logger.WriteInformation(2, "Stopping at '{StopTime}'", DateTimeOffset.UtcNow);
+            var endTime = DateTimeOffset.UtcNow;
+            _logger.WriteInformation(2, "Stopping at '{StopTime}'", endTime);
 
             _logger.WriteInformation("Stopping");
 
+            _logger.WriteInformation(Environment.NewLine);
+            _logger.WriteInformation("{2,-10}{0,15}{1,15}{3,15}", "RESULT", "START TIME", "END TIME", "DURATION(ms)");
+            _logger.WriteInformation("{2,-10}{0,15}{1,15}{3,15}", "------", "----- ----", "--- ----", "------------");
+            _logger.WriteInformation("{Result,-10}{StartTime,15:mm:s tt}{EndTime,15:mm:s tt}{Duration,15}", "SUCCESS", startTime, endTime, (endTime - startTime).TotalMilliseconds);
         }
     }
 }
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -1,6 +1,15 @@
 ﻿using System;
+using System.Collections.Generic;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Logging.Console;
+using ILogger = Microsoft.Framework.Logging.ILogger;
+#if !ASPNETCORE50
+using Serilog;
+using Serilog.Sinks.IOFile;
+using Serilog.Formatting.Raw;
+using Serilog.Sinks.RollingFile;
+using Serilog.Formatting.Json;
+#endif
 
 namespace SampleApp
 {
@@ -19,14 +28,25 @@ namespace SampleApp
             // providers may be added to an ILoggerFactory at any time, existing ILoggers are updated
 #if !ASPNETCORE50
             factory.AddNLog(new global::NLog.LogFactory());
-#endif
-            factory.AddConsole();
-            factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
-        }
 
+            factory.AddSerilog(new Serilog.LoggerConfiguration()
+                .Enrich.WithMachineName()
+                .Enrich.WithProcessId()
+                .Enrich.WithThreadId()
+                .MinimumLevel.Debug()
+                .WriteTo.RollingFile("file-{Date}.txt", outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} {Level}:{EventId} [{SourceContext}] {Message}{NewLine}{Exception}")
+                .WriteTo.Sink(new RollingFileSink("file-{Date}.json", new JsonFormatter(), null, null))
+                .WriteTo.Sink(new FileSink("dump.txt", new RawFormatter(), null)));
+#endif
+            //factory.AddConsole();
+            //factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
+        }
+        
         public void Main(string[] args)
         {
             _logger.WriteInformation("Starting");
+
+            _logger.WriteInformation(1, "Started at '{StartTime}' and 0x{Hello:X} is hex of 42", DateTimeOffset.UtcNow, 42);
 
             try
             {
@@ -49,6 +69,8 @@ namespace SampleApp
                 _logger.WriteInformation("Waiting for user input");
                 Console.ReadLine();
             }
+
+            _logger.WriteInformation(2, "Stopping at '{StopTime}'", DateTimeOffset.UtcNow);
 
             _logger.WriteInformation("Stopping");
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,108 @@
...
+            public Formatter(string format)
+            {
+                var sb = new StringBuilder();
+                var endIndex = format.Length;
+                for (var scanIndex = 0; scanIndex != endIndex;)
+                {
+                    var openBraceIndex = FindIndexOf(format, '{', scanIndex, endIndex);
+                    var closeBraceIndex = FindIndexOf(format, '}', openBraceIndex, endIndex);
+
+                    // Format item syntax : { index[,alignment][ :formatString] }.
```
Praburaj: @lodejard I had to update this as the format item syntax is per the comment here.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    public class LoggerStructureFormat : ILoggerStructure
+    {
+        private static ConcurrentDictionary<string, Formatter> _formatters = new ConcurrentDictionary<string, Formatter>();
+        private readonly Formatter _formatter;
+        private readonly object[] _values;
+
+        public LoggerStructureFormat(string format, params object[] values)
+        {
+            if (!_formatters.TryGetValue(format, out _formatter))
```
Eilon: Why not just always call `GetOrAdd` with a `Func`?

Eilon: This method: https://msdn.microsoft.com/en-us/library/ee378677(v=vs.110).aspx

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,105 @@
...
+        class Formatter
+        {
+            private readonly string _format;
+            private readonly List<string> _valueNames = new List<string>();
+
+            public Formatter(string format)
+            {
+                var sb = new StringBuilder();
+                var endIndex = format.Length;
+                for (var scanIndex = 0; scanIndex != endIndex;)
```
Eilon: Looks like a `while` loop to me.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return _formatter.GetValues(_values);
+        }
+
+        private Formatter ParseFormatter(string format)
+        {
+            return new Formatter(format);
+        }
+
+        class Formatter
```
Eilon: Fix coding style throughout.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,105 @@
...
+        {
+            return _formatter.Format(_values);
+        }
+
+        public IEnumerable<KeyValuePair<string, object>> GetValues()
+        {
+            return _formatter.GetValues(_values);
+        }
+
+        private Formatter ParseFormatter(string format)
```
Eilon: Is this method needed?

Praburaj: Can be removed. I updated.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: yeah..

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: Good point.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: I decorated all methods with `public/private/internal` - are there other specific comments?

samples/SampleApp/Program.cs
```diff
@@ -19,15 +28,27 @@ public Program()
...
+                .Enrich.WithProcessId()
+                .Enrich.WithThreadId()
+                .MinimumLevel.Debug()
+                .WriteTo.RollingFile("file-{Date}.txt", outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} {Level}:{EventId} [{SourceContext}] {Message}{NewLine}{Exception}")
+                .WriteTo.Sink(new RollingFileSink("file-{Date}.json", new JsonFormatter(), null, null))
+                .WriteTo.Sink(new FileSink("dump.txt", new RawFormatter(), null)));
 #endif
-            factory.AddConsole();
-            factory.AddConsole((category, logLevel) => logLevel >= LogLevel.Critical && category.Equals(typeof(Program).FullName));
+            //factory.AddConsole();
```
Eilon: Why are these commented? Perhaps I'm confused about the scenario for this but I thought it was to enable structure log formatting to the built-in loggers, such as the Console logger?

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    public class LoggerStructureFormat : ILoggerStructure
+    {
+        private static ConcurrentDictionary<string, Formatter> _formatters = new ConcurrentDictionary<string, Formatter>();
+        private readonly Formatter _formatter;
+        private readonly object[] _values;
+
+        public LoggerStructureFormat(string format, object[] values)
+        {
+            _formatter = _formatters.GetOrAdd(format, _ => new Formatter(format));
```
Eilon: Curious - can you pass `format` in as the parameter to the lambda? Otherwise it looks like it's putting `format` in the closure, which I don't think is intentional (or needed) here. Not 100% sure about this though.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,99 @@
...
+                        sb.Append(format, formatDelimiterIndex, closeBraceIndex - formatDelimiterIndex + 1);
+
+                        scanIndex = closeBraceIndex + 1;
+                    }
+                }
+
+                _format = sb.ToString();
+            }
+
+            private int FindIndexOf(string format, char ch, int startIndex, int endIndex)
```
Eilon: Can make `static`?

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,99 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+
+namespace Microsoft.Framework.Logging
+{
+    public class LoggerStructureFormat : ILoggerStructure
```
Eilon: Maybe add a doc comment to this?

samples/SampleApp/Program.cs
Praburaj: This shouldn't really be commented. I'll uncomment this. I just took over code and missed to notice this.

Praburaj: This feature adds ability to specify named placeholders as I mentioned in the PR description.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: We can't pass `format` as a parameter to the lambda.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: yeah. sure.

Praburaj:
Updated PR.

avanderhoorn:
Really like this PR!!!! Its been on our wish list for ever but never thought it would make it into the core platform. 

We are wanting to do things where we visually tokenise the trace messages we are showing and describe to people what they are seeing. With the options we have in `LoggerStructureFormat` currently, we probably can't do this however, as we only have access to the finial "formatted" or the list of key/value tokens. This isn't enough on its own for us to identify the tokens. Any chance we can have access to the underlying original format string as well? 

This would mean that when we visually show things, we can locate the correct places for highlighting.

Praburaj:
@avanderhoorn is the ask to just expose the original format string as a property on `LoggerStructureFormat`?

avanderhoorn:
Yes. I think thats all I would need to identify the tokens and build what I need. 

If i was making wishes to santa however, instead of me need to do that manually, it would be amazing if we could get the positions of the tokens... as at the moment, if we want to show the visualized version of the tokenized string, we have to recreate the logic in `LoggerStructureFormat` in order to put the `span`s around each token, etc. Does that make sense?

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            public Formatter(string format)
+            {
+                var sb = new StringBuilder();
+                var scanIndex = 0;
+                var endIndex = format.Length;
+
+                while (scanIndex < endIndex)
+                {
+                    var openBraceIndex = FindIndexOf(format, '{', scanIndex, endIndex);
```
Eilon: I think we need to consider how to escape curly braces. Currently I think it's broken if you try to put an arbitrary curly brace in the format. The typical way to escape curly braces is to double them, i.e. `{{` and `}}`. That's pretty much what `string.Format` does and also ASP.NET Routing.

lodejard: Good catch!

lodejard:
@avanderhoorn I like the idea of having the format string as a property, though not all messages which have structure will have a format string of course. How about if the code that parses and renders the message was public and exposed the info you described?

avanderhoorn:
Makes sense on the former (and I agree) and the latter would be great!!!
I'm trying to get to be point were it is as simple as possible to work out
what portions of the string where tokenized and where those tokens are (so
I can do the text highlighting) - showing that info would be a great step
in that direction.

src/Microsoft.Framework.Logging/LoggerStructureFormat.cs
Praburaj: I will do a bit more testing on this and update the PR.

Praburaj:
@Eilon @loudej @avanderhoorn - I have updated the PR addressing all the issues and adding more tests to cover the brace escape cases. 

@avanderhoorn I'm exposing the original format string as an item from the values returned from `GetValues()` with a special key named `__OriginalFormat`. I'm open to feedback on this.

avanderhoorn:
Thanks for getting this data out! In terms of feedback, I must be honest, accessing it via that key seems a little unintuitive and more like a private api... Would be nice if there was some more supported way of doing this.

test/Microsoft.Framework.Logging.Test/LoggerStructureFormatTest.cs
```diff
@@ -0,0 +1,61 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Microsoft.Framework.Logging.Test
+{
+    public class LoggerStructureFormatTest
```
Eilon: Can you add tests for un-balanced cases? E.g. various bad strings such as `{foo` and `bar}`.

Praburaj: There are already some tests on these unbalanced cases but escaped braces. Added some more negative ones.

Praburaj:
@loudej @avanderhoorn 

>Thanks for getting this data out! In terms of feedback, I must be honest, accessing it via that key seems a little unintuitive and more like a private api... Would be nice if there was some more supported way of doing this.

How about adding a property as well on the Formatter to give the original format string apart from this?

avanderhoorn:
I like the addition of the first class property. I'm still left scratching my head though over why we would need the `__OriginalFormat` in the `GetValues()`. As a caller of `GetValues()` I kinda expect that the collection I get back to be a collection of `keys` and `values` that I found in the `OriginalFormat`. I wouldn't expect one of the entries to be something that wasn't a found `key`/`value`... I'm not saying that its not useful, just that I doesn't really make sense given what the "contract" describes.

Eilon:
:+1: to what @avanderhoorn said. The original format string is not a real value, so we shouldn't put a fake value in the list.

Praburaj:
@loudej suggested to do this as if someone has a reference of `ILoggerStructure` one can directly get the original format by calling `GetValues()`. If its just a property one has to typecast the reference to the specific type to get the original format string. Exposing a property as well doesn't hurt if you have an instance of the `LoggerStructureFormatter` itself. So I ended up adding both.

yishaigalatzer:
@Praburaj you can probably add `{OriginalFormat}` as the key instead of `__OriginalFormat` ;)

Praburaj:
@yishaigalatzer yeah that's doable.

avanderhoorn:
What if `OriginalFormat` is a key the user already has?

Praburaj:
yeah. we'll have to special case this key. User cannot have this key name.

lodejard:
I think Yishai meant the key name would literally contain curly braces, which end-user format strings could not reproduce.

yishaigalatzer:
Yes exactly that

-----Original Message-----

Praburaj:
Addressed all the comments. If there are no other comments this PR can go in.

Eilon:
:shipit: If others have more comments, please log bugs so we can track those ideas separately. Thanks!

Praburaj:
Thanks everyone.

Issue #100, opened 2015-01-30
Make logging more "pay-for-play" by avoiding unnecessary allocations


justinvp:
Note: I'd be happy to submit a PR; I just wanted to first see if the owners are open to these suggestions.

There are opportunities to make logging more "pay-for-play," avoiding unnecessary allocations when possible.

For example, here's the current implementation of the `WriteVerbose` extension method:

```c#
public static void WriteVerbose([NotNull] this ILogger logger, string format, params object[] args)
{
    logger.Write(LogLevel.Verbose, 0,
        string.Format(CultureInfo.InvariantCulture, format, args), null, TheMessage);
}
```

With the current implementation, even when `LogLevel.Verbose` is disabled, this still results in allocations due to the call to `string.Format`.

A caller could workaround this by calling `IsEnabled(LogLevel.Verbose)` before calling `WriteVerbose`, but why not include the `IsEnabled` check inside the extension method itself?

For example:

```c#
public static void WriteVerbose([NotNull] this ILogger logger, string format, params object[] args)
{
    if (logger.IsEnabled(LogLevel.Verbose))
    {
        logger.Write(LogLevel.Verbose, 0,
            string.Format(CultureInfo.InvariantCulture, format, args), null, TheMessage);
    }
}
```

With the change above, the user doesn't have to know about or remember to check `IsEnabled`. The user simply calls `WriteVerbose` and gets the perf benefit without having to write any other code!

Next, consider the following usage where `streamIndex` and `count` are `int` and `id` is `ulong`.

```c#
_logger.WriteVerbose("OnReceived({0}, {1}, {2})", streamIndex, count, id);
```

This call will result in several allocations: the `params object[] args` array is allocated and the value types are boxed. It'd be nice if some additional extension method overloads were added to avoid these allocations when the `LogLevel` is disabled.

For example, the following generic overloads avoid the params array allocation and avoid boxing when the `LogLevel` is disabled:

```c#
public static void WriteVerbose<T>([NotNull] this ILogger logger, string format, T arg)
{
    if (logger.IsEnabled(LogLevel.Verbose))
    {
        logger.Write(LogLevel.Verbose, 0,
            string.Format(CultureInfo.InvariantCulture, format, new object[] { arg }), null, TheMessage);
    }
}

public static void WriteVerbose<T1, T2>([NotNull] this ILogger logger, string format, T1 arg1, T2 arg2)
{
    if (logger.IsEnabled(LogLevel.Verbose))
    {
[...]

PR #101, opened 2015-02-03
Make logging more "pay-for-play" by avoiding unnecessary allocations

justinvp:
This is an initial PR that demonstrates the changes as discussed in #100. This PR currently only includes changes for `LogLevel.Verbose` -- the remaining levels will be added, along with tests, if/when the owners are on board with such improvements.

### DIFF
--- a/src/Microsoft.Framework.Logging/LoggerExtensions.cs
+++ b/src/Microsoft.Framework.Logging/LoggerExtensions.cs
@@ -47,11 +47,85 @@ namespace Microsoft.Framework.Logging
         /// </summary>
         /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
         /// <param name="format">Format string of the log message.</param>
+        /// <param name="arg">The argument to format.</param>
+        public static void WriteVerbose<T>([NotNull] this ILogger logger, string format, T arg)
+        {
+            logger.Write(LogLevel.Verbose, 0, format, arg);
+        }
+
+        /// <summary>
+        /// Formats and writes a verbose log message.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
+        /// <param name="format">Format string of the log message.</param>
+        /// <param name="arg1">The first argument to format.</param>
+        /// <param name="arg2">The second argument to format.</param>
+        public static void WriteVerbose<T1, T2>([NotNull] this ILogger logger, string format, T1 arg1, T2 arg2)
+        {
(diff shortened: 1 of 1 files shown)


muratg:
@davidfowl does this look good to you?  I believe Justin may do the other levels if the approach is right. Perf improvements seem pretty good.

davidfowl:
Hey @justinvp, we made a bunch of changes to the logging APIs, it looks like these changes might not make such a big difference anymore but we should measure and confirm. There seems to be 2 general changes here:

- Avoid boxing if the log level isn't enabled
- Avoid string.Format

It might be worthwhile adding generic overloads if we realize that boxing is a huge cost but with the latest design, nothing should be formatted until the logger provider actually decides to log for that level. The overloads that used to format now allocate FormatLogValues object (https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/Internal/FormattedLogValues.cs#L13) which is only realized if the underlying logger decides it needs to log.

justinvp:
@davidfowl, the logging APIs, as currently implemented, are still allocation-heavy, even when the log level isn't enabled.

Consider the [following](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs#L46-L49):

```c#
public static void LogDebug([NotNull] this ILogger logger, string format, params object[] args)
{
    logger.Log(LogLevel.Debug, 0, new FormattedLogValues(format, args), null, MessageFormatter);
}
```

Allocations on each call to `LogDebug` (even when the log level isn't enabled):

 1. new params array
 2. boxing (for each arg that is a value type)
 3. new `FormattedLogValues`
 4. new `Func<object, Exception, string>` (for `MessageFormatter`)

The `Func` delegate allocation (4) can be trivially addressed by using a single cached `Func` instance. I've submitted a new PR for this, #193.

The `FormattedLogValues` allocation (3) can be avoided by checking `IsEnabled` before allocating it and calling the log method.

The params array allocation (1) can be avoided for a common number of arguments (typically three) with overloads.

The boxing allocations (2) can be avoided with generic overloads.

I'll update the PR to address 1-3 and measure to see.

Eilon:
@justinvp BTW our recommended logging pattern is for anyone concerned about perf to always check if logging is enabled for a particular level, and to only then do the logging.

muratg:
@justinvp Are you still looking at the changes here (in particular 1-3 above) and the measurements? Or should we close this one?

muratg:
@justinvp Let me know if you come back to this, and we can re-open the discussion/PR.  Thanks.

PR #105, opened 2015-02-10
Beautifying generic types

suhasj:
@Eilon @davidfowl 

Problem: The current experience of logging generics is that the name for the logger is type.fullname which in this case would be `namespace.classname[classname, assemblyinformation]` which is not friendly. It gets worse if the arity is 2

Fix: @Eilon provided a helper class that provides a beautified generic class name in the format `namespace.class<namespace.class>`. This should not affect the behavior of non generic classes

Sample output can be seen in the unit tests

### DIFF
--- a/src/Microsoft.Framework.Logging/LoggerFactoryExtensions.cs
+++ b/src/Microsoft.Framework.Logging/LoggerFactoryExtensions.cs
@@ -22,7 +22,8 @@ namespace Microsoft.Framework.Logging
                 throw new ArgumentNullException("factory");
             }
 
-            return factory.Create(typeof(T).FullName);
+            return factory.Create(TypeNameHelper.GetTypeDisplayFullName(typeof(T)));
         }
     }
+        
 }
--- a/src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
+++ b/src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
@@ -72,11 +72,7 @@ namespace Microsoft.Framework.Logging
                 text = t.Name;
             }
             var num = text.IndexOf('`');
-            if (num != -1)
-            {
-                return text.Substring(0, num);
-            }
-            return text;
+            return num != -1 ? text.Substring(0, num) : text;
         }
     }
 }
--- a/src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
+++ b/src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
@@ -1,0 +1,82 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.Framework.Logging
+{
+    public static class TypeNameHelper
+    {
+        public static string GetTypeDisplayFullName(Type type)
+        {
+            var sb = new StringBuilder(64);
+            ProcessTypeName(type, sb);
+            return sb.ToString();
+        }
+
+        private static void AppendGenericArguments(Type[] args, StringBuilder sb)
+        {
+            if (args.Length > 0)
+            {
+                sb.Append("<");
+                for (int i = 0; i < args.Length; i++)
+                {
+                    ProcessTypeName(args[i], sb);
+                    if (i + 1 < args.Length)
+                    {
+                        sb.Append(", ");
+                    }
+                }
+                sb.Append(">");
+            }
+        }
+
+        private static Type GetMostGenericTypeDefinition(Type t)
+        {
+            while (t.GetTypeInfo().IsGenericType)
+            {
+                var genericTypeDefinition = t.GetGenericTypeDefinition();
+                if (genericTypeDefinition == null || t == genericTypeDefinition)
+                {
+                    return t;
+                }
+                t = genericTypeDefinition;
+            }
+            return t;
+        }
+
+        private static void ProcessTypeName(Type t, StringBuilder sb)
+        {
+            if (t.IsGenericParameter)
+            {
+                sb.Append(t.GetTypeInfo().Name);
+                return;
+            }
+            if (t.GetTypeInfo().IsGenericType)
+            {
+                var mostGenericTypeDefinition = GetMostGenericTypeDefinition(t);
+                sb.Append(GetSimpleGenericTypeName(mostGenericTypeDefinition));
+                AppendGenericArguments(t.GetTypeInfo().GenericTypeArguments, sb);
+                return;
+            }
+            sb.Append(GetSimpleGenericTypeName(t));
+        }
+
+        private static string GetSimpleGenericTypeName(Type t)
+        {
+            var text = t.FullName;
+            if (text == null)
+            {
+                text = t.Name;
+            }
+            var num = text.IndexOf('`');
+            if (num != -1)
+            {
+                return text.Substring(0, num);
+            }
+            return text;
+        }
+    }
+}
--- a/src/Microsoft.Framework.Logging/project.json
+++ b/src/Microsoft.Framework.Logging/project.json
@@ -17,7 +17,8 @@
         },
         "aspnet50": {
             "frameworkAssemblies": {
-                "System.Collections.Concurrent": ""
+                "System.Collections.Concurrent": "",
+                "System.Reflection": ""
             }
         },
         "aspnetcore50": {
@@ -27,6 +28,7 @@
                 "System.Diagnostics.TraceSource": "4.0.0-beta-*",
                 "System.Globalization": "4.0.10-beta-*",
                 "System.Linq": "4.0.0-beta-*",
+		"System.Reflection": "4.0.10-beta-*",
                 "System.Threading": "4.0.10-beta-*"
             }
         },
--- a/test/Microsoft.Framework.Logging.Test/LoggerFactoryExtensionsTest.cs
+++ b/test/Microsoft.Framework.Logging.Test/LoggerFactoryExtensionsTest.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Framework.Logging.Test
     public class LoggerFactoryExtensionsTest
     {
 #if ASPNET50
-        [Fact] 
+        [Fact]
         public void LoggerFactoryCreateOfT_CallsCreateWithCorrectName()
         {
             // Arrange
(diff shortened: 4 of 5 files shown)


src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
Eilon: File name, folder name, and namespace don't match.

src/Microsoft.Framework.Logging/project.json
```diff
@@ -20,7 +20,8 @@
                 "Microsoft.Framework.Logging.Interfaces": { "version": "1.0.0-*", "type": "build" }
             },
             "frameworkAssemblies": {
-                "System.Collections.Concurrent": ""
+                "System.Collections.Concurrent": "",
+                "System.Reflection":""
```
Eilon: Formatting

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+            int num = text.IndexOf('`');
+            if (num != -1)
+            {
+                return text.Substring(0, num);
+            }
+            return text;
+        }
+    }
+
+    internal class TypeWrapper
```
Eilon: One top-level type per class. Should this be a private nested type?

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+    {
+        public static string GetTypeDisplayFullName(Type type)
+        {
+            return GetSimpleTypeName(type);
+        }
+
+        private static string GetSimpleTypeName(Type t)
+        {
+            StringBuilder sb = new StringBuilder(64);
+            ProcessTypeName(t, delegate (TypeWrapper writerItem)
```
Eilon: I think this should be simplified to not use the delegate (or lambda). Can just have a helper method that does this work and is called from wherever this delegate is otherwise passed around. I think that'll make the code quite simpler (just pass the StringBuilder around instead).

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
Eilon: (Or, per my other comment, remove this type entirely.)

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        private static void ProcessTypeName(Type t, Action<TypeWrapper> wrapperAction)
+        {
+            if (t.IsGenericParameter)
+            {
+                wrapperAction(new TypeWrapper(t.Name));
+                return;
+            }
+            if (t.GetTypeInfo().IsGenericType)
```
Praburaj: Can this go in an `else`?

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Type mostGenericTypeDefinition = GetMostGenericTypeDefinition(t);
```
Praburaj: `var` ?

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+                wrapperAction(new TypeWrapper(mostGenericTypeDefinition));
+                AppendGenericArguments(t.GetTypeInfo().GenericTypeArguments, wrapperAction);
+                return;
+            }
+            wrapperAction(new TypeWrapper(t));
+        }
+
+        private static string GetSimpleGenericTypeName(Type t)
+        {
+            string text = t.FullName;
```
Praburaj: `var`.

src/Microsoft.Framework.Logging/LoggerFactoryExtensions.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Text;
```
Praburaj: Is this needed?

suhasj: not needed. Fixed.

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
suhasj: may be `else if` but we need to check that condition

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
suhasj: Yeah I was able to trim that. I thought it was easier that way but after refactoring I was able to reduce the number the lines. In case if we need to have the ability to remove the namespace from the classname, then we would need it. I am not sure if want that fine grain capability

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
suhasj: last minute change so missed the classname. The namespace is consistent with all the files in that folder

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
suhasj: fixed

src/Microsoft.Framework.Logging/project.json
suhasj: fixed

src/Microsoft.Framework.Logging/Utils/TypeHelper.cs
suhasj: fixed all `vars`

src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+            }
+            var num = text.IndexOf('`');
+            if (num != -1)
+            {
+                return text.Substring(0, num);
+            }
+            return text;
+        }
+
+        private static void ConstructorTypeString(object typeInfo, StringBuilder sb)
```
Eilon: Just have two overloads of this method. Never take in `object` and try to cast it. Also, not sure what the method name is trying to convey... method names generally have verbs in them.

test/Microsoft.Framework.Logging.Test/LoggerFactoryExtensionsTest.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.Framework.Logging.Test
     public class LoggerFactoryExtensionsTest
     {
 #if ASPNET50
```
Eilon: Wondering - any reason for this `#ifdef`? Does this stuff not work on any other platform?

src/Microsoft.Framework.Logging/project.json
```diff
@@ -30,6 +31,7 @@
                 "System.Globalization": "4.0.10-beta-*",
                 "System.Linq": "4.0.0-beta-*",
                 "System.Threading": "4.0.10-beta-*",
+                "System.Reflection": "4.0.10-beta-*",
```
Praburaj: nit: `System.Reflection` before `System.Threading`.

test/Microsoft.Framework.Logging.Test/LoggerFactoryExtensionsTest.cs
Eilon: Oh is it because of Moq? If so then maybe at least some of these tests should avoid using Moq so that we get more coverage.

suhasj: The project.json does not have an entry for `aspnetcore50`. Our other repos that use Moq don't have them either. Should I remove the `ifdef` altogether since it won't be compiled for `aspnetcore50` anyway ?

src/Microsoft.Framework.Logging/project.json
suhasj: sure fixed

src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
suhasj: I was able to remove the method altogether. Thanks you for pushing on it :)

test/Microsoft.Framework.Logging.Test/LoggerFactoryExtensionsTest.cs
Eilon: Oh interesting. Yeah no point having the `#ifdef` in that case, but instead of that what would it take to have some coverage on `aspnetcore50`? I am especially interested in this case because some Reflection stuff could certainly be a bit different between the two. I'm not suggesting porting *all* the tests to that, but at least these new ones. If it's somehow super costly then we can reconsider.

suhasj: The `project.json` does not have `aspnetcore50` so the test project is not supposed to build against core clr. Should I enable that in this PR  or should we track that as a separate item ? IMO we should be looking at it separately. The project depends on Logging.Serilog project which is not doesn't compile for `coreclr` .

Eilon: Ah, yeah, the main logging test project shouldn't be depending on Serilog. Can you log a bug to split that out into a separate test project and have *that* one be `aspnet50`/`net45` (or whatever), and then change the main logging test project to run on everything? And then fix it? :smile:

suhasj: https://github.com/aspnet/Logging/issues/127 tracking this

suhasj:
@Eilon There is a separate bug to split the projects and add tests for coreclr. Can I merge this PR ? or do I club that with this as well ? I feel I can merge this and add that separately

avanderhoorn:
Thanks for making this accessible! As I'm sure you know, doing this type conversion is tedious, painful and error prone - https://github.com/Glimpse/Glimpse/blob/c61c88c8a4ba41b78b8030e1c49238a2f6400dcd/source/Glimpse.Core/SerializationConverter/CSharpTypeConverter.cs.

src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
```diff
@@ -0,0 +1,78 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.Framework.Logging
```
Eilon: Change to be in a ".Internal" namespace.

Eilon:
Yeah :shipit: after my last comment. Didn't realize this was still open.

src/Microsoft.Framework.Logging/Utils/TypeNameHelper.cs
suhasj: Thanks. I'll move it to `Microsoft.Framework.Logging.Internal`. Hope that is good

suhasj:
merged

PR #106, opened 2015-02-11
Remove portable target

bricelam:
Given the removal of ANIs, this isn't needed anymore


BrennanConroy:
Remove [Runtime.AssemblyNeutral] also?

BrennanConroy:
Oh, you did that in another PR already, never mind

bricelam:
(This depends on aspnet/EntityFramework#1682)

davidfowl:
Any reason we need to do this?

Issue #104, opened 2015-02-10
Update signature of Func<> param in logging to allow conditional evaluation of secure log messages


lukewaters:
The purpose of this change would be to both defer evaluation of debug level message, and also to avoid sending secure log messages to loggers that aren't listening on debug level

Signature for additional logging method would be like 

```
public void Write(LogLevel logLevel, int eventId, object state, Exception exception,
                           Func<bool, object, Exception, string> formatter)
```

Issue #103, opened 2015-02-09
Add Debug level to logging


lukewaters:
This logging level would be below verbose, and would indicate that sensitive application data is to be logged.  

With this change, the LogLevel enum would be changed to this

```
    public enum LogLevel
    {
        Debug = 1,
        Verbose = 2,
        Information = 3,
        Warning = 4,
        Error = 5,
        Critical = 6,
    }
```

PureKrome:
How does `Debug` differ from `Verbose`? Some people use `Trace` instead, but `Verbose` sounds like an acceptable _catchall_. I can't see the value of both `D` and `V`.

lukewaters:
This level differs from verbose in that log messages will contain data that would be omitted from verbose messages due to security concerns.  This bug works with issue #104 to allow messages to be conditionally evaluated, with secure information omitted from messages when log level is set to Verbose, and included in those log messages when the log level is set to debug

ycrumeyrolle:
Guidelines should be updated to explain the differences between debug and verbose.
https://github.com/aspnet/Logging/wiki/Guidelines

Eilon:
@lukewaters can you take a crack at this?

Eilon:
(And should this bug be closed as well?)

lukewaters:
Yep, I'll get to work on the docs

Eilon:
Thanks! :book:

PR #108, opened 2015-02-13
Adding debug level to logging and conditionally logging messages

lukewaters:
between debug and verbose levels

### DIFF
--- a/src/Microsoft.Framework.Logging.Serilog/SerilogLogger.cs
+++ b/src/Microsoft.Framework.Logging.Serilog/SerilogLogger.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Framework.Logging.Serilog
             return _logger.IsEnabled(ConvertLevel(logLevel));
         }
 
-        public void Write(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        public void Write(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, bool, string> formatter)
         {
             var level = ConvertLevel(logLevel);
             if (!_logger.IsEnabled(level))
@@ -49,7 +49,7 @@ namespace Microsoft.Framework.Logging.Serilog
             var message = string.Empty;
             if (formatter != null)
             {
-                message = formatter(state, exception);
+                message = formatter(state, exception, IsEnabled(LogLevel.Debug));
             }
             else
             {
@@ -89,6 +89,8 @@ namespace Microsoft.Framework.Logging.Serilog
                 case LogLevel.Information:
                     return LogEventLevel.Information;
                 case LogLevel.Verbose:
+                    return LogEventLevel.Debug;
+                case LogLevel.Debug:
                     return LogEventLevel.Verbose;
                 default:
                     throw new NotSupportedException();
--- a/src/Microsoft.Framework.Logging/LoggerExtensions.cs
+++ b/src/Microsoft.Framework.Logging/LoggerExtensions.cs
@@ -18,6 +18,131 @@ namespace Microsoft.Framework.Logging
         private static readonly Func<object, Exception, string> _loggerStructureFormatter = (state, error)
             => LoggerStructureFormatter((ILoggerStructure)state, error);
 
+        private static readonly Func<object, Exception, bool, string> SecureMessage =
+            (messages, error, secure) =>
+            {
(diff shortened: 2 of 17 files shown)


msotclas:
Hi __@lukewaters__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Luke Waters). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

src/Microsoft.Framework.Logging.Serilog/SerilogLogger.cs
```diff
@@ -89,6 +89,8 @@ private LogEventLevel ConvertLevel(LogLevel logLevel)
                 case LogLevel.Information:
                     return LogEventLevel.Information;
                 case LogLevel.Verbose:
+                    return LogEventLevel.Debug;
+                case LogLevel.Debug:
                     return LogEventLevel.Verbose;
                 default:
                     throw new NotSupportedException();
```
lukewaters: I looked at the log levels in 9 popular loggers, and serilog is the only logger that has debug as a lower level than verbose.  I am open to suggestions for as to what to do in this case

lukewaters:
This addresses issues #103 and #104

lukewaters:
Take a look @blowdart

blowdart:
LogLevel.Gossip? :) Debug works for me, doc comment should note the worry that PII etc could be pumped out at this level.

src/Microsoft.Framework.Logging/LoggerExtensions.cs
```diff
@@ -18,6 +18,131 @@ public static class LoggerExtensions
...
+            (state, error, secure) =>
+            {
+                var formatter = (ILoggerStructure[]) state;
+
+                if (secure)
+                {
+                    return LoggerStructureFormatter(formatter[0], error);
+                }
+                return LoggerStructureFormatter(formatter[1], error);
+
```
Eilon: Need to clean up lots of formatting (extra newline here; lots of other places are missing spaces).

src/Microsoft.Framework.Logging/LoggerExtensions.cs
```diff
@@ -18,6 +18,131 @@ public static class LoggerExtensions
...
+        //------------------------------------------DEBUG------------------------------------------//
+
+        /// <summary>
+        /// Writes a verbose log message.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
+        /// <param name="data">The message to log at verbose level.</param>
+        /// <param name="secureData">The message to log at debug level.</param>
+        // FYI, this field is called data because naming it message triggers CA1303 and CA2204 for callers.
+        public static void WriteDebug([NotNull] this ILogger logger, string secureData, string data)
```
Eilon: Not really a fan of calling I "secureXyz" in various places. This is the only place that the security nomenclature seems to be used. Everywhere else it's a distinction between "debug" and not, and that seems orthogonal to the distinction between "secure" and not.

davidfowl:
I dont understand why there an extra bool on the lambda. That looks completely misplaced. The only change that's supposed to happen was the addition of this Debug log level and a new `ILoggerFactory` property to set the minimum.

/cc @anurse @loudej

Eilon:
Yeah this change seems to have a lot of unrelated things going on here that perhaps ought to be separated. (I know I missed some of the design discussions.)

lodejard:
Yes, please submit a PR which adds the "Debug" level to the enum but does not change the formatting mechanics. The second step is to add an `ILoggerFactory.MinimumLevel` property which defaults to Verbose. The Logger instances at that point will check the level against the _factory.MinimumLevel in IsEnabled and WriteCore, and won't pass the call along to the providers' ILogger instances if the argument is lower than the property.

So, all sensitive information is logger at Debug, and to see it the app must set `loggerFactory.MinimumLevel = LogLevel.Debug` in code before it will flow.

lukewaters:
I'm going to close this pull request as I clearly need to understand the problem better before I can fix the issue.

Issue #113, opened 2015-02-17
Add ILoggerFactory.MinimumLevel


lukewaters:
This should be set to Verbose by default, anything logged at a level below MinimumLevel will not be sent to any registered logger.

Eilon:
How exactly is this method used? Note that the logger factory implementation is not generally something that is customized. How would an app tweak the minimum level?

lukewaters:
@lodejard  described this behavior in a comment on one of my previous pull requests

> The second step is to add an ILoggerFactory.MinimumLevel property which defaults to Verbose. The Logger instances at that point will check the level against the _factory.MinimumLevel in IsEnabled and WriteCore, and won't pass the call along to the providers' ILogger instances if the argument is lower than the property.

>So, all sensitive information is logger at Debug, and to see it the app must set loggerFactory.MinimumLevel = LogLevel.Debug in code before it will flow.

Eilon:
Got it, thanks!

davidfowl:
It's not a method

lukewaters:
Fixed in #114

PR #114, opened 2015-02-18
Fix for #113 changes to add minimum logLevel to ILoggerFactory

lukewaters:


lukewaters:
@davidfowl @lodejard @blowdart

src/Microsoft.Framework.Logging/LoggerFactory.cs
```diff
@@ -15,6 +15,7 @@ public class LoggerFactory : ILoggerFactory
         private readonly Dictionary<string, Logger> _loggers = new Dictionary<string, Logger>(StringComparer.Ordinal);
         private ILoggerProvider[] _providers = new ILoggerProvider[0];
         private readonly object _sync = new object();
+        private LogLevel? _minimumLevel;
```
davidfowl: Do we need this? Can't we just set the default?

```C#
public LogLevel MinimumLevel { get; set; } = LogLevel.Verbose;
```

Also the default level is Debug right?

src/Microsoft.Framework.Logging/Utils/TestLoggerFactory.cs
```diff
@@ -22,5 +22,7 @@ public ILogger Create(string name)
         public void AddProvider(ILoggerProvider provider)
         {
         }
+
+        LogLevel ILoggerFactory.MinimumLevel { get; set; }
```
davidfowl: Why is this implicit?

src/Microsoft.Framework.Logging/LoggerFactory.cs
lukewaters: no. The whole idea behind this bug is to prevent messages logged at the Debug level from being distributed to all the loggers by default.

lukewaters: but I'll change it to an implicit property

src/Microsoft.Framework.Logging/Utils/TestLoggerFactory.cs
lukewaters: I'll update this to be the same as above on all the loggers

src/Microsoft.Framework.Logging.Interfaces/ILoggerFactory.cs
```diff
@@ -16,5 +16,7 @@ public interface ILoggerFactory
         ILogger Create(string name);
 
         void AddProvider(ILoggerProvider provider);
+
+        LogLevel MinimumLevel { get; set; }
```
Eilon: Code style: Generally, properties go after ctors and before methods. Should update the other files too.

src/Microsoft.Framework.Logging/Utils/TestLoggerFactory.cs
```diff
@@ -22,5 +22,7 @@ public ILogger Create(string name)
         public void AddProvider(ILoggerProvider provider)
         {
         }
+
+        LogLevel ILoggerFactory.MinimumLevel { get; set; } = LogLevel.Verbose;
```
Eilon: This one will change too? BTW I think @davidfowl and you are confusing implicit/explicit. An *explicit* interface implementation is when you include the interface name in the implementation (hence it's explicit). But we normally do *implicit* interface implementations, where the interface name is omitted.

lukewaters:
Is there any more feedback on this change?

src/Microsoft.Framework.Logging.Interfaces/ILoggerFactory.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Framework.Logging
     /// </summary>
     public interface ILoggerFactory
     {
+        LogLevel MinimumLevel { get; set; }
```
Eilon: Doc comment.

src/Microsoft.Framework.Logging.Interfaces/ILoggerFactory.cs
```diff
@@ -9,6 +9,11 @@ namespace Microsoft.Framework.Logging
     public interface ILoggerFactory
     {
         /// <summary>
+        /// The minimum level of log messages sent to registered loggers
```
Eilon: Sorry, nit-picker's corner here - can you end the sentence with a period? :crying_cat_face:

Eilon:
I wasn't in the design meeting for this, so without knowing the design goals I can only sign off on what I've seen, so :shipit: on that.

PR #115, opened 2015-02-20
Adding extension methods for logging at debug level

lukewaters:
This should bring the extension support for the debug level on par with the other levels


lukewaters:
ping @anurse @davidfowl

anurse:
Looks like these parallel with the other LogLevels, so I say :shipit:

PR #128, opened 2015-02-25
LoggerStructureFormatter -> Internal Namespace

BrennanConroy:
#123


src/Microsoft.Framework.Logging.Interfaces/project.json
```diff
@@ -9,6 +9,8 @@
         "aspnet50": { },
         "aspnetcore50": {
             "dependencies": {
+                "System.Collections": "4.0.10-*",
```
davidfowl: The other versions have beta-*

Eilon:
Move the file too. It's in the wrong folder.

BrennanConroy:
Shouldn't it be in the Microsoft.Framework.Logging folder?

Eilon:
Namespaces and folder names are supposed to match.

BrennanConroy:
There isn't a .Internal folder, I can make one if we want to change that

Eilon:
`md Internal` :smile:

BrennanConroy:
Should we make a new project for it?

Eilon:
For interfaces that used to be ANI's we create new projects. But this is just some random type we want to stuff into an `Internal` namespace because it's an implementation detail, no?

BrennanConroy:
I looked at some other repos and saw how they did it, so I copied them. Just put a Internal folder inside the project

Eilon:
Yup.

Issue #122, opened 2015-02-21
API review: Remove NullLogger and related classes


muratg:
- NullLogger
- NullLoggerFactory
- NullDisposable

BrennanConroy:
NullDisposable is used in one place by the TestLogger. Current thought is to just use null instead of NullDisposable.Instance, this works fine as long as .Dispose is not explicitly called on the "IDisposable" which it isn't in all the tests that use this

PR #129, opened 2015-02-26
Removing Null* classes

BrennanConroy:
#122 
Question: Is there a better thing to do than return null in the TestLogger?
TODO: Look at other repos this may affect before pushing


Eilon:
I *think* that's fine, so :shipit: for now.

@lodejard in case maybe he has a thought on this.

davidfowl:
I believe null is handled.

lodejard:
For the test logger it would be pretty simple to return a private nested implementation of idisposable that no-ops. But in either case, :shipit:

PR #132, opened 2015-03-04
Snazzify the Console Output and move the extension methods to Interfaces

anurse:
This is a pre-req for aspnet/XRE#1314 that pulls logging in to the core runtime. It doesn't fully fix #119, but it does the first half by pulling the Extension methods over to the Interfaces assembly. I don't have the cycles to do the rename part :(.

@davidfowl suggested a minor tidy up of Console Logger if we're going to use it in the runtime:

capture

/cc @BrennanConroy @davidfowl @muratg

Issue #124, opened 2015-02-21
API review: LoggerStructureBase and LoggerStructureFormat polish


muratg:
**LoggerStructureBase**
- Rename to ReflecionBasedLogValues (Check with Glenn)
- Remove Message (another work item tracks removal of this from the interface)

**LoggerStructureFormat**
- Move to .Internal namespace
- Rename class to FormattedLogValues
- Should follow pattern similar to LoggerStructureBase
- Remove Message (another work item tracks removal of this from the interface)

BrennanConroy:
@glennc ReflecionBasedLogValues?
Also, LoggerStructureFormatter is only used in the Logging repo so I'm thinking move it to .Internal namespace too (and renaming, LogValuesFormatter?)

glennc:
Yes to both of those questions.

PR #133, opened 2015-03-04
Polish StructureBase and StructureFormat

BrennanConroy:
#124
Will need a little fiddling when #132 goes in (files moving folders)


muratg:
:shipit:

Issue #142, opened 2015-03-05
Move AddLogging to Logging


davidfowl:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNet.Hosting/HostingServicesCollectionExtensions.cs#L16

Issue #141, opened 2015-03-05
TraceSourceLoggerProvider > InitializeTraceSource is no use


leshin:
The Listeners and Switchs always to the same with _rootSourceSwitch !!

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/TraceSourceLoggerProvider.cs#L16

davidfowl:
?

leshin:
Why use TraceSource parentTraceSource = GetOrAddTraceSource(parentSourceName) ?
All the created logger by TraceSourceLoggerProvider that always use the _rootSourceSwitch and _rootTraceListener !
How to set different switch and listener for each logger?

davidfowl:
What's the scenario you're trying to accomplish? The pattern here allows you to filter traces by prefix. It means that you can turn tracing on for Microsoft.AspNet and get all of the logs:

http://www.tugberkugurlu.com/archive/logging-in-the-owin-world-with-microsoft-owin--introduction

Check out this blog for an example.

leshin:
Oh, I see !  I forgot can configure the switch and listener in the web.config.  Thank you very much!

Issue #140, opened 2015-03-05
WriteContext should be moved to Test


davidfowl:
It's not used anywhere else https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/Utils/WriteContext.cs

PR #143, opened 2015-03-05
Moar refactoring!

davidfowl:
- Moved TraceSourceLogger to Microsoft.Framework.Logging.TraceSource
- Removed Logging dependency from all providers, they only need
the .Interfaces reference
- Moved WriteContext to the test project
- Added an extension method AddTraceSource to add the TraceSource provider
- Added AddLogging to the Logging project
- Removed RootTraceName

#142
#141
#140


src/Microsoft.Framework.Logging.TraceSource/TraceSourceFactoryExtensions.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
Eilon: Headerz

davidfowl: Do we have a script for that?

Eilon: No but I'll go ahead and do it.

Eilon: Fixed in https://github.com/aspnet/Logging/pull/145

PR #146, opened 2015-03-06
[Feedback] Adding IConfiguration overload for console logger

suhasj:
@lodejard @glennc 

This a feedback PR so you can refrain from nitpick. The final PR will have unit tests as well

Some clarifications needed
- Is the subconfig name `ConsoleLogger` fine or should we rename it.
- Also should the key value pairs be `"category":"logLevel"` or should it be `"category":{"category1","category2"}`. I am trying to keep it simple to went with the first approach
- Should the loglevel allowed be >= loglevel or == level

### DIFF
--- a/src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs
+++ b/src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.ComponentModel;
+using Microsoft.Framework.ConfigurationModel;
 using Microsoft.Framework.Logging.Console;
 
 namespace Microsoft.Framework.Logging
@@ -33,6 +35,24 @@ namespace Microsoft.Framework.Logging
         public static ILoggerFactory AddConsole(this ILoggerFactory factory, LogLevel minLevel)
         {
             factory.AddProvider(new ConsoleLoggerProvider((category, logLevel) => logLevel >= minLevel));
+            return factory;
+        }
+
+        /// <summary>
+        ///     Adds a console logger that is enabled as defined by the IConfiguration object
+        /// </summary>
+        /// <param name="configuration">IConfiguration object with filter attributes</param>
+        /// <returns></returns>
+        public static ILoggerFactory AddConsole(this ILoggerFactory factory, IConfiguration configuration)
+        {
+            factory.AddProvider(new ConsoleLoggerProvider((category, logLevel) =>
+            {
+                LogLevel minLevel;
+
+                return Enum.TryParse(configuration?.GetSubKey("ConsoleLogger")?.Get(category), out minLevel) &&
+                       logLevel >= minLevel;
+            }));
+
             return factory;
         }
     }
--- a/samples/SampleApp/LocalConfig.json
+++ b/samples/SampleApp/LocalConfig.json
@@ -1,0 +1,5 @@
+﻿{
+    "ConsoleLogger": {
+        "SampleApp.Program": "Information"
+    }
+}
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Microsoft.Framework.ConfigurationModel;
+using System;
 using System.Collections.Generic;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Logging.Console;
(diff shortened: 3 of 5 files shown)


src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs
```diff
@@ -35,5 +37,23 @@ public static ILoggerFactory AddConsole(this ILoggerFactory factory, LogLevel mi
...
+        /// </summary>
+        /// <param name="configuration">IConfiguration object with filter attributes</param>
+        /// <returns></returns>
+        public static ILoggerFactory AddConsole(this ILoggerFactory factory, IConfiguration configuration)
+        {
+            factory.AddProvider(new ConsoleLoggerProvider((category, logLevel) =>
+            {
+                LogLevel minLevel;
+
+                return Enum.TryParse(configuration?.GetSubKey("ConsoleLogger")?.Get(category), out minLevel) &&
```
Eilon: You'd presumably want to do this just once, when the AddConsole method is called. That way you don't parse the enum every time, but more importantly, you fail fast in case there's some funky config. Otherwise the failure won't happen until super late (and ironically, probably won't get logged because the logger is busted!).

davidfowl: And can we get a unit test?

lodejard: Don't get a subkey at all, just get categories from the iconfiguration passed in

lodejard:
Just "category":"loglevel" is fine.

Also if that returns null, remove the last part of the category based on the final dot and try getting that. Repeat until one returns non-null. This enables the logger to be configured at a base namespace rather than needing one entry per category name.

Be sure that's cached per ilogger, like Eilon mentioned

src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs
davidfowl: Also is there an exact match or is it prefix matching? i.e.

```json
{
    "Microsoft.AspNet" : "off"
}
```

Oh btw, can I turn off sections completely?

suhasj:
@Eilon @davidfowl @loudej 

Thank you for the feedback. I got some good feedback here so I'll take this forward based on that. Will update once done

src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs
suhasj: @Eilon @lodejard  I won't know the `category` until the `CreateLogger` on the Provider is invoked. To read all categories before hand from configuration and cache it, I would need to get an `IEnumerable` from the Configuration object to parse each entry and maintain a Dictionary which I can pass to the Func. Do you think this is the right approach ? In that scenario, I need a method on IConfiguration something like `GetAllKey` so that I can iterate on them to read each value and perform the operation

Eilon: Oh I see... yeah maybe just cache each one as you get it... I don't see how you could read them all ahead of time.

suhasj: Related to this, should there be a method on IConfiguration that returns the list of all keys ?

Eilon: It's perhaps conceivable that there could be, but I don't think anyone has really needed one yet. Even the case here doesn't really need it that badly. I'm not even sure it would benefit very much from it, would it?

Eilon: Could log a bug to see if anyone else has thoughts on it.

lodejard: Does GetSubKeys() do what you were asking? I'm on phone atm, but I believe that's available.

suhasj: It returns a `IEnumerable<KeyValuePair<string,IConfiguration>>` so we have iterate of each object to read the key value. But I think this confuses me in the sense that the value in KeyValuePair is IConfiguration, so entries of type "category":"value" would be cast to IConfiguration ? Let me try out a small example and get back

davidfowl:
We can punt this for now and revisit later

PR #152, opened 2015-03-14
Fixing some Logger implementation details

lodejard:
* Logger.Log shouldn't call IsEnabled on individual loggers, at that
point it's just unnecessary overhead
* Logger.IsEnabled shouldn't use .Any for performance reasons

### DIFF
--- a/src/Microsoft.Framework.Logging/Logger.cs
+++ b/src/Microsoft.Framework.Logging/Logger.cs
@@ -27,9 +27,9 @@ namespace Microsoft.Framework.Logging
 
         public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
         {
-            foreach (var logger in _loggers)
+            if (logLevel >= _loggerFactory.MinimumLevel)
             {
-                if (logLevel >= _loggerFactory.MinimumLevel && logger.IsEnabled(logLevel))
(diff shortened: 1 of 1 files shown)


davidfowl:
Fix your git user name!!

lodejard:
Urb

src/Microsoft.Framework.Logging/Logger.cs
```diff
@@ -38,16 +38,27 @@ public void Log(LogLevel logLevel, int eventId, object state, Exception exceptio
...
         public bool IsEnabled(LogLevel logLevel)
         {
-            return logLevel >= _loggerFactory.MinimumLevel && _loggers.Any(l => l.IsEnabled(logLevel));
+            if (logLevel < _loggerFactory.MinimumLevel)
+            {
+                return false;
+            }
+            foreach (var logger in _loggers)
+            {
+                if (!logger.IsEnabled(logLevel))
```
BrennanConroy: Doesn't this change the original behavior? Or do we want to change the behavior.
Should be ```if(logger.IsEnabled(logLevel)) return true;``` return false at the end of the function because no logger was found for that logLevel.

lodejard: Wow! What a dumb mistake, nice catch!

BrennanConroy:
:shipit:

PR #159, opened 2015-03-31
Logger extension to create scope based on format string

kichalla:
@rynowak @loudej @suhasj 

FYI:
- I have verified that this change works successfully in ConsoleLogger(where its noop), SerilogLogger writing to ElasticSearch
- I had to make some changes in ElmLogger where it does not handle the case where the state is `ILogValues`...will send out a separate PR for it in diagnostics repo.

PR #160, opened 2015-03-31
Rename BeginScope to BeginScopeImpl in favor of using the BeginScope log...

kichalla:
...ger extension method.

This change will probably effect other repos too, so will make changes there too.

@rynowak @suhasj @loudej

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -63,7 +63,7 @@ namespace SampleApp
                 _logger.LogWarning("Unexpected warning", ex);
             }
 
-            using (_logger.BeginScope("Main"))
+            using (_logger.BeginScopeImpl("Main"))
             {
                 Console.WriteLine("Hello World");
 
--- a/src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs
@@ -138,7 +138,7 @@ namespace Microsoft.Framework.Logging.Console
             }
         }
 
-        public IDisposable BeginScope(object state)
+        public IDisposable BeginScopeImpl(object state)
         {
             return null;
         }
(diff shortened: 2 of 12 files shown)


msotclas:
Hi __@kichalla__, I'm your friendly neighborhood Microsoft Open Technologies, Inc. Pull Request Bot (You can call me MSOTBOT). Thanks for your contribution!
    <p>
        It looks like you're a Microsoft contributor (Kiran Challa). If you're full-time, we DON'T require a Contribution License Agreement. If you are a vendor, please DO sign the electronic Contribution License Agreement. It will take 2 minutes and there's no faxing! https://cla2.msopentech.com.
    </p>

TTYL, MSOTBOT;

samples/SampleApp/Program.cs
```diff
@@ -63,7 +63,7 @@ public void Main(string[] args)
                 _logger.LogWarning("Unexpected warning", ex);
             }
 
-            using (_logger.BeginScope("Main"))
+            using (_logger.BeginScopeImpl("Main"))
```
kichalla: Will change this sample (and others...like in Entropy) to use `BeginScope` logger extension once PR https://github.com/aspnet/Logging/pull/159 is checked in.

Eilon:
I assume this was something that was discussed with @lodejard ? The code looks :shipit: to me.

rynowak:
This rename was the best option we could come up with in the logging hackathon the other day. :laughing: 

The intent is to still define the interface in a simple way, but make calling this method directly unattractive.

Eilon:
Yeah you definitely achieved those goals... very unattractive :smile:

kichalla:
commit d072cde33ec988e879504232bd9c43bd75a9a9be

avanderhoorn:
This might just be me but `BeginScopeImpl` is really ugly. For instance, at first glance I'm not even sure what "Impl" means... If its something you don't want people to really use, surely there is a better way of doing this.

tugberkugurlu:
I 100% agree with @avanderhoorn here. Why is this prefixed with `Impl`? Looking at the PR and changes but I cannot understand why and it looks like this is a member which should not be public but it's even worse: it's on the interface :smile:

Eilon:
We're certainly open to suggestions on better ways of doing this... part of the idea is that most consumers of this interface will see the extension methods alongside the interface method and will call the "prettiest" and "most obviously correct" methods, and not some ugly "XyzImpl" method.

PR #172, opened 2015-04-14
Handle enumerable value when formatting log values

kichalla:
@rynowak @lodejard @Eilon 

Text loggers like Console use the `Format` method to get the string representation of the log values. The current logic causes `ToString()` to be called on values which implement IEnumerable. This PR handles one level depth of flattening this values for showing in text loggers.

Example scenario(in case when an MVC view is not found, we would like to provide a list of locations that the view was searched for):
```c#
var viewEngineResult = viewEngine.FindView(context, viewName);
if(!viewEngineResult.Success)
{
    logger.LogError(
        "The view '{ViewName}' was not found. Searched locations: {SearchedViewLocations}", 
        viewName,
        viewEngineResult.SearchedLocations);
}
```

### DIFF
--- a/src/Microsoft.Framework.Logging.Interfaces/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Framework.Logging.Interfaces/Internal/LogValuesFormatter.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using System.Globalization;
 using System.Text;
 
@@ -112,6 +114,32 @@ namespace Microsoft.Framework.Logging.Internal
 
         public string Format(object[] values)
         {
+            if (values != null)
+            {
+                for (int i = 0; i < values.Length; i++)
+                {
+                    var value = values[i];
+
+                    if (value == null)
+                    {
+                        continue;
+                    }
+
+                    // since 'string' implements IEnumerable, special case it
+                    if (value is string)
+                    {
+                        continue;
+                    }
+
+                    // if the value implements IEnumerable, build a comma separated string.
+                    var enumerable = value as IEnumerable;
+                    if (enumerable != null)
+                    {
+                        values[i] = string.Join(", ", enumerable.Cast<object>().Where(obj => obj != null));
+                    }
+                }
+            }
+
             return string.Format(CultureInfo.InvariantCulture, _format, values);
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Framework.Logging.Interfaces/Internal/LogValuesFormatter.cs
```diff
@@ -112,6 +113,44 @@ private static int FindIndexOf(string format, char ch, int startIndex, int endIn
...
+                        {
+                            if (sb.Length > 0)
+                            {
+                                sb.Append(", ");
+                            }
+
+                            sb.Append(enumerator.Current);
+                        }
+
+                        values[i] = sb.ToString();
```
rynowak: You can replace this whole block with `values[i] = string.Join(", ", enumerable.Cast<object>())`

rynowak:
unit test?

rynowak:
:watch:

Eilon:
Looks good, taking into account @rynowak 's comments.

src/Microsoft.Framework.Logging.Interfaces/Internal/LogValuesFormatter.cs
kichalla: :+1:

kichalla:
Updated. /cc: @rynowak @Eilon

test/Microsoft.Framework.Logging.Test/project.json
```diff
@@ -4,14 +4,15 @@
         "Microsoft.Framework.Logging.Console": "1.0.0-*",
         "Microsoft.Framework.Logging.TraceSource": "1.0.0-*",
         "Microsoft.Framework.Logging.Serilog": "1.0.0-*",
-        "xunit.runner.aspnet": "2.0.0-aspnet-*"
+        "xunit.runner.aspnet": "2.0.0-aspnet-*",
+        "Microsoft.AspNet.Http.Extensions": "1.0.0-*"
```
Eilon: Sort!

test/Microsoft.Framework.Logging.Test/FormattedLogValuesTest.cs
```diff
@@ -71,5 +72,80 @@ public void LogValues_With_UnbalancedBraces(string format)
                 logValues.ToString();
             });
         }
+
+        // message format, format arguments, expected message
+        public static TheoryData<string, object[], string> FormatsEnumerableValuesData
+        {
+            get
+            {
+                return new TheoryData<string, object[], string>
```
Eilon: What about test cases for:
1. The sub-enumerable has `null` values
2. Non-strings in the sub-enumerables

kichalla: :+1: thanks...add these test cases found a bug!

kichalla:
updated. /cc: @Eilon @rynowak

rynowak:
:shipit:

kichalla:
@Eilon I have updated the PR, any more comments?

Eilon:
:shipit: !

kichalla:
Sure, I can remove them...its not really required to use media type header value for these tests...I can do it in a different way...

PR #181, opened 2015-05-11
Initial implementation of the Event Bus (INotifier)

rynowak:
This implementation uses emit to support duck-typing of data for
notification events on Net45 and DNXCore50.

For now the initial dispatch or 'splat' of event data is done using
reflection. Next change will update it to also use code generation.


src/Microsoft.Framework.Notify/Notifier.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+
+namespace Microsoft.Framework.Notify
```
avanderhoorn: +1 to the namespace this lives in.

avanderhoorn:
@rynowak With the way you said "This implementation uses emit to support duck-typing of data for
notification events on Net45 and DNXCore50." Does this mean that you are expecting this to change or was it just info about the implementation for people info?

rynowak:
> This implementation uses emit to support duck-typing of data for
notification events on Net45 and DNXCore50

Mostly informative. There's also no emit on portable 159, so if we care about event bus on that platform it comes without duck-typing.

avanderhoorn:
Humm how do we fix this! We need it for `core` and we need it to have duck-typing! Who do we need to bribe, beg to, have a hit man sent after ( @davidfowl your our hitman :D ), etc???

rynowak:
There is duck-typing support on CoreCLR/dnx.

Portable Profile151 includes windows phone and store apps where there is no support for codegen/emit. We have some libraries that produce an additional flavor for this platform in addition to ney45 and dnxcore50

avanderhoorn:
Sounds good to me.

rynowak:
Moving this PR to another repo.

avanderhoorn:
Which repo it go to?

rynowak:
(will be going to) https://github.com/aspnet/EventNotification/

avanderhoorn:
Thanks!

PR #185, opened 2015-05-17
Removes the Serilog provider and its tests

nblumhardt:
The provider has moved to https://github.com/serilog/serilog-framework-logging (many thanks!)

This PR removes the obsolete _Microsoft.Framework.Logging.Serilog_ sample so that users are more likely to track down the shipped/maintained package.

Unfortunately the API of _m.f.Logging_ has changed slightly (`BeginScopeImpl()`) since VS2015 RC packages were published, so the `SampleApp` can't yet be built against the RC _s.f.Logging_ package on NuGet.

Having the JSON dump in the sample was a nice way to inspect the structured data on a log event, it can definitely be lived without however we could also:

1. Add a "JSON dump" provider to the sample to emulate what the Serilog one was doing, or
2. Reinstate that part of the sample once `dev` and the NUPKG line up again (I'd happily PR this when the time comes)


Eilon:
I don't see the actual provider being deleted here... is GitHub just not showing it?

nblumhardt:
Ooops - that was a somewhat significant omission. Fixed now :-)

Eilon:
Any chance you could squash the commits into one? It'll make for a cleaner merge.

nblumhardt:
Should be good to go now :+1:

Eilon:
Thanks!

PR #187, opened 2015-05-22
[Fixes #80] Formatter shouldn't add exception for overloads where exc…

### LINKED ISSUE #80: Formatter shouldn't add exception for overloads where exception isn't passed in
Specific example,
```
        private static void Write(
            this ILogger logger,
            LogLevel logLevel,
            ILoggerStructure state,
            Exception exception = null)
        {
            logger.Write(logLevel, 0, state, exception, _loggerStructureFormatter);
        }
```
the formatter there always adds a newline and the exception, when it should just be the ILS formatted message

kichalla:
…eption isn't passed in

@Eilon @BrennanConroy

### DIFF
--- a/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs
@@ -578,14 +578,25 @@ namespace Microsoft.Framework.Logging
             logger.Log(logLevel, eventId, state, exception, MessageFormatter);
         }
 
-        private static string MessageFormatter(object state, Exception exception)
+        private static string MessageFormatter(object state, Exception error)
         {
-            if (exception == null)
+            if (state == null && error == null)
             {
-                return state?.ToString();
+                throw new InvalidOperationException("No message or exception details were found " +
+                    "to create a message for the log.");
             }
 
-            return string.Format(CultureInfo.CurrentCulture, "{0}{1}{2}", state, Environment.NewLine, exception);
+            if (state == null)
+            {
+                return error.ToString();
+            }
+
+            if (error == null)
+            {
+                return state.ToString();
+            }
+
+            return string.Format(CultureInfo.CurrentCulture, "{0}{1}{2}", state, Environment.NewLine, error);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs
```diff
@@ -581,17 +575,17 @@ public static void LogCritical([NotNull] this ILogger logger, int eventId, strin
...
-        private static string LogValuesFormatter(ILogValues state, Exception exception)
+        private static string MessageFormatter(object state, Exception exception)
         {
             if (exception == null)
             {
                 return state?.ToString();
             }
 
-            return state?.ToString() + Environment.NewLine + exception;
+            return string.Format(CultureInfo.CurrentCulture, "{0}{1}{2}", state, Environment.NewLine, exception);
```
BrennanConroy: Is state no longer able to be null now?

kichalla: good question...state being null is like the log message being null...yeah, a log statement without a message is unusual but definitely possible(scenario being: an exception is caught and user just wants to log the exception and does not want to provide any message)...I was initially thinking of why not have NotNull attributes on all the extensions where we accept `state`, but the above scenario invalidates it... @Eilon, thoughts?

Eilon: I would think that `state` could definitely be null, no? I bet the typical quick & dirty logging pattern is to just have a log level and a message.

kichalla: > I bet the typical quick & dirty logging pattern is to just have a log level and a message.

Here `state` represents the message actually...so did you mean to say 'is to just have a log level and an exception'?

Eilon: Oh, state is a message? Well that's nice :smile: I really did mean log level + message. Imagine a call to `Logger.LogWarning("blah blah bad thing happened")` - that's super common. But not message? Maybe less common...

kichalla: I hope I understood what you meant in the last comment, but I have updated the PR now. please take a look /cc: @Eilon @BrennanConroy

BrennanConroy:
Looks good to me
:shipit:

kichalla:
commit 73bd91bbe9aebe74743e3215ac0dbb5d6612320e

PR #193, opened 2015-06-11
Cache the formatter delegate

justinvp:
Avoid allocating a new delegate instance each time log is called.

### Details

The existing extension methods look like this:

```c#
public static void LogVerbose([NotNull] this ILogger logger, string format, params object[] args)
{
    logger.Log(LogLevel.Verbose, 0, new FormattedLogValues(format, args), null, MessageFormatter);
}
```

Which is equivalent to:

```c#
public static void LogVerbose([NotNull] this ILogger logger, string format, params object[] args)
{
    logger.Log(LogLevel.Verbose, 0, new FormattedLogValues(format, args), null, new Func<object, Exception, string>(MessageFormatter));
}
```

A new `Func` delegate is allocated each time the method is called.

### Solution

Instead, a single `Func` delegate can be cached and reused.

### DIFF
--- a/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Framework.Logging.Abstractions/LoggerExtensions.cs
@@ -13,6 +13,8 @@ namespace Microsoft.Framework.Logging
     /// </summary>
     public static class LoggerExtensions
     {
+        private static readonly Func<object, Exception, string> _messageFormatter = MessageFormatter;
+
         //------------------------------------------DEBUG------------------------------------------//
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


kichalla:
:shipit:  
/cc: @muratg

muratg:
Looks good to me too, thanks @justinvp!  @davidfowl you OK with this?  If so, @kichalla can take this one in.

davidfowl:
:shipit:

hishamco:
@kichalla after the commit https://github.com/aspnet/Logging/commit/98fe4f77b99dc93fb7faf0888f0430195377c0ff i think this PR should be closed

kichalla:
commit 98fe4f77b99dc93fb7faf0888f0430195377c0ff

Thanks @hishamco

PR #195, opened 2015-06-19
Make FormatMessage protected virtual

prasannavl:
Make `FormatMessage` method protected so that subclasses can override it for adding other useful info like time, and custom prefixes or suffixes without writing another console logger from scratch.

Also abstracted out `GetLogLevelString` as a protected virtual method.

### DIFF
--- a/src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Framework.Logging.Console
                 SetConsoleColor(logLevel);
                 try
                 {
-                    Console.WriteLine(FormatMessage(logLevel, message));
+                    Console.WriteLine(FormatMessage(logLevel, _name, message));
                 }
                 finally
                 {
@@ -73,10 +73,10 @@ namespace Microsoft.Framework.Logging.Console
             }
         }
 
-        private string FormatMessage(LogLevel logLevel, string message)
+        public virtual string FormatMessage(LogLevel logLevel, string logName, string message)
         {
             var logLevelString = GetRightPaddedLogLevelString(logLevel);
-            return $"{logLevelString}: [{_name}] {message}";
+            return $"{logLevelString}: [{logName}] {message}";
         }
 
         public bool IsEnabled(LogLevel logLevel)


hishamco:
@prasannavl  why we need to make ```FormatMessage``` ```protected virtual``` while there no classes inherit from ```ConsoleLogger```, if your aim to add useful information like time and so on, i advice you to have a look to the sample here https://github.com/prasannavl/Logging/blob/961949433d18e0552e83945167763114a830d590/samples/SampleApp/Program.cs

prasannavl:
@hishamco, 

> Why FormatMessage has been renames while it formats the message rather than get LogLevel string?!!

There are no renames. I'm afraid you probably mixed something up, or missed something.

> Unexpected error by shifting FormatMessage declaration inside another method?!!

Builds perfectly for me.

> @prasannavl I think this commit contains unexpected errors

I don't think so. And its very unlikely such a small change actually breaks something.


> @prasannavl why we need to make FormatMessage protected virtual while there no classes inherit from ConsoleLogger, if your aim to add useful information like time and so on, i advice you to have a look to the sample here

The example you gave cannot prefix info like those. It simply uses them as a part of the log message itself. The only way to do it would be to reimplement the whole `ConsoleLogger` again, which is unnecessary effort, and pointless duplication, when a protected method could easily help `let` other classes derive from it with no disadvantage, or side effects.

hishamco:
Look to the line 101 in https://github.com/prasannavl/Logging/commit/961949433d18e0552e83945167763114a830d590#commitcomment-11831027, perhaps you fix it in the next commit

hishamco:
Can you provide an example for what you want to achieve?

prasannavl:
> Look to the line 101 in prasannavl@9619494#commitcomment-11831027, perhaps you fix it in the next commit

Hmm. Odd.

**First commit:**

But it doesn't rename anything. It merely abstracts the LogLevel detection to another protected method, so that it can be used from the derived classes, and makes the FormatMessage protected virtual. 

**Second commit:**

Renames the abstracted `GetLogLevelString` to `GetDefaultLogLevelString`, and removes the virtual on it, since `virtual`. 

Eg: 

```
    public class CustomConsoleLogger : ConsoleLogger
    {
        public ConsoleLoggerWithTime(string name, Func<string, LogLevel, bool> filter) : base(name, filter) { }
        protected override string FormatMessage(LogLevel logLevel, string message)
        {
            var res = base.FormatMessage(logLevel, message);
            return DateTime.Now + " => [More random prefix] => " + res + " => more suffix";
        }
    }
```

hishamco:
So, you wanna do something like log layout in NLog, Am I right?

prasannavl:
That is one potential feature this tiny change opens the door to without any disadvantages, Yes.

hishamco:
I think it will be nice if we get rid off ```FormatMessage``` and let the user handles the formatted messages (String Interpolation) via  ```LoggerExtensions``` methods which accept the ```message`` and ```formatter```, also this will help us in the future if we have a plan to support sort of layout by passing special tokens into the log message itself 

**"[{LogLevel}] This is a simple log @ {DateTime.Now.ToString()}"**

prasannavl:
I agree. But the LogLevel, and message will have to exposed to the formatter somehow, which currently is being exposed by `FormatMessage` method. So, one could possibly move the logic into the extension method from the `FormatMessage`. And since this change doesn’t interfere with any of the core logger interfaces, but merely the `ConsoleLogger` class, I see no side-effects of having it virtual, while you the method exists. 

 

Today, the `ConsoleLogger` gives only two options - use-as-is, or do-everything-yourself, and hence my commit. Having the ability to override `FormatMessage` should take care of most scenarios when it comes to the concrete class of ConsoleLogger.

 

Regards,

Prasanna V. Loganathar

muratg:
Do we expect developers to extend `ConsoleLogger` or implement an `ILogger` for their own needs? @lodejard @davidfowl @kichalla

davidfowl:
This change looks fine.

davidfowl:
We should however avoid optional parameters in public APIs

hishamco:
@muratg the suggestion in #212 may solve format issue for all the logger providers

prasannavl:
> We should however avoid optional parameters in public APIs

@davidfowl, well, I could quite simply remove the default arguments. Would you prefer I do that?

> We should however avoid optional parameters in public APIs

If I may - on just an educational note, is there a reason to avoid optional parameters in public APIs - or is it just an internal design principle?

davidfowl:
@prasannavl http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

muratg:
@hishamco #212 is in the backlog, so we'll potentially look at it post V1. A change like that is more substantial.

@kichalla Can you take this in after @prasannavl removes the optional param?

prasannavl:
@davidfowl - Thanks. That was informative :)

@muratg - Well, I actually removed the whole thing, and move the padding into `FormatMessage` alleviating the need for it at all.

muratg:
@kichalla :shipit;

kichalla:
@prasannavl Can you rebase your PR on the latest changes on `dev`?

prasannavl:
@kichalla - Done.

kichalla:
@prasannavl We try to keep the commit graph as linear as possible. Looks like you merged dev branch onto your feature branch. I would suggest to either open a new PR with just your change or fix this PR. 
Just FYI...when we get a signoff, we usually do the following: 
- squash your list of commits
- checkout the dev branch and sync it up to the latest
- checkout featurebranch and rebase on top of dev and resolve any conflicts
- build feature branch to see if its still successful and push this updated featurebranch to Github

(the following might not be applicable to you in this scenario)
- checkout dev branch and merge featurebranch
- push the dev branch to Github

prasannavl:
@kichalla, Sorry for delay in cleaning up the convoluted tree. Made a new commit on top the latest dev. Cheers! :)

muratg:
@prasannavl Thanks! @kichalla Could you review and merge?  Thanks!

muratg:
:watch:

muratg:
@kichalla Could you please merge?  Thanks.

kichalla:
Commit: 5f98617bf36cd7680e02b4d7be8f2824aba190f9
Thanks for your contribution @prasannavl

src/Microsoft.Framework.Logging.Console/ConsoleLogger.cs
```diff
@@ -73,10 +73,10 @@ public void Log(LogLevel logLevel, int eventId, object state, Exception exceptio
             }
         }
 
-        private string FormatMessage(LogLevel logLevel, string message)
+        public virtual string FormatMessage(LogLevel logLevel, string logName, string message)
```
moander: Should be protected instead of public

hishamco:
@prasannavl did you subclass the `ConsoleLogger` and used the new logger, coz I didn't the same things, seems the overridden method doesn't caught at all

PR #200, opened 2015-06-24
Remove extra space

hishamco:


muratg:
I believe this is by design.

hishamco:
@muratg It only used in this line https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/LogFormatter.cs#L101 I don't think it by design, if it is so, i think it will be nice if we rename the constant to ```spaces``` instead of ```space```

Issue #171, opened 2015-04-09
Remove ReflectionBasedLogValues type and instead use Custom formatted log values


davidfowl:
This should probably be shared source somewhere.

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/ReflectionBasedLogValues.cs

muratg:
Move to Common.

kichalla:
commit 3912fc89c8217cd89e6e6568b63be260f2587032

PR #205, opened 2015-06-29
Remove ReflectionBasedLogValues type

kichalla:
Fixes issue : https://github.com/aspnet/Logging/issues/171

The original(if you look at the renaming of the title for the above issue) issue mentioned about moving this type to an assembly which does not depend on DependencyInjection.

The type [FormattedLogValues](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/Internal/FormattedLogValues.cs) is actually part of Logging.Abstractions and does not have dependency on DependencyInjection.

Also I checked that ReflectionBasedLogValues is not being used in any other repos.
https://github.com/search?utf8=%E2%9C%93&q=user%3Aaspnet+ReflectionBasedLogValues&type=Code&ref=searchresults

/cc: @muratg @Eilon


muratg:
@kichalla is `ReflectionBasedLogValues` used in other repos? If so, you'll need to change those similar to the tests. Building Universe with this change would find most of them (but probably not all.)

muratg:
If all good, :shipit:

kichalla:
@muratg No, it's not being used in any other repos : https://github.com/search?utf8=%E2%9C%93&q=user%3Aaspnet+ReflectionBasedLogValues&type=Code&ref=searchresults

Eilon:
:shipit:

PR #206, opened 2015-06-29
Map LogLevels to a string using a jump table instead of a dictionary

justinvp:
- This is ~22x faster than the dictionary approach as the C# compiler will use a jump table for the switch statement
- The strings are pre-padded to avoid unnecessary intermediate string allocations
- Minor cleanup:
 - Remove unused usings
 - The `_filter` field can be `readonly`

/cc: @kichalla @muratg


Eilon:
Looks :shipit: to me.

@kichalla, looks good to you?

kichalla:
:shipit:

kichalla:
Commit 400e191c33d0d2d09f9be23e3b3fa15fa098e3c4

Thanks! @justinvp

Issue #194, opened 2015-06-15
Add a DebugLoggerProvider that writes to Debug.WriteLine


DamianEdwards:
Would be almost identical to the existing `ConsoleLoggerProvider` but it would write to `Debug.WriteLine`. It would enable itself only when the debugger is attached.

We should have this provider in the runtime by default so that any app will log messages to the debugger output, e.g. in Visual Studio.

DamianEdwards:
Something like the one at https://github.com/jongalloway/ndc2015-aspnet5-workshop/blob/master/Labs/4%20-%20Dependency%20Injection%20%26%20Unit%20Testing.md#configure-a-logger-so-you-can-see-the-request-id-messages

hishamco:
It will be nice if the logger will be added automatically by the runtime if the debugger is attached
:+1:

muratg:
We think this can be done in DNX instead of a separate package

muratg:
@kichalla We can potentially do this in DNX by bringing logging abstractions to DNX layer when you bring ILogger to DNX.  

https://github.com/aspnet/dnx/issues/1118

davidfowl:
i would avoid the dnx for now and keep it localized to hosting

DamianEdwards:
@davidfowl any particular reason? Because of the churn we're expecting in the DNX? We have other issues RE bringing the `LoggerFactory` down to DNX so it seems odd to not do it all at once.

davidfowl:
Putting things int the dnx requires care in order to avoid pinning before the application loads. That plus the currently refactoring going on it seems unwise to make that layering decision at this point. We need an easy way to move things without pinning it before the app runs

DamianEdwards:
If that's the case why don't we just put it in the templates only? Hosting still isn't really optional for apps anyway so maybe better to require it to be added by the app explicitly.

davidfowl:
+1. We should at least do the feature for beta6 it's trivial. We can discuss moving it in by default at a later date.

DamianEdwards:
K. What package should we put it in? Microsoft.Framework.Logging?

davidfowl:
Separate package like the other providers

DamianEdwards:
And you get a package! And you get a package! You all get packages!

PureKrome:
![Developers Developers Developers!!!](http://upboat.me/att/packages/for-all-the-developers!.jpg)

victorhurdugaci:
[...]

Issue #201, opened 2015-06-24
Flat File Logger Provider


hishamco:
It will be nice if we provide a logger that logs and persist into files such as text, xml, json .. etc
the implementation will something like
```csharp
public abstract class FileLogger : ILogger
{
    private readonly string _name;

    public FileLogger(string name)
    {
        _name = name;
    }

    public virtual IDisposable BeginScopeImpl(object state)
    {
        return null;
    }

    public virtual bool IsEnabled(LogLevel logLevel)
    {
        return false;
    }

    public abstract void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter);
}
```
After that we can create an implementation of ```FileLogger```

```csharp
public class JsonFileLogger : FileLogger
{
    public override void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
    {
        ...
    }
}
```

muratg:
We don't plan to ship more logging implementations at this point. We recommend using a 3rd party logger (like NLog or Serilog) for this type of situation.

PR #207, opened 2015-06-30
Debug logger

victorhurdugaci:
Fixes https://github.com/aspnet/Logging/issues/194

Please review @davidfowl @DamianEdwards @mikaelm12

### DIFF
--- a/src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
+++ b/src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
@@ -1,0 +1,77 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Framework.Logging.Debug
+{
+    public class DebugLogger : ILogger
+    {
+        private readonly Func<string, LogLevel, bool> _filter;
+        private readonly string _name;
+
+        public DebugLogger(string name)
+            : this(name, filter: null)
+        {
+        }
+
+        public DebugLogger(string name, Func<string, LogLevel, bool> filter)
+        {
+            _name = string.IsNullOrEmpty(name) ? nameof(DebugLogger) : name;
+            _filter = filter ?? ((category, logLevel) => true);
+        }
+
+        public IDisposable BeginScopeImpl(object state)
+        {
+            return null;
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            if (!Debugger.IsAttached)
+            {
+                return false;
+            }
+
+            return _filter(_name, logLevel);
+        }
+
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            string message;
+            var values = state as ILogValues;
+            if (formatter != null)
+            {
+                message = formatter(state, exception);
+            }
+            else if (values != null)
+            {
+                message = LogFormatter.FormatLogValues(values);
+                if (exception != null)
+                {
+                    message += Environment.NewLine + exception;
+                }
+            }
+            else
+            {
+                message = LogFormatter.Formatter(state, exception);
+            }
+
+            if (string.IsNullOrEmpty(message))
+            {
+                return;
+            }
+
+            message = $"{ logLevel }: {message}";
+            System.Diagnostics.Debug.WriteLine(message, _name);
+        }
+    }
+}
--- a/src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
+++ b/src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
@@ -1,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Logging.Debug;
+
+namespace Microsoft.Framework.Logging
+{
+    public static class DebugLoggerFactoryExtensions
+    {
+        /// <summary>
+        /// Adds a debug logger that is enabled for <see cref="LogLevel"/>.Information or higher.
+        /// </summary>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory)
+        {
+            return AddDebug(
+                factory,
+                (category, logLevel) => logLevel >= LogLevel.Information);
+        }
+
+        /// <summary>
+        /// Adds a debug logger that is enabled as defined by the filter function.
+        /// </summary>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, Func<string, LogLevel, bool> filter)
+        {
+            factory.AddProvider(new DebugLoggerProvider(filter));
+            return factory;
+        }
+
+        /// <summary>
+        /// Adds a debug logger that is enabled for <see cref="LogLevel"/>s of minLevel or higher.
+        /// </summary>
+        /// <param name="minLevel">The minimum <see cref="LogLevel"/> to be logged</param>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, LogLevel minLevel)
+        {
+            return AddDebug(
+               factory,
+               (category, logLevel) => logLevel >= minLevel);
+        }
+    }
+}
--- a/src/Microsoft.Framework.Logging.Debug/DebugLoggerProvider.cs
+++ b/src/Microsoft.Framework.Logging.Debug/DebugLoggerProvider.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Framework.Logging.Debug
+{
+    public class DebugLoggerProvider : ILoggerProvider
+    {
+        private readonly Func<string, LogLevel, bool> _filter;
+
+        public DebugLoggerProvider(Func<string, LogLevel, bool> filter)
+        {
+            _filter = filter;
+        }
+
+        public ILogger CreateLogger(string name)
+        {
+            return new DebugLogger(name, _filter);
+        }
+    }
+}
--- a/src/Microsoft.Framework.Logging.Debug/project.json
+++ b/src/Microsoft.Framework.Logging.Debug/project.json
@@ -1,0 +1,20 @@
+﻿{
+    "version": "1.0.0-*",
+    "dependencies": {
+        "Microsoft.Framework.Logging.Abstractions": "1.0.0-*"
+    },
+
+    "frameworks": {
+        "net45": {
+        },
+        "dnx451": {
+        },
+        "dotnet": {
+            "dependencies": {
+                "System.Collections": "4.0.10-beta-*",
+                "System.Globalization": "4.0.10-beta-*",
+                "System.Linq": "4.0.0-beta-*"
+            }
+        }
+    }
+}
--- a/Logging.sln
+++ b/Logging.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 14
-VisualStudioVersion = 14.0.22823.1
+VisualStudioVersion = 14.0.23017.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.Logging", "src\Microsoft.Framework.Logging\Microsoft.Framework.Logging.xproj", "{19D1B6C5-8A62-4387-8816-C54874D1DF5F}"
 EndProject
@@ -22,6 +22,8 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{699DB330-0095-4266-B7B0-3EAB3710CA49}"
 EndProject
 Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.Logging.TraceSource", "src\Microsoft.Framework.Logging.TraceSource\Microsoft.Framework.Logging.TraceSource.xproj", "{1A3EB66F-9E64-4676-852F-24995549ED8A}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "Microsoft.Framework.Logging.Debug", "src\Microsoft.Framework.Logging.Debug\Microsoft.Framework.Logging.Debug.xproj", "{FFEDC225-D5BD-44E0-B7A6-A98FCFECC694}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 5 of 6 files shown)


src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+            {
+                return;
+            }
+
+            string message;
+            var values = state as ILogValues;
+            if (formatter != null)
```
mikaelm12: Can you use the{ NotNull] attribute instead?

DamianEdwards: That attribute throws if the value is null, here it is actually optional.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                message = formatter(state, exception);
+            }
+            else if (values != null)
+            {
+                var formattedMessage = new StringBuilder();
+                FormatLogValues(formattedMessage, values);
+                message = formattedMessage.ToString();
```
kichalla: curious, was it intentional that you excluded writing out `exception` because in debug mode one could break at exceptions?...like there are scenarios where exceptions are swallowed but a message with exception is still logged...so probably we want to include exception too?

DamianEdwards: I think the assumption is that it's always one or the other. This logic should be similar to the existing console logger.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+            else
+            {
+                message = LogFormatter.Formatter(state, exception);
+            }
+
+            if (message == null)
```
kichalla: IsNullOrEmpty?

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+            {
+                return;
+            }
+
+            System.Diagnostics.Debug.WriteLine(message, _name);
```
kichalla: intentional that you excluded writing out `loglevel` as a prefix since we are in `debug` mode?

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        private void FormatLogValues(StringBuilder builder, ILogValues logValues, int level = 1)
```
kichalla: can we pull this out to a common location to use in both Console logger and here?

kichalla: Probably we should have something like an abstract text logger which both Console and this one inherit from...like there was a request(https://github.com/aspnet/Logging/issues/201) from a user about a flat file logger in which case they could derive from this type...

src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.Framework.Logging
+{
+    public static class DebugLoggerFactoryExtensions
+    {
+        /// <summary>
+        /// Adds a debug logger that is enabled for <see cref="LogLevel"/>.Information or higher.
+        /// </summary>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory)
+        {
+            factory.AddProvider(new DebugLoggerProvider((category, logLevel) => logLevel >= LogLevel.Information));
```
kichalla: you can use the other `AddDebug` extension here and in other places below

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
victorhurdugaci: It was not intentional. I will add it. Even if you catch it, it is still nice to have it in the debug output

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
victorhurdugaci: Will add

victorhurdugaci:
Updated

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                message = LogFormatter.Formatter(state, exception);
+            }
+
+            if (string.IsNullOrEmpty(message))
+            {
+                return;
+            }
+
+            message = $"{ logLevel }: {message}";
```
kichalla: take a look at the following recently merged in external PR and see if we should do something like that for log level:
https://github.com/aspnet/Logging/commit/400e191c33d0d2d09f9be23e3b3fa15fa098e3c4

kichalla:
Looks good to me. :shipit:

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+
+        public DebugLogger(string name)
+            : this(name, filter: null)
+        {
+        }
+
+        public DebugLogger(string name, Func<string, LogLevel, bool> filter)
+        {
+            _name = string.IsNullOrEmpty(name) ? nameof(DebugLogger) : name;
+            _filter = filter ?? ((category, logLevel) => true);
```
DamianEdwards: I think we should just leave `_filter` as `null` and check for it in `IsEnabled` rather than assign a default `true` returning delegate. Why run anything at all when checking `IsEnabled` if we can just check for `null`?

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            if (!Debugger.IsAttached)
+            {
+                return false;
+            }
+
+            return _filter(_name, logLevel);
```
DamianEdwards: As per comment above:
``` C#
return _filter == null || _filter(_name, logLevel);
```

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
DamianEdwards: nit: extra whitespace in the placeholders here.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+                message = LogFormatter.Formatter(state, exception);
+            }
+
+            if (string.IsNullOrEmpty(message))
+            {
+                return;
+            }
+
+            message = $"{ logLevel }: {message}";
+            System.Diagnostics.Debug.WriteLine(message, _name);
```
DamianEdwards: No need to full qualify, `System.Diagnostics` is already imported at top of file.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Framework.Logging.Debug
+{
+    public class DebugLogger : ILogger
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+    {
+        private readonly Func<string, LogLevel, bool> _filter;
+        private readonly string _name;
+
+        public DebugLogger(string name)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+            : this(name, filter: null)
+        {
+        }
+
+        public DebugLogger(string name, Func<string, LogLevel, bool> filter)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            _name = string.IsNullOrEmpty(name) ? nameof(DebugLogger) : name;
+            _filter = filter ?? ((category, logLevel) => true);
+        }
+
+        public IDisposable BeginScopeImpl(object state)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            return null;
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            if (!Debugger.IsAttached)
+            {
+                return false;
+            }
+
+            return _filter(_name, logLevel);
+        }
+
+        public void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)
```
DamianEdwards: Add doc comments (can possibly just `/// <inheritdoc />`)

src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Microsoft.Framework.Logging.Debug;
+
+namespace Microsoft.Framework.Logging
+{
+    public static class DebugLoggerFactoryExtensions
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            return AddDebug(
+                factory,
+                (category, logLevel) => logLevel >= LogLevel.Information);
+        }
+
+        /// <summary>
+        /// Adds a debug logger that is enabled as defined by the filter function.
+        /// </summary>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, Func<string, LogLevel, bool> filter)
```
DamianEdwards: Add doc comments for parameters.

src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+using Microsoft.Framework.Logging.Debug;
+
+namespace Microsoft.Framework.Logging
+{
+    public static class DebugLoggerFactoryExtensions
+    {
+        /// <summary>
+        /// Adds a debug logger that is enabled for <see cref="LogLevel"/>.Information or higher.
+        /// </summary>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory)
```
DamianEdwards: Add doc comments for parameters.

src/Microsoft.Framework.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        /// <summary>
+        /// Adds a debug logger that is enabled for <see cref="LogLevel"/>s of minLevel or higher.
+        /// </summary>
+        /// <param name="minLevel">The minimum <see cref="LogLevel"/> to be logged</param>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, LogLevel minLevel)
+        {
+            return AddDebug(
+               factory,
+               (category, logLevel) => logLevel >= minLevel);
```
DamianEdwards: nit: rename `category` to `_` as it's unused

src/Microsoft.Framework.Logging.Debug/DebugLoggerProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Framework.Logging.Debug
+{
+    public class DebugLoggerProvider : ILoggerProvider
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLoggerProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+    {
+        private readonly Func<string, LogLevel, bool> _filter;
+
+        public DebugLoggerProvider(Func<string, LogLevel, bool> filter)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLoggerProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+        {
+            _filter = filter;
+        }
+
+        public ILogger CreateLogger(string name)
```
DamianEdwards: Add doc comments.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public IDisposable BeginScopeImpl(object state)
+        {
+            return null;
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            if (!Debugger.IsAttached)
```
davidfowl: This doesn't seem correct

DamianEdwards: Yeah this seems like it could be simplified to:
```
return Debugger.IsAttached && (_filter == null || _filter(_name, logLevel))
```

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
DamianEdwards: @kichalla that doesn't make as much sense here as we're passing the `_name` through to the underlying `Debug.WriteLine` call as the category, so we're not controlling the left-hand side "column" width like we do in the console logger.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
victorhurdugaci: I knew I should have put a comment. `Debug` conflicts with the last part of this class' namespace `Microsoft.Framework.Logging.Debug` and it is not resolved as a class name.

DamianEdwards: Ah got it

Eilon: :exclamation: Ummmmm.......... this won't work, like ever, will it? `Debug.WriteLine` has `[Conditional("DEBUG")]` on it and won't even be in the code when we compile release, which is what we should be doing.

I think @davidfowl and I once came up with a trick to make this work by putting the code in a separate file and `#undef`ing the `DEBUG` symbol unconditionally in that one file. I think if you do it with a `partial class` you can make it be ugly-but-not-horrible.

src/Microsoft.Framework.Logging.Debug/project.json
```diff
@@ -0,0 +1,20 @@
+﻿{
+    "version": "1.0.0-*",
+    "dependencies": {
```
Eilon: Add a description to this.

src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
kichalla: > Debug.WriteLine has [Conditional("DEBUG")] on it and won't even be in the code when we compile release, which is what we should be doing.

@Eilon for my own understanding, this logger is supposed(like the template will probably add this provider within the `Development` environment) to be used during development time and when in Debug mode and so curious as to why Release matters?

kichalla: Ah, I think I got what you are saying...just tried in a sample app(in debug mode) referencing this package and since this package is built for Release, my log statement never spitted out the log statements(probably because of the reason you mentioned)..am i right?

Eilon: Yes, but *this* package is always compiled in release mode, and will *never* have the call to `Debug.WriteLine` in it. Completely orthogonal to what the *user's* project is doing.

Eilon: Yes

kichalla: Yeah, I was initially misled into believing this works as it is when I used the sample app present in Logging solution's samples and because the whole solution is built for debug it seemed working as it supposed to.

PR #213, opened 2015-07-06
Make TraceSourceScope internal

justinvp:
I understand the reasoning behind exposing some types publicly under an `*.Internal` namespace (so its available to "use at your own risk"), but I don't think `TraceSourceScope` warrants being public. It's not providing any new functionality that can't be called directly and an instance can be accessed through `ILogger.BeginScopeImpl()`.

### DIFF
--- a/test/Microsoft.Framework.Logging.Test/TraceSourceScopeTest.cs
+++ b/test/Microsoft.Framework.Logging.Test/TraceSourceScopeTest.cs
@@ -3,9 +3,6 @@
 
 using System.Diagnostics;
 using Xunit;
-#if DNX451
-using Moq;
-#endif
 
 namespace Microsoft.Framework.Logging.Test
 {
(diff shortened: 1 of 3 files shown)


test/Microsoft.Framework.Logging.Test/TraceSourceScopeTest.cs
```diff
@@ -6,7 +6,6 @@
 #if DNX451
```
Eilon: This file doesn't even use Moq... can we remove this `using` and get rid of all the `#if`s in this file?

Eilon:
Seems fine to me.

test/Microsoft.Framework.Logging.Test/TraceSourceScopeTest.cs
justinvp: I noticed that too (but held off on removing it to not clutter this PR with unrelated changes).

I'll go ahead and add a commit to this PR that removes the `#if` around `using Moq;`.

I'll keep the other `#if` because the implementation of `TraceSourceScope` currently has `#if`s in its implementation around the calls to `Trace.CorrelationManager` and I haven't had a chance to investigate if those can be safely removed.

Eilon: O I C, thanks!

PR #214, opened 2015-07-06
Remove System.Diagnostics.TraceSource dependency

bricelam:
It was unused and prevented the package from installing on UWP projects.

@rowanmiller @natemcmaster


Eilon:
:shipit: - the TraceSource stuff was just split out into a separate package, so this is definitely no longer needed.

Issue #216, opened 2015-07-08
DebugLogger nuget package doesn't contain Debug.WriteLine


ravengerUA:
Microsoft.Framework.Logging.Debug nuget package on Myget doesn't contain Debug.WriteLine() in the code.

davidfowl:
That's right. it's in this package https://www.nuget.org/packages?q=System.Diagnostics.Debug. The System* packages are where the BCL types exist.

ravengerUA:
I'm talking about this line
````C#
            if (string.IsNullOrEmpty(message))
            {
                return;
            }

            message = $"{ logLevel }: {message}";
            * System.Diagnostics.Debug.WriteLine(message, _name); *
````

Debug.WriteLine marked with ConditionalAttribute and won't go to compiled code, when you compile and publish this code to nuget packages.

kichalla:
@davidfowl https://github.com/aspnet/Logging/pull/207/files#r33876417

Eilon:
Yes, this never worked at all due to the comment @kichalla pointed out.

Eilon:
@victorhurdugaci can you take a look at this for beta 6?

PR #217, opened 2015-07-13
Define the DEBUG symbol on Release

victorhurdugaci:
Fixes https://github.com/aspnet/Logging/issues/216

@kichalla @Eilon

### DIFF
--- a/src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
+++ b/src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Framework.Logging.Debug
     /// <summary>
     /// A logger that writes messages in the debug output window only when a debugger is attached.
     /// </summary>
-    public class DebugLogger : ILogger
+    public partial class DebugLogger : ILogger
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.Logging.Debug/DebugLogger.cs
```diff
@@ -1,6 +1,11 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+// We need to define the DEBUG symbol because we want the logger
+// to work even when this package is compiled on release. Otherwise,
+// the call to Debug.WriteLine will not be in the release binary
+#define DEBUG
```
kichalla: I believe @Eilon suggestion of having this defined in a separate file using partial class was to limit the scope of the code which sees this symbol defined. You can create a partial class with a private method like `WriteLog` having just the debug writeline statement and the DEBUG defined.

victorhurdugaci:
Updated

kichalla:
:shipit:

Eilon:
Yup, this is the evil trick! :shipit:

PR #224, opened 2015-07-30
[Fixes #208] Exception in a logger stops execution of other registere…

### LINKED ISSUE #208: Exception in a logger stops execution of other registered loggers
We should continue calling into other loggers even though one of the loggers throws an exception.

Source:
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/Logger.cs#L34
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging/Logger.cs#L47

Repro sample:
```c#
using System;
using Microsoft.Framework.Logging;

namespace TestApp
{
    public class Program
    {
        public void Main(string[] args)
        {
            loggerFactory.AddProvider(new CustomLoggerProvider());
            loggerFactory.AddConsole(LogLevel.Information);

            var logger = loggerFactory.CreateLogger<Program>();
            logger.LogInformation("Hello!");
        }
    }

    public class CustomLoggerProvider : ILoggerProvider
    {
        public ILogger CreateLogger(string name)
        {
            return new CustomLogger(name);
        }
    }

    public class CustomLogger : ILogger
    {
        private readonly string _name;

        public CustomLogger(string name)
        {
            _name = name;
        }
        public IDisposable BeginScopeImpl(object state)
        {
            return null;
        }

        public bool IsEnabled(LogLevel logLevel)
        {
            return true;
        }

        public void Log(
            LogLevel logLevel, int eventId, object state, 
            Exception exception, Func<object, Exception, string> formatter)
        {
            throw new Exception("Failed to connect to database or a file was not found etc.");
        }
    }
}

```

kichalla:
…d loggers

@muratg

### DIFF
--- a/src/Microsoft.Framework.Logging/Logger.cs
+++ b/src/Microsoft.Framework.Logging/Logger.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Microsoft.Framework.Logging
@@ -29,9 +30,28 @@ namespace Microsoft.Framework.Logging
         {
             if (logLevel >= _loggerFactory.MinimumLevel)
             {
+                List<Exception> exceptions = null;
                 foreach (var logger in _loggers)
                 {
-                    logger.Log(logLevel, eventId, state, exception, formatter);
+                    try
+                    {
+                        logger.Log(logLevel, eventId, state, exception, formatter);
+                    }
+                    catch (Exception ex)
+                    {
+                        if (exceptions == null)
+                        {
+                            exceptions = new List<Exception>();
+                        }
+
+                        exceptions.Add(ex);
+                    }
+                }
+
+                if (exceptions != null && exceptions.Count > 0)
+                {
+                    throw new AggregateException(
+                        message: "An error occurred while writing to logger(s).", innerExceptions: exceptions);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Framework.Logging/Logger.cs
```diff
@@ -31,7 +31,14 @@ public void Log(LogLevel logLevel, int eventId, object state, Exception exceptio
...
                 foreach (var logger in _loggers)
                 {
-                    logger.Log(logLevel, eventId, state, exception, formatter);
+                    try
+                    {
+                        logger.Log(logLevel, eventId, state, exception, formatter);
+                    }
+                    catch
+                    {
+                        // ignore exceptions thrown by a logger and continue calling other loggers
```
Eilon: Should we log if there's an exception? Just kidding. But this kind of freaks me out...

Eilon: Can we instead ensure that we log to all the loggers but still fail if any of them failed?

muratg: This sounds like a good idea to me.

Eilon: Need to discuss with @lodejard and @DamianEdwards because I'm not quite sure what the best behavior is. It might be best to let the whole system go down in flames. My concern is that if a logger is being used for business purposes or logging important security information (e.g. a failed audit) and that logger throws an exception, should the app continue to run as if nothing happened? Or should the one bad logger be ignored? What if that was the only logger and now the app goes silent?

kichalla: @Eilon I update the PR now after discussing with @lodejard . The change(in latest commit) here is to aggregate exceptions from all loggers and throw one `AggregateException`. Please take a look.

cc @muratg

Eilon: Thanks, that looks much nicer!

src/Microsoft.Framework.Logging/Logger.cs
```diff
@@ -29,9 +30,28 @@ public void Log(LogLevel logLevel, int eventId, object state, Exception exceptio
...
-                    logger.Log(logLevel, eventId, state, exception, formatter);
+                    try
+                    {
+                        logger.Log(logLevel, eventId, state, exception, formatter);
+                    }
+                    catch (Exception ex)
+                    {
+                        if (exceptions == null)
+                        {
+                            exceptions = new List<Exception>(_loggers.Length);
```
Eilon: Why initialize the length of the list? That assumes that *all* loggers will throw, and pre-allocates the list to have that many slots. I would think the common case is that if a logger fails, it's just that one failure.

Eilon: (And that either way, the perf of this doesn't matter.)

kichalla: Sure, I was thinking of Perf here but probably not important here. Updated now

Eilon:
:shipit:

PR #147, opened 2015-03-07
Initial prototype of Windows Event Log logger

JesperTreetop:
As requested by @davidfowl for use in issue #83. Built using Visual Studio 2015 CTP 6.

PR #225, opened 2015-07-30
Logger for Windows Event Log

kichalla:
@Eilon

- This PR does not reflect all the details of the PR : https://github.com/aspnet/Logging/pull/147/ as it has details like HttpContext(which is incorrect as the logger can be used outside the context of aspnet) and creation of event source(which requires admin privileges and is not really a job of this logger).

### DIFF
--- a/samples/SampleApp/project.json
+++ b/samples/SampleApp/project.json
@@ -9,6 +9,7 @@
     "frameworks": {
         "dnx451": {
             "dependencies": {
+                "Microsoft.Framework.Logging.EventLog": "1.0.0-*",
                 "Microsoft.Framework.Logging.NLog": "1.0.0-*"
             }
         },
--- a/src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
+++ b/src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
@@ -1,0 +1,126 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace Microsoft.Framework.Logging.EventLog
+{
+    /// <summary>
+    /// A logger that writes messages to Windows Event Log.
+    /// </summary>
+    public class EventLogLogger : ILogger
+    {
+        private readonly System.Diagnostics.EventLog _eventLog;
+        private readonly string _name;
+        private readonly EventLogSettings _settings;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventLogLogger"/> class.
+        /// </summary>
+        /// <param name="name">The name of the logger.</param>
+        public EventLogLogger(string name)
+            : this(name, settings: new EventLogSettings())
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EventLogLogger"/> class.
+        /// </summary>
+        /// <param name="name">The name of the logger.</param>
+        /// <param name="settings">The <see cref="EventLogSettings"/>.</param>
+        public EventLogLogger(string name, EventLogSettings settings)
+        {
+            _name = string.IsNullOrEmpty(name) ? nameof(EventLogLogger) : name;
+            _settings = settings;
+
+            var logName = string.IsNullOrEmpty(settings.LogName) ? "Application" : settings.LogName;
+            var sourceName = string.IsNullOrEmpty(settings.SourceName) ? "Application" : settings.SourceName;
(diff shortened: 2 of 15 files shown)


src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
```diff
@@ -0,0 +1,119 @@
...
+            // 1. Log name & source name existence check only works on local computer.
+            // 2. Source name existence check requires Administrative privileges.
+
+            _eventLog = new System.Diagnostics.EventLog(logName, machineName, sourceName);
+        }
+
+        /// <inheritdoc />
+        public IDisposable BeginScopeImpl(object state)
+        {
+            return null;
```
kichalla: This logger is different than the Console and Debug in the sense that these loggers are more suitable for development time where as EventLog could be used in production environments too...in this case the scoping is interesting as for example, you might want to correlate messages between different requests...even if we capture this scoping information and write to the log, I couldn't find info on the web about how to structure this data in order to make this searchable...I guess for scenarios like these users have to look for other options like elasticsearch, mongodb etc....this makes me think if we should even support an out of the box provider?

samples/SampleApp/project.json
```diff
@@ -9,7 +9,8 @@
     "frameworks": {
         "dnx451": {
             "dependencies": {
-                "Microsoft.Framework.Logging.NLog": "1.0.0-*"
+                "Microsoft.Framework.Logging.NLog": "1.0.0-*",
+                "Microsoft.Framework.Logging.EventLog": "1.0.0-*"
```
Eilon: Sort :trollface:

src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
Eilon: That's a good question, I'm not sure what the answer is. I think that maybe the main idea for this logger is because it's super easy to get to on Windows - nothing special has to be installed or set up on the machine. So maybe having *some* logging is better than none...

Eilon:
Overall the code looks fine to me.

src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
PinpointTownes: Should a no-op `IDisposable` be returned here for the edge case where one wouldn't use `BeginScopeImpl` with `using` and would manually call `Dispose`?

I'm probably paranoid, but I guess it would be better to make all loggers consistent.

src/Microsoft.Framework.Logging.EventLog/project.json
```diff
@@ -0,0 +1,16 @@
+﻿{
+    "version": "1.0.0-*",
+    "repository": {
+        "type": "git",
+        "url": "git://github.com/aspnet/logging"
+    },
+    "description": "Logger for Windows Event Log.",
+    "dependencies": {
+        "Microsoft.Framework.Logging.Abstractions": "1.0.0-*",
+        "Microsoft.Framework.NotNullAttribute.Sources": "1.0.0-*"
```
PinpointTownes: Nice. Has the syntax been simplified? (IIRC, the previous syntax was `"Microsoft.Framework.NotNullAttribute.Sources": { "type": "build", "version": "1.0.0-*" }`)

Eilon: Yes, it's been simplified here but it's wrong :smile: It does need the `type: build` annotation. Good catch!

davidfowl:
Can you paste a screenshot of some samples errors in the event log?

src/Microsoft.Framework.Logging.EventLog/project.json
kichalla: good catch!

kichalla:
@davidfowl
image

Default behavior without user modifying any settings(The event id has to be greater than 0 and also a custom event source should be provided to see the message as in the above image):
image

src/Microsoft.Framework.Logging.EventLog/EventLogLogger.cs
kichalla: @PinpointTownes good point...i will fix this and other loggers also.

kichalla:
Updated.

kichalla:
Updated @Eilon

Eilon:
:shipit:

davidfowl:
@kichalla Where is the category information stored?

kichalla:
@davidfowl Currently the category information is always `None` as the `ILogger` interface doesn't have a way to provide this information: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Abstractions/ILogger.cs#L22

PR #228, opened 2015-08-06
Microsoft.Framework.Logging.Test -> Microsoft.Framework.Logging.Tests

hishamco:
According to the guidelines https://github.com/aspnet/Home/wiki/Engineering-guidelines#solution-and-project-folder-structure-and-naming, we should rename the test project to ```Microsoft.Framework.Logging.Tests``` instead of ```Microsoft.Framework.Logging.Test```


Eilon:
See https://github.com/aspnet/EventNotification/pull/12#issuecomment-128427434.

PR #229, opened 2015-08-07
Tracked scopes

veccsolutions:
This adds a very useful feature that we use in our logging where we automatically add Enter/Exit messages and timings to the logging so you can do something like

using (_logger.BeginTrackedScope(LogLevel.Information, "Start", "End", true, "Prefix {0} ", "xxx"))
{
}

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -54,6 +54,14 @@ namespace SampleApp
                 Console.ReadLine();
             }
 
+            Console.WriteLine("Hello World - tracked");
+            using (_logger.BeginTrackedScopeImpl("Main - Tracked", LogLevel.Information, "Start", "End", true))
+            {
+                _logger.LogInformation("Waiting for user input");
+                var userInput = Console.ReadLine();
+                _logger.LogInformation("User input: {0}", userInput);
+            }
+
             var endTime = DateTimeOffset.UtcNow;
             _logger.LogInformation(2, "Stopping at '{StopTime}'", endTime);
 
--- a/src/Microsoft.Framework.Logging.Abstractions/ILogger.cs
+++ b/src/Microsoft.Framework.Logging.Abstractions/ILogger.cs
@@ -34,5 +34,16 @@ namespace Microsoft.Framework.Logging
         /// <param name="state">The identifier for the scope.</param>
         /// <returns>An IDisposable that ends the logical operation scope on dispose.</returns>
         IDisposable BeginScopeImpl(object state);
+
+        /// <summary>
+        /// Begins a logical tracked scope. This will track time elapsed in the scope as well as an enter and exit message.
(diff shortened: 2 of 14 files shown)


davidfowl:
This is just a proposal right? We don't take unplanned features like this without consensus.

Eilon:
This seems like something that doesn't need to be built into `ILogger`. It can be built into specific logger implementations that have a need to track timings.

veccsolutions:
@davidfowl - Sorry about that, I didn't know about the consensus stuff.
@Eilon - I put the tracking scope into ILogger to follow the same way that the current scoping piece was implemented.

davidfowl:
I also agree with @Eilon, this isn't something that belong on the ILogger interface. It's an extension method at best and it shouldn't be in the default logging implementation either.

veccsolutions:
There were other issues (at least with the NLog one) that putting this on ILogger allows for preventing.
The callsite is showing the method doing the logging is the incorrect method. Which this pull request does not fix. Putting this inside of ILogger would allow the logger implementation to pass the information needed to NLog (or whatever other logger) to show the correct method signature.
As an extension method you no longer have access to the actual underlying logger to pass that information.

I'm using NLog as the example because that's the one I'm working with, any logger that looks at the stack to determine what method is calling for logging purposes would have this problem.

Does that make sense?

davidfowl:
Not really, if you really wanted to do this, you could have a convention where the data object passed into BeginScope had some very specific fields on it that a logger could respect. The first class method on ILogger is just out of place IMO for something as specific as this.

veccsolutions:
Can you be a little more specific as to what the "first class method" is? The Microsoft.Framework.Logging.Logger:BeginTrackedScopeImpl()?

davidfowl:
I don't believe "tracked scopes" should be a method on ILogger

veccsolutions:
If I was to re-work it so that the BeginScopeImpl is a virtual method, the logger classes public with virtual log methods, and the scopes public (so they can also be inherited) would that work better for you? It would allow for a lot more flexibility so people can build out their own scopes and log methods without duplicating code that's already done.

veccsolutions:
Here's my new proposal with example classes in the sampleapp project. It would be a new pull request, scrapping this one.

https://github.com/veccsolutions/Logging/commit/1a969da6ca945ed1185a4947c27645f79bf58353

Eilon:
I think showing how to do timing in logging on a per-scope basis would make for a good community project. It's currently out of scope (har-har :smile: ) for logging itself, which is focused on reporting app status and behavior.

If there are some specific API changes that need to be made to logging, such as making more methods virtual, I think that could be well within the scope of the project. (It's often hard to predict what should be virtual until a scenario comes up.)

veccsolutions:
I just did a new pull request with making methods virtual and the scopes public. We can close this request.

PR #230, opened 2015-08-10
Virtualized log and beginscopeimpl for easy extension of loggers

veccsolutions:


Eilon:
Do you have an example of how making the methods virtual makes extensibility easier? I want to make sure we have a good grasp of this to ensure that this is the right change.

veccsolutions:
Here you go:

https://github.com/veccsolutions/Logging/commit/9798b9259a23310fd303bbbebf465949d432ef4e

Eilon:
Thanks.

I wonder if there's a simpler way to do this by just wrapping arbitrary loggers. Is there any value to deriving from the NLog logger? I could imagine this code could work just the same - if not *better* - by using encapsulation instead of inheritance. Just take in `ILogger`, scopes, etc. and always call the "inner" object and then do any pre/post work and call into the logging methods again to report enter/exit/timing. Would that work?

veccsolutions:
That would be great, until someone needs to modify how NLog is called to fix an issue with it and/or pass custom/different information into the LogEventInfo properties. Which is what I will end up needing to do here in the near future with the project I'm working on. Wrapping it is a great option as well for most people, if it was just the timed scoping like I was doing at first then wrapping would be fine. Does that help make sense?

muratg:
@lodejard @Eilon Do we want to take this in?  

Also, not sure why `TraceSourceLogger` and `TraceSourceScope` need to be public.  Seems unrelated to making the methods virtual.

Eilon:
I'm still not seeing any difference between changing the code to add `virtual` versus using encapsulation. The entry points are the same either way. Your new method will be the one that's directly called (whether an `override` or a separate method), and then it's just a matter of calling the NLog implementation (either via `base.Foo()` or `_inner.Foo()`).

In my view this change doesn't improve much.

muratg:
Agreed with @Eilon.  This doesn't seem widely applicable for most users and encapsulation seems to work for most cases. Closing this one for now. Thanks @veccsolutions, and feel free to file a bug for discussion if you still have a use case for this and discuss it further.

PR #243, opened 2015-09-12
Fixes issue when logging test fails if console background is not black

dmitry-ra:
TestConsole now has the own implementation of ResetColor() and ConsoleLoggerTest use a TestConsole instead System.Console. This fixes aspnet/Logging#235.


davidfowl:
Thanks!

Issue #239, opened 2015-08-31
LoggerExtensions Can't Combine Exception and Format String


ardalis:
If you want to log:
1) eventId
2) Exception
3) Formatted message

There is currently no extension method allowing this. Can we add one, or should this scenario just use the Log method? Here is the proposed signature:
        public static LogCritical([NotNull] this ILogger logger, int eventId, string format, Exception error, params object[] args)

I just ran into this while trying to log an exception in a catch {} block in which I also wanted to include some details about the request:
                    _logger.LogCritical(eventId,
                        "{eventId}: Unexpected error handling request {path}.",
                        ex, eventId, context.Request.Path);

It doesn't work as desired because the exception (ex) is part of the args array for this overload (so the exception is output in place of {eventId} and the eventId is output in place of {path}).

ardalis:
This seems to be a current workaround - 
                    var values = new FormattedLogValues("{eventId}: Unexpected error handling request {path}:", eventId, context.Request.Path);
                    _logger.LogCritical(eventId, values, ex);

muratg:
Please see https://github.com/aspnet/Logging/issues/331 for the work that we're currently doing to rationalize logger extensions.

PR #244, opened 2015-09-14
LoggingExtensions: Additional methods

thoean:
LoggingExtensions will benefit from being able to log Exceptions and messages with format values separately to allow the underlying logging framework to format the exception differently than the message, for example to support extracting the stack trace for structured logging.

This change resolves aspnet/Logging#239.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
@@ -384,6 +384,18 @@ namespace Microsoft.Extensions.Logging
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


thoean:
Workaround for logging an exception with structured data in the custom error message:

```
var values = new FormattedLogValues("Error during operation {Operation}..", myOperation);
logger.LogError(values, ex);
```

This pull request simplifies this to:

```
logger.LogError(ex, "Error during operation {Operation}..", myOperation);
```

304NotModified:
I think this will be fixed in https://github.com/aspnet/Logging/issues/331

muratg:
Thanks @thoean, we had a design meeting and came out with the design in #331, which is already implemented.  Closing this PR.

PR #253, opened 2015-09-25
Make ILogger covariant

anpete:


Eilon:
@lodejard FYI. This seems fine to me.

Eilon:
Just curious what the reason for this is in terms of scenario? I don't see any problem with this change - just wondering what code would be unblocked with this.

anpete:
```c#
public class Base
{
    private readonly ILogger<Base> _logger;

    public Base(ILogger<Base> logger)
    {
        _logger = logger;
    }
}

private class Derived : Base
{
    public Derived(ILogger<Derived> logger)
        : base(logger)
    {
                
    }
}
```

Eilon:
Ah, cool, so I think for that code you want it to be covariant, right? :smile: 

:shipit:

anpete:
That's what I said :smile:

Eilon:
:trollface:

anpete:
:expressionless:

PR #254, opened 2015-09-26
Reduce allocs when no loggers

benaadams:
* Only create logger array when loggers
* Normal for loop condition in constructor
* Log nop when no loggers
* BeginScope to return null when no loggers
* Logging is disabled when no loggers
* Add provider to only create one resized array; rather than several

### DIFF
--- a/src/Microsoft.Extensions.Logging/Logger.cs
+++ b/src/Microsoft.Extensions.Logging/Logger.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 
 namespace Microsoft.Extensions.Logging
 {
@@ -11,7 +10,7 @@ namespace Microsoft.Extensions.Logging
     {
         private readonly LoggerFactory _loggerFactory;
         private readonly string _name;
-        private ILogger[] _loggers = new ILogger[0];
+        private ILogger[] _loggers;
 
         public Logger(LoggerFactory loggerFactory, string name)
         {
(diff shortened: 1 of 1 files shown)


benaadams:
Hmm, takes the allocations from 12,000 objects -> 4,020 objects but memory is same; needs work

benaadams:
K, is good...

Before Allocations 12000 per 4k
pre-allocs
Before Bytes 448kB per 4k
pre-bytes

After Allocations 8000 per 4k
post-allocs
After Bytes 256kB per 4k
post-bytes

benaadams:
Rebased

benaadams:
This doesn't resolve issues with the actual log calls boxing and struct wrapping; whether or not its going to log anything; but I believe this is being addressed elsewhere

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -96,10 +108,16 @@ public bool IsEnabled(LogLevel logLevel)
 
         public IDisposable BeginScopeImpl(object state)
         {
+            if (_loggers == null)
+            {
+                return null;
```
lodejard: When `_loggers == null` return a singleton IDisposable object which does nothing in the Dispose method. No allocations, but BeginScopeImpl can always returned a non-null value.

Also when `_loggers.Count == 1` you can `return _loggers[0].BeginScopeImpl(state)` - there's no reason to create an aggregate disposable around a single instance

benaadams: Done

lodejard:
:shipit:

PR #260, opened 2015-10-07
Make logging tests run in dnxcore50

pranavkm:


davidfowl:
Just get rid of mow until we have it

pranavkm:
It's harder to test things without it. We're using this pattern of selectively using Moq in Mvc and it's working fine

PR #263, opened 2015-10-13
Change for Logging Harmonization.

vancem:


nblumhardt:
Hi Vance! This looks like a complete reset on _Logging_ - is there a discussion anywhere on the motivations here?

The new API appears to drop `"{Named}"` format parameters, which would be a shame given the growing support for this in the ecosystem (not just as a front-end to Serilog, but in [other](https://github.com/akkadotnet/akka.net/issues/1045) [projects](https://github.com/SuaveIO/suave/pull/288) that want structured logging support.)

The new API also doesn't seem to mesh with anything else currently in use, and requires an allocation (as well as the assembly footprint of an anonymous type?) in order to pass structured data:

```csharp
_logger.Log("InfoMessage", LogLevel.Informational, new { startTime = startTime, hex = 0x42 });
```

The output is also not going to be as pleasant to read at the console as the existing text-friendly format.

Given the change in direction, huge downstream churn, unfamiliar API and additional footprint, it seems like there must be a strong force driving this - can we please have some explanation of this PR?

Cheers,
Nick

davidfowl:
These are the proposed changes right? I don't want to comment on little stuff it's just a design PR (there's tons of it).

- Why are the extension methods gone? 
- Using IObservable to receive log messages seems like a misuse of an existing API
- Why were the other providers (like Nlog) deleted? We need to see how this new system works with third  party logging systems.
- +1 to what @nblumhardt said. Maybe we can salvage it somehow. @lodejard is looking at optimizing logging in general for format strings.
- nit: Microsoft.Extensions.Logging.Observer that package name is weird
- Why did you change MinimumLevel to MaximumLevel on the factory
- Why does subscribe take a filter? Why is that first class? Isn't it sink specific?
- I see lots of deleted code with no replacement. That could mean you moved it or that ypu just deleted it.

General Engineering:
The code doesn't follow our engineering guidelines but I'm assuming that will be fixed once the other things are fixed.

adamchester:
+1 to what @nblumhardt said too. 

Re: optimised logging with format strings, I have done [some work on this recently](http://adamchester.github.io/MessageTemplates/performance.html), particularly with the message template format described by Serilog. Not sure if any of it helps you here, but I'd be interested to see what you had in mind :)

vancem:
Let me address @nblumhardt's feedback on the motivation.    The main motivation is to not have fragmentation in logging mechanism/format that at least the Framework/Microsoft DLLs use to log telemetry/diagnostics.     Ideally we would have one logging system that lots of things could plug into on both the production and consumption side.   

Sadly one logging mechanism not possible, the .NET Framework currently has two (but they are strongly related), that are used in the framework. 
    System.Diagnostics.Tracing.EventSource
    System.Diagnostics.DiagnosticSource 
The main reason for having two is that DiagosticSource can log ANY object to an in process listener, where EventSource can log a broad class of SERIALIZABLE objects out of process (to files/DBs ...) 
Both of these adopt a 'semantic logging' model where at its heart you don't log formatted strings, you log named events with arguments.   Thus they APIs like Log(string name, object arguments).   

The argument for this approach is that generally speaking, formatted strings are for humans to read, logging volume quickly makes it more desirable to records so that the logs can be manipulated/filtered mechanically before look at them.    Note that the semantic logging model does not exclude formatted strings (you can pass a format strings if you wish), it just does not make them primary, because logging is really not about making pretty strings, it is about getting diagnostic data (the arguments the key, not the formatting strings).    In the framework at least, the format strings tend to be boilerplate of the general form "EVENT happened with NAME1 = VALUE1 NAME2 = VALUE2", which can be generated with the semantic information.  

As you point out we are booting an eco system, so ideally we are promoting best practices.   In a world of big data, formatting strings are really more of a liability than an asset, and so ideally we don't force their use in our logging APIs.  
 
The existing Serilog style syntax does  support a 'semantic' approach, but it really does keep the formatted string as 'primary'.    In particular in typical use, you don't give a message an ID, (and even in advanced use you give it a numeric ID, which makes difficult to keep unique, further discouraging users from using IDs).    Thus you pass a format, which is a not valuable in mechanical processing, but don't provide an ID which is useful).   Using a name (ID) also side-steps the issue of foreign language support.  

Still the Serilog style could be used, but it is used nowhere else in the framework, and worse, C# 6.0 has adopted a syntax (called string interpolation) is identical, but means something different (in C# it captures local variables and emits standard formatting strings, in Serilog, it give other parameters names).   

it is true that the use of anonymous types does force an object allocation if multiple arguments are used (a single anonymous argument does not have the problem).    This is rather unfortunate, and forces guidance to guard logging calls with an IsEnabled() check if they are on hot code paths.    However it is also true that the current 'params' structure in any non-trivial logging (where there is more than a single string object) as a allocation (for the 'params' array) as well as potential boxing for any numeric arguments (reasonably likely).   The use of anonymous types in in the limit faster than using 'params' because while it is true that need to use reflection THE FIRST TIME a particular anonymous type is used, it can be very fast for the next instance and avoid boxing-unboxing overhead of integral values.  

Thus the motivations are to 

  * Encourage logging that is first and foremost designed for machine processing while still supporting human readable logging well (The strings we can produce from the semantic information are really very readable).   

  * Interoperate seamlessly with the existing .NET logging systems (we don't have ANOTHER logging system with unclear guidance on when to use it).  Ideally it would simply use, or build on top of the existing system (thus not really introduce interoperation overhead/mismatches).    Ideally it looks like ONE logging system, that has many extensions and clients.    In particular I should be able to write ONE piece of code that can subscribe to ANYTHING in the system.  I should not have to write connecting' code depending on which API was done to log them message.   

These are the primary motivations.

lodejard:
@vancem yeah, I believe the nature of this reset is a non-starter, it regresses too many scenarios and doesn't take into account some of the perf implications that we've needed to drive down recently. Lets continue this conversation offline to be sure the harmonization is a PR we will be able to accept.

vancem:
cc @rahku

merbla:
It would be great if any discussions/reasoning relating to a change of this magnitude could be done in the open.

vancem:
@davidfowl 's question is a design PR.  YES it is.  That is why it is incomplete.  We wanted to get top level feedback.   

•Why are the extension methods gone?   

Specifically which ones?   the PR is showing the design.   Smaller is better assuming things are equally convenient, but if we have feedback that they add value then we put them back 

•Using IObservable to receive log messages seems like a misuse of an existing API

Please elaborate.   My understanding is IObservable is a 'callback' version of IEnumerable, and thus is appropriate any time you have a sequence of things.   That applies here.    More pragmatically, there is a nice library (the reactive framework) which can do cool filtering/selection/grouping on IObservables which are valuable here.   

•Why were the other providers (like Nlog) deleted? We need to see how this new system works with third party logging systems.

The PR is incomplete.   NLog support will be there.  

•nit: Microsoft.Extensions.Logging.Observer that package name is weird

Agreed.  Updated.

•Why did you change MinimumLevel to MaximumLevel on the factory

Sadly, the existing .NET logging system has the LogLevel increase in Verbosity not Severity (thus the largest number is the most verbose (least Severe)   Thus Min -> Max.   It is unfortunate that both conventions existed in the wild before this code.   We picked the existing .NET Framework convention.  

•Why does subscribe take a filter? Why is that first class? Isn't it sink specific?

Prefiltering (at IsEnabled time) on event name or level is something that can be done independently of any specific sync.    (Note that in the design, the sink is an Observer, not a log creation factory).

rahku:
@davidfowl 
NLog support was not complete yday....I have added NLog support and pushed the changes. You can see the implementation now.

rogeralsing:
+1 on @nblumhardt
Please don't kill off existing **working** solutions with some Not Invented Here approach just because you have the weight behind you as an organization to do so..

nblumhardt:
Hi Vance, thank you for the detailed reply. I should mention that your work on `EventSource` was one of the things that led me to appreciate the importance and opportunity of good instrumentation, and predates my interest in this area considerably, so please accept my comments here with the disclaimer that I'm addressing the proposed change and not the merits of different logging technologies in general.

In reply:

> The main motivation is to not have fragmentation in logging mechanism/format that at least the Framework/Microsoft DLLs use to log telemetry/diagnostics.

The needs of .NET Framework components and the needs of application programmers diverge around this. `System.Diagnostics.Tracing` has been available since the first version of .NET, and `EventSource` in the framework for several years now. Yet, libraries like log4net, NLog, and now to a small-but-growing extent Serilog have been the preference of many individual developers, teams and companies.

Diversity is already the norm; where it's not possible to support all approaches, remaining agnostic at the perimeter is best for the largest number of .NET developers. Though it's regrettable that ASP.NET is in the position of _needing_ to pick a logging implementation, _Microsoft.Framework.Logging_ succeeds at working in harmony with the solutions used by developers today. One can use _Logging_ and log4net, or _Logging_ and Serilog in the same application and have a first-class experience whether using printf-style or `"{token}-style"` structured data.

> Both of these adopt a 'semantic logging' model where at its heart you don't log formatted strings, you log named events with arguments. Thus they APIs like Log(string name, object arguments).

> In a world of big data, formatting strings are really more of a liability than an asset, and so ideally we don't force their use in our logging APIs.

This is also the case with _Logging_ as it stands today. When used with a structured-data back-end, a statement like:

```csharp
_log.LogInformation("Loading took {Elapsed} ms", elapsed);
```

is not necessarily formatted into text, unless it needs to go to the console or a flat file. Many users of this style of logging create an event id by hashing the format string, so `"Loading took {Elapsed} ms"` becomes the id of the event (perhaps boiled down to SHA1), and `Elapsed = 123` the data payload.

The advantage here is that, while human-readable formatted strings are available as a projection, the highly-compressible, machine-readable format is the underlying representation.

Given the popularity of printf-style logging today, misuse of the token-based approach by a developer encountering it for the first time might appear as:

```csharp
_log.LogInformation("Loading took {0} ms", elapsed); 
```

In that case, diagnostic tools _still get the format string_ as a unique event type, so automated processing is aided (though [without heuristics](https://github.com/akkadotnet/akka.net/pull/1295) the arguments are positional rather than named).

In the proposed API style `Log(string name, object arguments)`, misuse might look like:

```csharp
_log.LogInformation(string.Format("Loading took {0} ms", elapsed), null); 
```

At this point, the API encourages the user into exactly the pattern we're trying to avoid, and the structure of the data is irrecoverable because formatting is applied before the event enters the logging pipeline.

> The existing Serilog style syntax does support a 'semantic' approach, but it really does keep the formatted string as 'primary'. In particular in typical use, you don't give a message an ID, (and even in advanced use you give it a numeric ID, which makes difficult to keep unique, further discouraging users from using IDs). Thus you pass a format, which is a not valuable in mechanical processing, but don't provide an ID which is useful). Using a name (ID) also side-steps the issue of foreign language support.

As mentioned above, SHA1 (or even Murmur32) hashing of the format string yields a convenient fixed-size event id for mechanical processing. This is already in use by developers using Serilog with various back-ends today. The apparent primacy of the formatted string is just a demonstration of how both human and mechanical needs are served by the same API: you notice the formatted string because you're a human reading the event as text; once the event is fed into a log processing system, it's structured data all the way.

Because of the difficulty of creating unambiguous global namespaces, I would not be surprised if identifiers generated this way conflict less than those assigned directly as names by programmers. (Compare the hash value generated from `"Loading took {Elapsed} ms"`, and the hastily-assigned name `"LoadingFinished"`).

> Still the Serilog style could be used, but it is used nowhere else in the framework, and worse, C# 6.0 has adopted a syntax (called string interpolation) is identical, but means something different (in C# it captures local variables and emits standard formatting strings, in Serilog, it give other parameters names).

Given this constraint, LINQ might have been omitted from C# 3 because of the similarity with SQL literals. In both cases, syntax highlighting in Visual Studio makes this a non-issue in pratice. It's also worth considering that there are millions of lines of code in existence using `"numbered {0}"` holes and these need to (and will) stand clearly alongside interpolation expressions as well.

> it is true that the use of anonymous types does force an object allocation if multiple arguments are used (a single anonymous argument does not have the problem). This is rather unfortunate, and forces guidance to guard logging calls with an IsEnabled() check if they are on hot code paths.

In addition to the allocated object, each logging line that uses anonymous types will also generate an entire class that will have to be maintained in the assembly metadata and cannot be optimised away. Used for logging, hundreds of times within even a single assembly, the footprint may be measurable. The cost of the proposed API appears to be both an object allocation, _and_ an anonymous type.

> However it is also true that the current 'params' structure in any non-trivial logging (where there is more than a single string object) as a allocation (for the 'params' array) as well as potential boxing for any numeric arguments (reasonably likely).

Generic overloads [like this one](https://github.com/serilog/serilog/blob/dev/src/Serilog/Log.cs#L149) eliminate all allocations using the token-based API, and a PR has been put forward to _Logging_ that would trivially support this. Neither an `object[]` allocation for the parameters, nor any boxing for numeric values, is required for three or fewer arguments.

> Thus the motivations are to

> Encourage logging that is first and foremost designed for machine processing while still supporting human readable logging well (The strings we can produce from the semantic information are really very readable).

As we've seen, the existing API has the same or better machine-readability characteristics, while producing optimally human-readable output when required.

> Interoperate seamlessly with the existing .NET logging systems (we don't have ANOTHER logging system with unclear guidance on when to use it). Ideally it would simply use, or build on top of the existing system (thus not really introduce interoperation overhead/mismatches). Ideally it looks like ONE logging system, that has many extensions and clients. In particular I should be able to write ONE piece of code that can subscribe to ANYTHING in the system. I should not have to write connecting' code depending on which API was done to log them message.

In the proposal:

 * interoperation with existing formatting-based loggers (log4net, NLog) will be awkward because plain text is second-class
 * interoperation with the emerging token-based structured loggers (Serilog and the independent message template implementation) will be suboptimal because the format cannot be supported; and
 * languages like F# without anonymous types will be second-class.

In all three interop scenarios the existing _Logging_ implementation is adequate.

The challenges of maintaining the .NET core platform are immense, and I understand that a level of consistency benefits all users over the long-term. The current proposal doesn't achieve consistency apart from within a narrow band of framework libraries, and does this at the expense of the open, diverse ecosystem building around DNX and ASP.NET 5.

andrexx:
+1 on @nblumhardt

vancem:
@nblumhardt.   That you for your thoughtful feedback.   I was out on Friday, and today I have been in discussion with various people about this issue.   I had hoped to send something today (and may yet), but this is just to let you know that we are not ignoring this issue.    I realize that it there may be a desire for visibility into what is happening on our side.   I do think however that a certain amount of 'private validation' is more efficient for everyone involved.    Thanks for your patience.

adamchester:
@vancem do you think that the reasoning will be made open once it's "privately validated"?

PureKrome:
> I do think however that a certain amount of 'private validation' is more efficient for everyone involved.

I think what some people (like @nblumhardt) are asking is that it's ok to do heaps of private chats but when some pretty large decisions have been more-or-less decided on, can they be posted up here for feedback before coding effort is made. That's all. We're all a bit shy/cautious/(sometimes even burnt) by a massive change to some popular framework without public consultation which (many in the public feel) are sometimes some steps backwards, not forwards ... and that it's not inclusive with those people who will be consuming the end-product(s).

In this scenario, I feel that a number of peeps saw this PR and went: "_ahh-what-the? who/why/how? did I miss the discussion about this? why, why and why?_"

In the end we all would :heart: a good outcome for all parties :+1: .. and we have lots of :sparkles: faith in the New-MS and vnext :smile:

vancem:
@adamchester,  - Yes the rationale will be made public.    To be clear, we are not building another implementation, but another proposal.   You will certainly get your chance to weigh in.  

@PureKrome  - For what it is worth, this PR was a piece of a companion pull request https://github.com/dotnet/corefx/pull/3822 and I unfortunately, in my ignorance and haste did not repeat the information that was posted there in this PR.   Please see that PR as it  provides at least some background.   It also mentions that the PR was TENTATIVE with the intention of gathering feedback (it was a very early implementation.   I apologize for any angst that this has caused.   

We are taking all the feedback into account.    Thank you for participating.

PureKrome:
and @vancem thanks for chatting here in the open to everyone :clap: :cake: Keep up the great work - we all do appreciate everyone's efforts :+1:

cwe1ss:
Since most comments so far have been written by framework developers (each one of you is doing a truly amazing job by the way!) I’d like to give you some feedback as a regular „framework user“.

I would like to do this by showing you how we currently (have to) handle logging within our .NET codebase. 

* In our business and application layers (regular DLLs), we use the supporting libraries from „ASP.NET Boilerplate“. This project uses the ILogger abstraction from the Castle-Project.
* To make sure we only have one logging concept in our business layer, we are also using the ILogger from the Castle-project to create our own log entries.
* Our business layer is hosted in WebAPI projects on Application servers. We are using OAuth to protect the communication our services. (Azure Active Directory) To do this, we are using the OWIN projects from Katana. These components are using Microsoft.Owin.Logging to emit events.
* We plan on hosting our APIs in the upcoming Azure Service Fabric which uses System.Diagnostics.Tracing.EventSource to emit events. 
* To be honest, I haven’t heard about EventSource until I started reading about Service Fabric. It seems like the .NET framework (and lots of other components) emit lots of useful events through this library.
* To provide SSO for our end-users on our public website, we are using „IdentityServer3“. This project uses the logger abstraction LibLog. (In its previous version it used System.Diagnostics.Trace)
* Some other libraries are using System.Diagnostics.Trace to emit events
* We have some PoC projects for ASP.NET 5 which use the new Microsoft.Extensions.Logging framework.

To actually see the generated events, all of these logging frameworks have to be configured, which requires a certain amount of knowledge about each logging framework and uses different APIs (web.config entries, static method calls, dependency injection, transparent configuration [LibLog], …). It is almost impossible to configure them to directly write to a central location (Console, file) and in an unified format. And since this has to be done in every application, it is very hard to keep these settings in sync. 

To fix this, we are doing our best to forward all events into yet another logging framework, Serilog, which then is responsible to store the logs to actual sinks (console, files, ApplicationInsights, …). To keep the application configuration simple, we moved the configuration about forwarding of the logging abstractions to Serilog to shared libraries – this way, our application developers only have to focus on how to configure the Serilog sinks.

All of this work is required to store events from *one technology stack (.NET)* into one central location. To complete our enterprise logging strategy, we have to aggregate this log with logs from other technology stacks (IIS, Windows, Load Balancer, …). But of course, this is not topic of this issue.

### my opinion:

My main point is, that it shouldn’t be that hard to consolidate events from one single technology stack. I had to read **a lot** to create our current solution.

Tracing/Logging is a feature that should be supported *by the core of the technology stack* and it should be easy for framework and application developers to add their own events to this central tracing mechanism. This way, people only have to concentrate on where they want to store their events.

*In this concrete szenario, tracing should be handled by the .NET core. ASP.NET is just a framework on top of .NET and therefore shouldn’t implement its own logging. It should only have extensions for it.* 

I really like the easy syntax of Serilog and Microsoft.Extensions.Logging, but I also understand @vancem’s arguments for EventSource. And since this is the recommended solution inside the core of .NET this should be the solution for ASP.NET as well. 

I think, EventSource is ok to use and just needs more marketing :-) and some collective effort to improve the tooling. E.g., maybe there should be a default EventSource with an application specific name that can be used by simple apps to mimic the current Debug/Info/Warning/Error log messages of other logging frameworks. And there should be an easy way to output all events to the console during development time (to support scenarios where the "diagnostics events" window of visual studio is not available). This would give us good development-time experience and also the amazing production performance of EventSource. The P&P team already did a good job with their Semantic Logging Application Block, but it seems like there is no active development.

I'm wondering if this issue was also discussed in the initial phase of ASP.NET vNext? I’m afraid it might be way too late now for bigger changes which means, Microsoft.Extensions.Logging just adds to the long list of logging frameworks every application developer has to configure.

I know this is a very hard and opinionated topic and I really appreciate the work you guys are doing. I'm really looking forward to the release!

mteper:
:+1:

Current glut of logging frameworks doesn't compose well in real life apps.

davidfowl:
There are many ways to accomplish sending data to the same pipe without forcing the exact same abstraction to be used. They all just need to agree on a good default sink.

Eilon:
Closing this as we have decided to stick closer to the existing approach. Thanks to everyone for all the feedback!

PR #264, opened 2015-10-14
[Fixes #247] ConsoleLogger doesn't use LoggerFactory.MinimumLevel by …

### LINKED ISSUE #247: ConsoleLogger doesn't use LoggerFactory.MinimumLevel by default
See this line: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Framework.Logging.Console/ConsoleLoggerFactoryExtensions.cs#L16

This means the following doesn't result in verbose messages being emitted:

``` csharp
loggerFactory.MinimumLevel = LogLevel.Verbose;
loggerFactory.AddConsole();
```

kichalla:
@Eilon @AspNetSmurfLab Currently, the default level on logger factory is `Verbose`, which we want to change to `Information`, right? https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/LoggerFactory.cs#L34

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/Properties/AssemblyInfo.cs
+++ b/src/Microsoft.Extensions.Logging.Console/Properties/AssemblyInfo.cs
@@ -3,6 +3,8 @@
 
 using System.Reflection;
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
 [assembly: NeutralResourcesLanguage("en-us")]
+[assembly: InternalsVisibleTo("Microsoft.Extensions.Logging.Test")]
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs
@@ -13,8 +13,7 @@ namespace Microsoft.Extensions.Logging
         /// </summary>
         public static ILoggerFactory AddConsole(this ILoggerFactory factory)
         {
-            factory.AddProvider(new ConsoleLoggerProvider((category, logLevel) => logLevel >= LogLevel.Information));
-            return factory;
+            return factory.AddConsole(factory.MinimumLevel);
         }
 
         /// <summary>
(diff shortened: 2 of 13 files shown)


Eilon:
Wait, why do we want to change it? What was wrong?

@DamianEdwards

src/Microsoft.Extensions.Logging.Console/Properties/AssemblyInfo.cs
```diff
@@ -3,6 +3,8 @@
 
 using System.Reflection;
 using System.Resources;
+using System.Runtime.CompilerServices;
 
 [assembly: AssemblyMetadata("Serviceable", "True")]
-[assembly: NeutralResourcesLanguage("en-us")]
\ No newline at end of file
+[assembly: NeutralResourcesLanguage("en-us")]
+[assembly: InternalsVisibleTo("Microsoft.Extensions.Logging.Test")]
```
Eilon: IVT is only from a product assembly to its own unit test assembly. Never use across misc assemblies.

DamianEdwards:
Not sure about changing the default level. Just wanted to ensure the extension methods to add providers honored the default level.

Eilon:
I'm thoroughly confused. What is "the default level"? The minimum level has nothing to do with any defaults - it just says that any logs below the minimum will not even be seen by loggers, no matter how hard they try.

What exactly was the bug or undesirable behavior?

kichalla:
@Eilon Before my current change the default for the Console, Debug and EventLog was Information and now since we would default to logger factory's level, it would be Verbose...do we want Verbose to be the default?

src/Microsoft.Extensions.Logging.Console/Properties/AssemblyInfo.cs
kichalla: Ok, but there is only test project for all these Console, Debug and EventLog projects...are you suggesting to create different test projects respectively?

kichalla:
@Eilon fyi https://github.com/aspnet/Logging/issues/247

Eilon:
Unfortunately, that bug is not a bug, it's by design.

DamianEdwards:
The design is bad then. The code in the bug should result in logging at the default level of the factory IMO

Eilon:
There's *no such thing* as a default level for the factory. The property is called `MinimumLevel`. Adding a `DefaultLevel` would be a separate feature request.

PR #265, opened 2015-10-17
Logging improvements

kichalla:
@loudej @lodejard @rynowak @Eilon


NuGet.config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
kichalla: probably BOM...will revert it

rynowak: This is my favorite part so far.

samples/SampleApp/LoggerExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        static LoggerExtensions()
+        {
+            LoggerMessage.DefineScope(out _purchaceOrderScope, "PO:{PurchaceOrder}");
+            LoggerMessage.Define(out _programStarting, LogLevel.Information, 1, "Starting", "at '{StartTime}' and 0x{Hello:X} is hex of 42");
+            LoggerMessage.Define(out _programStopping, LogLevel.Information, 2, "Stopping", "at '{StopTime}'");
+        }
+
+        public static IDisposable PurchaceOrderScope(this ILogger logger, string purchaceOrder)
+        {
+            return _purchaceOrderScope(logger, purchaceOrder);
```
rynowak: Fix spelling here :laughing:

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+using Microsoft.Extensions.Logging.Internal;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Creates delegates which can be later cached to log messages in a performant way.
+    /// </summary>
+    public static class LoggerMessage
```
rynowak: Name seems a little iffy - @Eilon ?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+        /// <summary>
+        /// Creates a delegate which can be used for creating a log scope.
+        /// </summary>
+        /// <param name="scope">A delegate which creates a log scope</param>
+        /// <param name="formatString">The named format string</param>
+        public static void DefineScope(out Func<ILogger, IDisposable> scope, string formatString)
+        {
+            var formatter = new LogValuesFormatter(formatString);
+
+            scope = logger => logger.BeginScopeImpl(new LogValues(formatter));
```
rynowak: Should we close over `new LogValues` as well? In the case of just a format string we can remove that allocation

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Creates delegates which can be later cached to log messages in a performant way.
+    /// </summary>
+    public static class LoggerMessage
+    {
+        /// <summary>
+        /// Creates a delegate which can be used for creating a log scope.
```
rynowak: `can be used for creating a log scope` -> `can be invoked to create a log scope`

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+        {
+            var formatter = new LogValuesFormatter(formatString);
+
+            scope = (logger, arg1) => logger.BeginScopeImpl(new LogValues<T1>(formatter, arg1));
+        }
+
+        /// <summary>
+        /// Creates a delegate which can be used for creating a log scope.
+        /// </summary>
+        /// <typeparam name="T1">The parameter passed to the named format string</typeparam>
```
rynowak: This is the type of the parameter, not the parameter

rynowak: Use periods, these are generally complete sentences.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+        /// <param name="formatString">The named format string</param>
+        public static void DefineScope<T1, T2, T3>(out Func<ILogger, T1, T2, T3, IDisposable> scope, string formatString)
+        {
+            var formatter = new LogValuesFormatter(formatString);
+
+            scope = (logger, arg1, arg2, arg3) => logger.BeginScopeImpl(new LogValues<T1, T2, T3>(formatter, arg1, arg2, arg3));
+        }
+
+        /// <summary>
+        /// Creates a delegate which can be used for logging a message.
```
rynowak: `can be used for logging a message` -> `can be invoked to log a message`

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+
+            scope = (logger, arg1) => logger.BeginScopeImpl(new LogValues<T1>(formatter, arg1));
+        }
+
+        /// <summary>
+        /// Creates a delegate which can be used for creating a log scope.
+        /// </summary>
+        /// <typeparam name="T1">The parameter passed to the named format string</typeparam>
+        /// <typeparam name="T2">The parameter passed to the named format string</typeparam>
+        /// <param name="scope">A delegate which creates a log scope</param>
```
rynowak: `delegate`?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+        /// <summary>
+        /// Creates a delegate which can be used for logging a message.
+        /// </summary>
+        /// <typeparam name="T1">The parameter passed to the named format string</typeparam>
+        /// <param name="message">A delegate which creates a log message</param>
+        /// <param name="logLevel">The <see cref="LogLevel"/></param>
+        /// <param name="eventId">The event id</param>
+        /// <param name="eventName">The event name</param>
+        /// <param name="formatString">The named format string</param>
+        public static void Define<T1>(out Action<ILogger, T1, Exception> message, LogLevel logLevel, int eventId, string eventName, string formatString)
```
rynowak: Where did `eventName` from from? @loudej @lodejard

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+                }
+            };
+        }
+
+        private class LogValues : ILogValues
+        {
+            public static Func<object, Exception, string> Callback = (state, exception) => ((LogValues)state)._formatter.Format(((LogValues)state).ToArray());
+
+            private static IEnumerable<KeyValuePair<string, object>> _getValues = new KeyValuePair<string, object>[0];
+            private static object[] _toArray = new object[0];
```
rynowak: Find a better name for this

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
...
+
+            private readonly LogValuesFormatter _formatter;
+
+            public LogValues(LogValuesFormatter formatter)
+            {
+                _formatter = formatter;
+            }
+
+            public IEnumerable<KeyValuePair<string, object>> GetValues() => _getValues;
```
rynowak: Is the change to make this indexable coming as part of this PR?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
rynowak: Should we still include `OriginalFormat`?

test/Microsoft.Extensions.Logging.Test/TestLogger.cs
```diff
@@ -31,7 +31,7 @@ public IDisposable BeginScopeImpl(object state)
                 Scope = state,
             });
 
-            return null;
+            return new NoopDisposable();
```
rynowak: Why is this needed?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,343 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.Internal;
```
Eilon: Sort 'em.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
Eilon: Yeah not a huge fan of the name either. Usually these types of static classes are more verby and less nouny because they merely contain helper methods, and do not themselves represent anything.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
Eilon: I think it's `categoryName` elsewhere: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging/LoggerFactory.cs#L20

test/Microsoft.Framework.Logging.Test/LoggerMessageTest.cs
```diff
@@ -0,0 +1,204 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Framework.Logging.Test;
+using Xunit;
+using Xunit.Sdk;
+
+namespace Microsoft.Framework.Logging
+{
+    [Trait("test", "loggermessage")]
```
Eilon: Do we actually use traits anywhere?

test/Microsoft.Framework.Logging.Test/TestLoggerExtensions.cs
```diff
@@ -0,0 +1,110 @@
+﻿using System;
```
Eilon: Insert file header.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
kichalla: This would be part of the message itself and if am not wrong should ideally be different than the logger name...
In the following logger, the name of the logger is already part of the message but `eventname` would be additional detail to the message.
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L79

test/Microsoft.Framework.Logging.Test/LoggerMessageTest.cs
kichalla: Not in many places(ex: musicstore uses them), but I like them as it allows me to run only these specific tests...will remove it.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,344 @@
...
+            };
+
+            public object[] ToArray() => new object[] { _value0 };
+
+            public override string ToString() => _formatter.Format(ToArray());
+        }
+
+        private class LogValues<T0, T1> : ILogValues
+        {
+            public static Func<object, Exception, string> Callback = (state, exception) => ((LogValues<T0, T1>)state)._formatter.Format(((LogValues<T0, T1>)state).ToArray());
```
kichalla: @lodejard @loudej Callbacks like these could just be `state.ToString()` as we are overrding ToString in them?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -0,0 +1,344 @@
...
+            private readonly T1 _value1;
+
+            public LogValues(LogValuesFormatter formatter, T0 value0, T1 value1)
+            {
+                _formatter = formatter;
+                _value0 = value0;
+                _value1 = value1;
+            }
+
+            public IEnumerable<KeyValuePair<string, object>> GetValues() => new[]
```
kichalla: @loudej @lodejard wondering why do we need to do it this way...like we could use the formatter which already does this for us `formatter.GetValues(ToArray())`
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs#L146

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
rynowak: I still have more questions than answers after reading your reply. I know what it **does** from reading the code. Why do we need it an why is it part of this PR?

test/Microsoft.Extensions.Logging.Test/TestLogger.cs
kichalla: This was to make sure that the test logger's scope was created when the func was executed

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
kichalla: :+1: looks like `OriginalFormat` should be included based on how `LogValuesFormatter` implements this. Updated the test accordingly.

kichalla:
Updated @loudej @lodejard @rynowak @Eilon

rynowak:
:shipit:

kichalla:
7c01b9c3639801d89bd88bb607600c57dfd4cabf

PR #268, opened 2015-10-20
Added doc comments for main APIs:

DamianEdwards:
Coherence-Signed#75

Adds or fixes doc comments for the main APIs, especially those used from templates.

@Eilon

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LogLevel.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LogLevel.cs
@@ -3,6 +3,9 @@
 
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Defines logging severity levels.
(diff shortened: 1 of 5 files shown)


src/Microsoft.Extensions.Logging.Abstractions/LogLevel.cs
```diff
@@ -3,6 +3,9 @@
 
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Defines logging severity levels.
+    /// </summary>
     public enum LogLevel
     {
         Debug = 1,
```
Eilon: We can perhaps copy some of the guidance from https://github.com/aspnet/Logging/wiki/Guidelines to the enum values.

Eilon:
:shipit:

DamianEdwards:
Rebased and merged

PR #271, opened 2015-10-23
Added scoping information to Console logger

kichalla:
@rynowak @loudej @lodejard

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -21,7 +21,7 @@ namespace SampleApp
             factory.AddNLog(new global::NLog.LogFactory());
             factory.AddEventLog();
 #endif
-            factory.AddConsole();
+            factory.AddConsole(includeScopes: true);
         }
 
         public void Main(string[] args)
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
@@ -1,0 +1,84 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+#if NET451
+using System.Runtime.Remoting;
+using System.Runtime.Remoting.Messaging;
+#else
+using System.Threading;
+#endif
+
+namespace Microsoft.Extensions.Logging.Console
+{
+    public class ConsoleLogScope
+    {
+        private readonly string _name;
+        private readonly object _state;
+
+        internal ConsoleLogScope(string name, object state)
+        {
+            _name = name;
+            _state = state;
+        }
+
+        public ConsoleLogScope Parent { get; private set; }
+
+#if NET451
+        private static string FieldKey = typeof(ConsoleLogScope).FullName + ".Value";
+        public static ConsoleLogScope Current
+        {
+            get
+            {
+                var handle = CallContext.LogicalGetData(FieldKey) as ObjectHandle;
+                if (handle == null)
+                {
+                    return default(ConsoleLogScope);
+                }
+
+                return (ConsoleLogScope)handle.Unwrap();
+            }
+            set
+            {
+                CallContext.LogicalSetData(FieldKey, new ObjectHandle(value));
+            }
+        }
+#else
+        private static AsyncLocal<ConsoleLogScope> _value = new AsyncLocal<ConsoleLogScope>();
+        public static ConsoleLogScope Current
+        {
+            set
+            {
+                _value.Value = value;
+            }
+            get
+            {
+                return _value.Value;
+            }
+        }
+#endif
+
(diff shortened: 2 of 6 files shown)


samples/SampleApp/Program.cs
```diff
@@ -21,7 +21,7 @@ public Program()
             factory.AddNLog(new global::NLog.LogFactory());
             factory.AddEventLog();
 #endif
-            factory.AddConsole();
+            factory.AddConsole(showScope: true);
```
rynowak: `includeScopes`

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+
+#if NET451
+using System.Runtime.Remoting;
+using System.Runtime.Remoting.Messaging;
+#else
+using System.Threading;
+#endif
+
+namespace Microsoft.Extensions.Logging.Console
```
rynowak: should be probably be `.Internal`

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
```diff
@@ -0,0 +1,89 @@
...
+        private readonly string _name;
+        private readonly object _state;
+
+        public ConsoleLogScope(string name, object state)
+        {
+            _name = name;
+            _state = state;
+        }
+
+        public ConsoleLogScope Parent { get; private set; }
```
rynowak: Should we pass parent in the constructor as well?

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
```diff
@@ -0,0 +1,89 @@
...
+                _value.Value = value;
+            }
+            get
+            {
+                return _value.Value;
+            }
+        }
+#endif
+
+        public static IDisposable Push(ConsoleLogScope scope)
```
rynowak: We we rejigger this to just take a name, and state?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -247,6 +269,29 @@ private ConsoleColors GetLogLevelConsoleColors(LogLevel logLevel)
...
+        private string GetScopeInformation()
+        {
+            var current = ConsoleLogScope.Current;
+            var output = new StringBuilder();
+            string scopeLog = string.Empty;
+            while (current != null)
+            {
+                if (output.Length == 0)
+                {
+                    scopeLog = $"=> {current}";
```
rynowak: you don't really need to do a string.Format if you have a builder already

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs
```diff
@@ -13,16 +13,18 @@ public static class ConsoleLoggerExtensions
...
+            return factory.AddConsole(showScope: false);
         }
 
         /// <summary>
-        /// Adds a console logger that is enabled as defined by the filter function.
+        /// Adds a console logger that is enabled for <see cref="LogLevel"/>.Information or higher.
         /// </summary>
-        public static ILoggerFactory AddConsole(this ILoggerFactory factory, Func<string, LogLevel, bool> filter)
+        /// <param name="factory"></param>
+        /// <param name="showScope">Flag for indicating if log scope information should be displayed
```
rynowak: A value which indicates whether....

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -464,6 +464,162 @@ public void WriteCore_LogsCorrectMessages()
         }
 
         [Fact]
+        public void NoLogScope_DoesNotWriteAnyContentToOutput()
```
rynowak: `NoLogScope_DoesNotWriteAnyScopeContentToOutput`

rynowak:
:shipit: after fixing minor comments

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
kichalla: I thought about this but left it like this because we allow users to override `WriteMessage` and probably they would be interested in getting the log scope to write the way they want to

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
kichalla: Probably no as the details of this relationship is encapsulated within this class and the ConsoleLogger class need not know about it

src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
kichalla: Sure, will update it

PR #273, opened 2015-10-24
Adding netcore50 to Logging packages

pranavkm:


pranavkm:
cc @natemcmaster 

Do we care about adding netcore50 to all the packages or just the ones that are required by EF?

natemcmaster:
Definitely those required by EF, but it doesn't hurt to add it for all Microsoft.Extension.* because users may wish to use these libraries in UWP projects. 

:shipit:

PR #276, opened 2015-10-26
Add more configuration options for logging, add reload

rynowak:


samples/SampleApp/Program.cs
```diff
@@ -21,7 +25,46 @@ public Program()
...
+                args.ChangeToken = files.Watch("logging.json");
+                args.Settings = settings = new ConsoleLoggerSettings()
+                {
+                    Switches =
+                    {
+                        ["Default"] = (LogLevel)(DateTimeOffset.Now.Second % 5 + 1),
+                        ["Microsoft"] = (LogLevel)(DateTimeOffset.Now.Second % 5 + 1),
+                    }
+                };
+            });
```
rynowak: /cc @divega - is this more of what you had in mind?

samples/SampleApp/ConfigurationRootExtensions.cs
```diff
@@ -0,0 +1,32 @@
...
+            callback = _ =>
+            {
+                var token = files.Watch(filename);
+                config.Reload();
+                token.RegisterChangeCallback(callback, null);
+            };
+            files.Watch(filename).RegisterChangeCallback(callback, null);
+            return config;
+        }
+    }
```
rynowak: Ignore this file, I've already opened a PR on configuration.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -2,23 +2,109 @@
...
+
+            _settings = settings;
+
+            if (_settings.ChangeToken != null)
+            {
+                _settings.ChangeToken.RegisterChangeCallback(OnConfigurationReload, null);
+            }
+        }
+
+        private void OnConfigurationReload(object state)
```
kichalla: nit: rename to `OnSettingsReload` (for my own understanding: as in this view there is no Configuration but just settings...which could come from anywhere where our Configuration system is one of them..)

rynowak: Sure

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -2,23 +2,109 @@
...
+        {
+            _settings.Reload();
```
kichalla: do we need to reload the values explicitly here...like if we there was a event(or change token) which said like `OnSettingsChanged`, we should be having a reloaded/changed settings already and we could just use them...
For example: https://github.com/aspnet/Configuration/pull/320/files#diff-3716cadb2612035150f3750b7e26ba58R84, we are already reloading them here and the `ConfigurationConsoleLoggerSettings` is just falling back on these changed settings...yeah, I know that settings could be built without configuration too, but thought we could make the `IConsoleLoggerSettings` interface a bit simpler...

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -2,23 +2,109 @@
...
+
+            foreach (var logger in _loggers.Values)
+            {
+                logger.Filter = GetFilter(logger.Name, _settings);
+                logger.IncludeScopes = _settings.IncludeScopes;
+            }
+
+            // The token will change each time it reloads, so we need to register again.
+            if (_settings?.ChangeToken != null)
```
kichalla: Can we avoid this `ChangeToken` pattern on the `IConsoleLoggerSettings` and just have a simple publisher/subscriber callback mechanism where I register a callback for something like `OnSettingsChanged`...`ConfigurationConsoleLoggerSettings` should internally watch for Configuration's ChangeToken and raise the event...

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
rynowak: > yeah, I know that settings could be built without configuration too, but thought we could make the IConsoleLoggerSettings interface a bit simpler...

Regardless of how you implement the settings you **need** to provide a new `IChangeToken` after signaling your current one. This handshake is required to have the correct concurrency behavior.

kichalla:
:shipit:

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
rynowak: We might discuss this more: https://github.com/aspnet/specs/blob/master/notes/reloadable_options.md#open-issues

But this PR isn't going to reinvent the pattern for using `IChangeToken`. If we find a way to simplify the pattern, we'll do that independently and clean this up later.

rynowak:
Checked in to release and merged to dev.

PR #277, opened 2015-10-28
Add test loggers for ILogger<T>

rynowak:
Inspired by https://github.com/aspnet/Home/issues/1042

/cc @kichalla

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/NullLoggerOfT.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/NullLoggerOfT.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Logging.Testing/NullLoggerOfT.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class NullLogger<T> : ILogger<T>
```
kichalla: There is a `NullLogger`...so `NullLogger<T> : NullLogger, ILogger<T>` should suffice?

kichalla:
:shipit:

src/Microsoft.Extensions.Logging.Testing/NullLoggerOfT.cs
rynowak: I didn't do this because none of the other loggers use inheritance between `<T>` and non-generic

PR #278, opened 2015-10-28
Simplify logic of applying colors to the messages

kichalla:
@rynowak I was reading about using struct and disposable pattern and came across following article which says that the the using pattern is optimized for value types where it does *not* box, but I thought its better to just simplify this
http://ericlippert.com/2011/03/14/to-box-or-not-to-box/
This is not for RC1

Following decompiled code from Telerik JustDecompile which contradicts the above article and is probably wrong
image


rynowak:
Eric Lippert knows more about the language than your decompiler.

rynowak:
:shipit: - as far as I'm concerned, do whatever you want here. Either of these versions of the code is correct and reasonable.

kichalla:
f6a2f4936161dbabcf6f7fe61373fa7f6f843d6b

PR #289, opened 2015-11-04
Clean up duplicated #if's for NET451

### LINKED ISSUE #287: Clean up duplicated #if's for NET451
Stuff like this:

```c#
#if NET451 || NET451 
            // whatever 
#endif 
```

E.g. here: https://github.com/aspnet/Logging/blob/cceb0dcc43e572ebcd3650b7628e052e22308b85/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceScope.cs#L24

Not sure if this appears in other repos, but I'm guessing due to mass search & replace this might be in a few other places.

pranavkm:
Fixes #287


rynowak:
:shipit:

Issue #272, opened 2015-10-23
Add LoggerMessage.Define overloads for 0, 4, 5, and 6 args


victorhurdugaci:
Please add a non generic version of `LoggerMessage.Define` because there are cases in which the logged message doesn't have any parameters

kichalla:
Ideally using `Define` is not required in the case where you are *not* using a message format string and do not have any parameters to pass. `Define` was optimized for these cases for perf reasons. 

That said, we currently have an extension for `DefineScope` which does not take in any arguments...probably this should be removed to keep it consistent with `Define` @rynowak 
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs#L20

Eilon:
We should have overloads for: no parameters, and also need to add overloads for 4, 5, and 6 args.

PR #296, opened 2015-11-18
Add more overloads to LoggerMessage.Define

pakrym:
@lodejard

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
@@ -69,6 +69,26 @@ namespace Microsoft.Extensions.Logging
         /// <summary>
         /// Creates a delegate which can be invoked for logging a message.
         /// </summary>
+        /// <param name="logLevel">The <see cref="LogLevel"/></param>
+        /// <param name="eventId">The event id</param>
+        /// <param name="formatString">The named format string</param>
+        /// <returns>A delegate which when invoked creates a log message.</returns>
+        public static Action<ILogger, Exception> Define(LogLevel logLevel, int eventId, string formatString)
+        {
+            var formatter = new LogValuesFormatter(formatString);
+
+            return (logger, exception) =>
+            {
+                if (logger.IsEnabled(logLevel))
+                {
+                    logger.Log(logLevel, eventId, new LogValues(formatter), exception, LogValues.Callback);
+                }
+            };
+        }
+
+        /// <summary>
+        /// Creates a delegate which can be invoked for logging a message.
+        /// </summary>
         /// <typeparam name="T1">The type of the first parameter passed to the named format string.</typeparam>
         /// <param name="logLevel">The <see cref="LogLevel"/></param>
         /// <param name="eventId">The event id</param>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -69,6 +69,26 @@ public static class LoggerMessage
         /// <summary>
         /// Creates a delegate which can be invoked for logging a message.
         /// </summary>
+        /// <param name="logLevel">The <see cref="LogLevel"/></param>
+        /// <param name="eventId">The event id</param>
+        /// <param name="formatString">The named format string</param>
+        /// <returns>A delegate which when invoked creates a log message.</returns>
+        public static Action<ILogger, Exception> Define(LogLevel logLevel, int eventId, string formatString)
```
kichalla: This extension should not be required as the intention of the extensions in this file is to optimize for the scenarios where we pass parameters to the format string (where as here we do not pass any)...

Example:
```
public static void OrderPlaced(this ILogger logger)
{
    if (logger.IsEnabled(LogLevel.Information)
    { 
        logger.LogInformation(eventId: 10, "Order placed");
    }
}
```

pakrym: True, but we get LoggingExtensions with two usage patterns inside which doesn't look good.

Eilon: @kichalla  I said it was ok :smile:

Eilon:
This addresses this issue: https://github.com/aspnet/Logging/issues/272

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
Eilon: @kichalla in this particular case it's more about consistency and not any actual perf concern.

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -268,6 +363,125 @@ public LogValues(LogValuesFormatter formatter, T0 value0, T1 value1, T2 value2,
...
+                new KeyValuePair<string, object>(_formatter.ValueNames[5], _value5),
+                new KeyValuePair<string, object>(_formatter.ValueNames[6], _value6),
+                new KeyValuePair<string, object>("{OriginalFormat}", _formatter.OriginalFormat),
+            };
+
+            public object[] ToArray() => new object[] { _value0, _value1, _value2, _value3, _value4, _value5, _value6 };
+
+            public override string ToString() => _formatter.Format(ToArray());
+        }
+
```
Eilon: :open_mouth: extra lines!

Eilon:
Is it possible to add at least a basic test for these?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
kichalla: Sound good...I agree.

pakrym:
@eilon will do

pakrym:
:up: :date:

test/Microsoft.Extensions.Logging.Test/LoggerMessageTest.cs
```diff
@@ -162,6 +163,48 @@ public void LogScope_WithThreeParameters()
                 actualLogValues.ToString());
         }
 
+        [Theory]
+        [MemberData(nameof(LogMessagesData))]
+
```
Eilon: Remove space :trollface:

test/Microsoft.Extensions.Logging.Test/LoggerMessageTest.cs
```diff
@@ -162,6 +163,48 @@ public void LogScope_WithThreeParameters()
...
+            var testSink = new TestSink();
+            var testLogger = new TestLogger("testlogger", testSink, enabled: true);
+            var exception = new Exception("TestException");
+            var parameterNames = Enumerable.Range(0, argumentCount).Select(i => "P" + i).ToArray();
+            var parameters = new List<object>();
+            parameters.Add(testLogger);
+            parameters.AddRange(parameterNames);
+            parameters.Add(exception);
+
+            var expectedFormat = "Log " + string.Join(" ", parameterNames.Select(p=>"{"+p+"}"));
```
Eilon: Format codes :trollface:

Eilon:
:shipit: after code formatting.

PR #297, opened 2015-11-19
Add IncludeTimestamp and TimestampFormat

prasannavl:
ConsoleLogger now has the capability to include timestamps along with the message and with the optional format.


muratg:
@prasannavl Thanks for the PR, although we don't want add more to the ConsoleLogger.  You should be able to create a Wrapper logger to add this functionality and package it and release it on NuGet for other though!

PR #307, opened 2015-11-30
Update README.md

Kagamine:


Eilon:
Is there a purpose to this change? BTW the `aspnetcidev` should generally not be used by anyone outside of core contributors to ASP.NET, so I don't think we want to publicize that beyond any core needs.

Issue #298, opened 2015-11-20
Remove ILoggerFactory.MinimumLevel


Eilon:
This property has two things going against it:
1. Everyone thinks it does something other than what it does. (This can be mitigated by a rename of some sort.)
2. It's not needed anymore. It was added to avoid sensitive information being logged inadvertently. This was later decided to be a non-issue.

This is teh codez: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs#L14-L17

@lodejard @muratg @DamianEdwards

lodejard:
:+1: this should have been removed a while ago!

muratg:
Agreed!

@JunTaoLuo :gift:

JunTaoLuo:
merged to dev

joeaudette:
so is there a new way to set the loglevel?

Eilon:
@joeaudette the thing is, this never set the log level. That's why we're removing it. It just sets the minimum *allowed* log level. Each individual logger (e.g. DebugLogger, ConsoleLogger, EventLogLogger, SerilogLogger, etc.) has their own configuration and settings to decide what they track.

joeaudette:
that makes sense, Thanks!

joeaudette:
just to help others since breaking change announcements usually provide the old and the new way of doing things.
old way:

    loggerFactory.MinimumLevel = LogLevel.Warning;
    loggerFactory.AddConsole();

new way:

    loggerFactory.AddConsole(minLevel: LogLevel.Warning);

JunTaoLuo:
@joeaudette thanks for the suggestion.

sunsided:
@joeaudette that makes things clearer indeed.

Eilon:
BTW, you *always* needed `loggerFactory.AddConsole(minLevel: LogLevel.Warning);` to get the right behavior. So in theory a properly-configured app should just remove the call to `loggerFactory.MinimumLevel`, but everyone was confused, so it was more complicated :smile:

PR #308, opened 2015-12-02
Removing MinimumLevel from ILoggerFactory #298

JunTaoLuo:
#298 

I will update the following repos:

- EntityFramework https://github.com/aspnet/EntityFramework/pull/3956
- KestrelHttpServer https://github.com/aspnet/KestrelHttpServer/pull/445
- SignalR-Redis https://github.com/aspnet/SignalR-Redis/pull/21

These repos are also affected but they are decoupled from our builds and most of them are updated only when we release. Also some don't seem relevant.

- EntityFramework.Benchmark
- NodeServices
- MusicStoreAngular2
- live.asp.net
- go.asp.net
- get.asp.net
- Docs
- KExpense
- PackageSearch

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
@@ -12,11 +12,6 @@ namespace Microsoft.Extensions.Logging
     public interface ILoggerFactory : IDisposable
     {
         /// <summary>
-        /// The minimum level of log messages sent to loggers.
-        /// </summary>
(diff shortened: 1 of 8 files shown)


Eilon:
:shipit:

Should definitely update all the repos you mention. BTW they should be updated *before* this gets merged to avoid build breaks (this is one of those rare cases where we can avoid any breaks at all).

JunTaoLuo:
@Eilon I doubt the other repos can be updated first since removing `MinimumLevel` will result in the interface to not be completely implemented until the changes in Logging is propagated.

Issue #309, opened 2015-12-02
NullReferenceException in RuntimeEnvironmentHelper under .NET 4.5 using Logging


adamhathcock:
I wrote this under DNX but maybe it's more appropriate here: https://github.com/aspnet/dnx/issues/3218

Basically, the usage of a DNX specific thing causes an exception when trying to use Console logging.  I have a sample and more details on the other issue.

davidfowl:
Yep, it's a bug. It should be resolved in RC2.

PR #310, opened 2015-12-02
Avoid DNXisms when unneeded

davidfowl:
- Use RuntimeInformation to detect the OS

#309

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections;
+using System.Runtime.InteropServices;
 using System.Text;
-using Microsoft.Dnx.Runtime;
 using Microsoft.Extensions.Logging.Console.Internal;
 
 namespace Microsoft.Extensions.Logging.Console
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -42,7 +42,7 @@ public ConsoleLogger(string name, Func<string, LogLevel, bool> filter, bool incl
             Filter = filter ?? ((category, logLevel) => true);
             IncludeScopes = includeScopes;
 
-            if (RuntimeEnvironmentHelper.IsWindows)
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
borgdylan: I do not know how this is compiling under dnx451....but since the type does not exist anywhere but on CoreFX, mono is failing due to not being able to load the RuntimeInformation type. Also notice that travis had failed on this PR.

davidfowl: Travis actually passed https://travis-ci.org/aspnet/Logging/builds/94358265

borgdylan: Why did you link the runtime info api as a build only dep on dnx451? The library is not being found at runtime. I might try adding it as a dep myself to see if stuff gets fixed.

borgdylan: No issues occur if I add
```
"runtime.linux.System.Runtime.InteropServices.RuntimeInformation": "4.0.0-*"
```

to the dependency list.

PR #317, opened 2015-12-10
Add Log4Net

fengjb:


Eilon:
@fengjb thanks for the contribution but we are not looking to add providers for 3rd party loggers at this time. Instead, we'd like logging systems to have the providers and then we will link to them from https://github.com/aspnet/Logging/blob/dev/README.md.

PR #321, opened 2015-12-15
Use a platform determiner which is available on all platforms

ryanbrandenburg:
Should fix test failures in Entropy on Mac.


kichalla:
Travis ci seems to fail with PR... :shipit: once its green

kichalla:
It would be nice if we can make this repo also run on Mac (travis.yml) as its a core framework piece

Issue #311, opened 2015-12-02
Regression: The logging infrastructure crashes and disallows app startup


borgdylan:
Some functionality like action link generation was changed in Mvc and in other areas. The logging packages were not changed to reflect this. This results in failing application startups as the required types cannot be loaded.

kichalla:
Can you give more details...what kind of exception are you seeing?

borgdylan:
```
Application startup exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.IO.FileNotFoundException: Could not load file or assembly 'System.Runtime.InteropServices.RuntimeInformation, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
File name: 'System.Runtime.InteropServices.RuntimeInformation, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
  at Microsoft.Extensions.Logging.Console.ConsoleLoggerProvider.CreateLoggerImplementation (System.String name) <0xb02675e0 + 0x0008b> in <filename unknown>:0 
  at System.Collections.Concurrent.ConcurrentDictionary`2[TKey,TValue].GetOrAdd (System.Collections.Concurrent.TKey key, System.Func`2 valueFactory) <0xb50956d0 + 0x00063> in <filename unknown>:0 
  at Microsoft.Extensions.Logging.Console.ConsoleLoggerProvider.CreateLogger (System.String name) <0xb0267498 + 0x0008f> in <filename unknown>:0 
  at Microsoft.Extensions.Logging.Logger.AddProvider (ILoggerProvider provider) <0xb02673b8 + 0x00022> in <filename unknown>:0 
  at Microsoft.Extensions.Logging.LoggerFactory.AddProvider (ILoggerProvider provider) <0xb0266f70 + 0x00163> in <filename unknown>:0 
  at Microsoft.Extensions.Logging.ConsoleLoggerExtensions.AddConsole (ILoggerFactory factory, System.Func`3 filter, Boolean includeScopes) <0xb0266bf0 + 0x0004e> in <filename unknown>:0 
  at Microsoft.Extensions.Logging.ConsoleLoggerExtensions.AddConsole (ILoggerFactory factory, Boolean includeScopes) <0xb0266b18 + 0x000b3> in <filename unknown>:0 
[...]

PR #324, opened 2015-12-22
Remove dependency on System.Runtime.InteropServices.RuntimeInformation

kichalla:
@ryanbrandenburg I see that your commit had more changes than this one. Not sure why they were required, but here after making this change I built EF against this change and it built just fine...so I assume all is good to check this in.

Should fix https://github.com/aspnet/Logging/issues/311

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -6,6 +6,7 @@ using System.Collections;
 using System.Runtime.InteropServices;
 using System.Text;
 using Microsoft.Extensions.Logging.Console.Internal;
+using Microsoft.Extensions.PlatformAbstractions;
 
 namespace Microsoft.Extensions.Logging.Console
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -6,6 +6,7 @@
 using System.Runtime.InteropServices;
```
ryanbrandenburg: Doesn't this line need to go too?

davidfowl:
@kichalla I don't think we should make this change. We need to fix RuntimeInformation (on corefx) if it doesn't work.

ryanbrandenburg:
@kichalla the problem I ran into that made me try all the other fixes was that we were getting a mismatched System.Runtime (20 vs 21). If that doesn't show up for you on mono, coreclr, windows mac or linux then guess you're good to go, though I've no idea what makes yours work and mine not.

ryanbrandenburg:
I do agree with @davidfowl that making inerop show up on linux is the preferable fix, though I've no suggestion on how to do that (last time I just got lucky and @anurse knew the answer)

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
kichalla: :+1:

kichalla:
@davidfowl I agree that probably the right thing to do is fix it in CoreFx but can we file a bug in CoreFx and check this in as a temporarily...we already seem to be using this in several places: 

https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Tooling/Program.cs#L46
https://github.com/aspnet/dnx/blob/dev/src/Microsoft.Dnx.Runtime.Sources/Impl/RuntimeEnvironmentHelper.cs

muratg:
This is a simple workaround until CoreFX can get their side done.  Let's just get it in to unblock our cross plat builds.

:shipit:

kichalla:
I pushed this change and MusicStore now builds runs on Mac

kichalla:
f7c9fc32a801bc0c756931adfdca73ab4657727a

PR #330, opened 2015-12-29
Split large messages when logging into EventLog

kichalla:
@Eilon

### DIFF
--- a/src/Microsoft.Extensions.Logging.EventLog/EventLogSettings.cs
+++ b/src/Microsoft.Extensions.Logging.EventLog/EventLogSettings.cs
@@ -32,7 +32,7 @@ namespace Microsoft.Extensions.Logging.EventLog
         public Func<string, LogLevel, bool> Filter { get; set; }
 
         /// <summary>
-        /// For unit testing purposes only
+        /// For unit testing purposes only.
         /// </summary>
         public IEventLog EventLog { get; set; }
     }
--- a/src/Microsoft.Extensions.Logging.EventLog/EventLogSettings.cs
+++ b/src/Microsoft.Extensions.Logging.EventLog/EventLogSettings.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Extensions.Logging.EventLog.Internal;
 
 namespace Microsoft.Extensions.Logging.EventLog
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Logging.EventLog/Internal/IEventLog.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Diagnostics;
+
+namespace Microsoft.Extensions.Logging.EventLog.Internal
+{
+    public interface IEventLog
+    {
+        int MaxMessageSize { get; }
```
Eilon: Why have a property for this? Might be easier to hard-code it in the calling class, and the unit tests can use the real number. (Yeah it means huge strings, but is that a real problem?)

src/Microsoft.Extensions.Logging.EventLog/EventLogSettings.cs
```diff
@@ -29,5 +30,10 @@ public class EventLogSettings
         /// The function used to filter events based on the log level.
         /// </summary>
         public Func<string, LogLevel, bool> Filter { get; set; }
+
+        /// <summary>
+        /// For unit testing purposes only
```
Eilon: End sentence with a period

src/Microsoft.Extensions.Logging.EventLog/Internal/IEventLog.cs
kichalla: Using huge strings is not a real problem but I think having small max size makes the test easy to read (For example: https://github.com/aspnet/Logging/pull/330/files#diff-5f8faf58806279e51869fddb0ff57828R155)

Eilon: I actually found those tests a bit difficult to read :smile: It uses continuationString in some places but not others.

I think if it's going to go that route, I'd avoid "refactoring" the code too much and have the expected strings be more like:

```c#
"Test" + Environment.NewLine + "a...",
"...aaaaaa...",
"...aaaaaa...",
"...aaaaa",
```

Because that would really *show* the behavior, as opposed to *computing* the behavior.

kichalla: Sure, I agree...will change it...BTW, I noticed that my tests were failing in Travis due to new line representations being different in Linux(`\n` instead of `\r\n`)...I temporarily enabled this test to run only on Windows as EventLog is a windows concept...this is because of the same reason that computing the strings(taking into account the different newline lenghts) would make the test not easy to read and itself can have bugs...

Eilon: I'm fine disabling the entire test class on non-Windows...

Eilon:
:shipit: as-is or even disable on non-Windows, I'm fine either way.

kichalla:
6c29db36515ccd86dc367b5b07d790c2ff7a898b

PR #336, opened 2016-01-06
Build with dotnet

victorhurdugaci:
cc @pranavkm @BrennanConroy

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,18 @@
 #!/usr/bin/env bash
 
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+repoFolder="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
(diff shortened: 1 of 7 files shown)


build.sh
```diff
@@ -11,33 +24,25 @@ else
...
+    mono $nugetPath install KoreBuild-dotnet -ExcludeVersion -o $buildFolder -nocache -pre
 fi
 
-if ! type dnx > /dev/null 2>&1 || [ -z "$SKIP_DNX_INSTALL" ]; then
-    dnvm install latest -runtime coreclr -alias default
-    dnvm install default -runtime mono -alias default
-else
-    dnvm use default -runtime mono
-fi
+source $koreBuildFolder/build/KoreBuild.sh
```
BrennanConroy: I don't think this actually does anything, as soon as you exit this script it is no longer sourced

test/Microsoft.Extensions.Logging.Testing.Tests/project.json
```diff
@@ -14,12 +16,16 @@
     "dnx451": {
```
davidfowl: We can probably start changing this to net451 since we're using the console runner

build.sh
BrennanConroy: Ok, I guess it does stuff, but we could just run KoreBuild.sh instead of sourcing

victorhurdugaci: That's the idea. That you source it while building and then goes away

test/Microsoft.Extensions.Logging.Testing.Tests/project.json
victorhurdugaci: I would prefer to do that in one big pass rather than having some repos with net451 and some with dnx451

Issue #331, opened 2015-12-31
Changes to Logging API surface


muratg:
- Have all levels (`Critical`, `Debug`, `Error`, `Warning`, `Information`, `Verbose`) have the same overloads as extension methods: 
  - `Log....(this ILogger logger, EventId eventId, Exception exception, string message, params object[] args);`
  - `Log....(this ILogger logger, EventId eventId, string message, params object[] args);`
  - `Log....(this ILogger logger, string message, params object[] args);`
- `EventId` has implicit conversion from/to Int and if you want to use string, you do the conversion yourself
- Our frameworks will be passing in both `int` and `string` based `EventId`s. 
- Event IDs (string versions) in our frameworks should be PascalCased and should typically be two words. Some examples:
  - "QueryExecuting"
  - "ActionExecuting"
  - "ContentNegotiationFailed"
  - ...
- Change the `LoggerMessage.Define` to take an `EventId` instead of `int`
  - There may be other places where `int eventId` could be used. Convert them to EventId.
- `MessageFormatter` should be state to `string` (and not do formatting at all) 
- `MessageFormatter` state should never be null, null check shouldn’t be done in runtime
- Remove `ILogValues` and use `IReadOnlyList<KeyValuePair<string, object>>`
- `ILogger` change: Instead of `void Log(LogLevel logLevel, int eventId, object state, Exception exception, Func<object, Exception, string> formatter)` we should use `void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)`

muratg:
cc @Eilon @rynowak @lodejard @davidfowl @divega @bricelam

muratg:
We will start this work shortly.

NickCraver:
Might I suggest a `[breaking-change]` label rather than (or in addition to) `[bug]` for these? That will help library authors a great deal in filtering what's most important in the future.

davidfowl:
We do that on the announcements repository not here

NickCraver:
@davidfowl that may work for the Microsoft side, but consumers aren't going to keep checking the monolithic announcements repo and filter out the noise to see what affects them on every change. Have you actually tried to consume parts of that? What's the harm in a label on the specific issues as well? It's really very little overhead and is beneficial. On top of that, bug is incorrect - most of these items aren't bugs.

davidfowl:
> Have you actually tried to consume parts of that?
Yes, it's actually very good for breaking API changes.

[...]

PR #338, opened 2016-01-19
Logging API changes

BrennanConroy:
#331 

Still have some changes to make, and need to review it myself.

@Tratcher @muratg @lodejard @davidfowl


src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
```diff
@@ -0,0 +1,32 @@
...
+        private int _id;
+        private string _name;
+
+        public EventId(int id, string name = null)
+        {
+            _id = id;
+            _name = name;
+        }
+
+        public static implicit operator int(EventId e)
```
Tratcher: Do these converters need to be bidirectional? Isn't this just to help you call logger.Write(12,...)? When do you need to convert from EventId to int?

src/Microsoft.Extensions.Logging.Abstractions/ILogValues.cs
```diff
@@ -5,11 +5,11 @@
...
 namespace Microsoft.Extensions.Logging
 {
-    public interface ILogValues
-    {
-        /// <summary>
-        /// Returns an enumerable of key value pairs mapping the name of the structured data to the data.
-        /// </summary>
-        IEnumerable<KeyValuePair<string, object>> GetValues();
-    }
+    //public interface ILogValues
```
Tratcher: ?

src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
```diff
@@ -780,44 +362,28 @@ public static void LogCritical(this ILogger logger, int eventId, string message,
...
-        private static void LogWithEvent(
-            this ILogger logger,
-            LogLevel logLevel,
-            int eventId,
-            ILogValues state,
-            Exception exception = null)
-        {
-            logger.Log(logLevel, eventId, state, exception, _messageFormatter);
-        }
+        //private static void Log(
```
Tratcher: ?

src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            return e._id;
+        }
+
+        public static implicit operator EventId(int i)
+        {
+            return new EventId(i);
+        }
+
+        public static explicit operator string(EventId e)
```
lodejard: Should have Id and Name get-only properties. Also to @Tratcher's question, we might not need the implicit operators for the int/string values.

src/Microsoft.Extensions.Logging.Abstractions/ILogValues.cs
lodejard: Does `?` mean delete commented code?

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -144,6 +144,16 @@ public string Format(object[] values)
             return string.Format(CultureInfo.InvariantCulture, _format, values);
         }
 
+        public KeyValuePair<string, object> GetValue(object[] values, int index)
+        {
+            if (values.Length < index)
```
lodejard: hmmm - maybe also need a range exception when index < 0 or index > Length?

src/Microsoft.Extensions.Logging.Abstractions/LogFormatter.cs
```diff
@@ -27,7 +27,7 @@ public static string Formatter(object state, Exception e)
             var result = string.Empty;
             if (state != null)
             {
-                var values = state as ILogValues;
+                var values = state as IReadOnlyList<KeyValuePair<string, object>>;
```
lodejard: Hmmm... I didn't notice this before. It turns out the entire LogFormatter class should be deleted - there is no case where we should have a standard name-value string builder which appends the exception with a newline. Can you delete this file and see how much depended on it?

src/Microsoft.Extensions.Logging.Abstractions/ILogValues.cs
davidfowl: yah

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -112,14 +113,18 @@ public void Log(LogLevel logLevel, int eventId, object state, Exception exceptio
             }
             else
             {
-                message = LogFormatter.Formatter(state, exception);
+                message = state?.ToString();
```
BrennanConroy: POI - is this acceptable?
Need a default if formatter is null and since we removed LogFormatter

BrennanConroy:
:up: :date:

lodejard:
:shipit:

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -192,15 +217,10 @@ public IDisposable BeginScopeImpl(object state)
             return ConsoleLogScope.Push(Name, state);
         }
 
-        private void FormatLogValues(StringBuilder builder, ILogValues logValues, int level, bool bullet)
+        private void FormatLogValues(StringBuilder builder, IReadOnlyList<KeyValuePair<string, object>> logValues, int level, bool bullet)
```
JunTaoLuo: What's the point of this method. It's private and it's only used by itself.

JunTaoLuo: Just checked again, this is not used anywhere. Remove it.

Issue #351, opened 2016-01-28
LogInformation(string) throws FormatException


Tratcher:
https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Server.Testing/Deployers/SelfHostDeployer.cs#L81
`Logger.LogInformation(dataArgs.Data ?? string.Empty);`
```
System.AggregateException: An error occurred while writing to logger(s). (Index (zero based) must be greater than or equal to zero and less than the size of the argument list.) ---> System.FormatException: Index (zero based) must be greater than or equal to zero and less than the size of the argument list.
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(IFormatProvider provider, String format, Object[] args)
   at Microsoft.Extensions.Logging.Internal.LogValuesFormatter.Format(Object[] values)
   at Microsoft.Extensions.Logging.Internal.FormattedLogValues.ToString()
   at Microsoft.Extensions.Logging.LoggerExtensions.MessageFormatter(Object state, Exception error)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   --- End of inner exception stack trace ---
   at Microsoft.Extensions.Logging.Logger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
   at Microsoft.Extensions.Logging.LoggerExtensions.LogInformation(ILogger logger, String message, Object[] args)
   at Microsoft.AspNetCore.Server.Testing.SelfHostDeployer.<StartSelfHost>b__3_1(Object sender, DataReceivedEventArgs dataArgs)
   at System.Diagnostics.Process.OutputReadNotifyUser(String data)
```

If the data contains curly brackets then logging may throw a format exception. The logger should not attempt to format data for a simple string with no parameters.
https://github.com/aspnet/Logging/blob/4528bdbc2816a8b5f933d9c6cf2da806ce4a72ea/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L174

This failed some tests in MusicStore

[...]

PR #354, opened 2016-01-29
Fix case where no args passed to Log* extensions

BrennanConroy:
#351 
@Tratcher @lodejard


Tratcher:
Why even create the FormattedLogValues when there are no params? You could do the check in LoggerExtensions and call Log with just the string.

BrennanConroy:
Verbal squirrel from @lodejard

Issue #355, opened 2016-01-29
Exception details are not logged


moozzyk:
Stack trace is not written when an exception happens so it can be very hard to tell where it originated from:

```
dbug: Microsoft.AspNetCore.Mvc.Razor.Internal.RoslynCompilationService[1]
      Compilation of the generated code for the Razor file at '/Views/Home/Index.cshtml' started.
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      An unhandled exception was thrown by the application.
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      An unhandled exception was thrown by the application.
      System.InvalidOperationException
      Can not resolve assembly lib\dotnet5.4\Microsoft.AspNetCore.Antiforgery.dll location
      System.InvalidOperationException
      Can not resolve assembly lib\dotnet5.4\Microsoft.AspNetCore.Antiforgery.dll location
info: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 1727.9688ms 200
```

BrennanConroy:
Fixed via 46c6d523fd323eba101f230b96111d3fa4dcfa46

PR #356, opened 2016-01-29
Print exception not just message

BrennanConroy:
#355 
@moozzyk @lodejard


davidfowl:
screnshot?

BrennanConroy:
ex

lodejard:
I guess if Exception.ToString is always going to show the full type, then showing the exception type in its own line is redundant. Also if the call stack is indented there's no reason to indent the exception itself. the stack frame indentation should provide the visual break to the next line.

BrennanConroy:
ex

BrennanConroy:
Ping @davidfowl @lodejard

muratg:
:eyes: :watch:

lodejard:
:rotating_light: :moneybag: :ok_hand: :shipit: :smile_cat:

PR #360, opened 2016-02-11
ILoggerFactory does not need to require IDisposable

### LINKED ISSUE #319: ILoggingFactory should not implement IDisposable
A similar issue has been raised [here](https://github.com/aspnet/Mvc/issues/3726). Abstractions should not implement `IDisposable`, as @nblumhardt explained this very clearly almost six years ago (and quoted by @ploeh):

> an interface [...] generally shouldn't be disposable. There's no way for the one defining an interface to foresee all possible implementations of it - you can always come up with a disposable implementation of practically any interface.

While the default `LoggingFactory` implementation can (and probably must) be disposable, the interface shouldn't be.

lodejard:
Closes #319

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Extensions.Logging
     /// Represents a type used to configure the logging system and create instances of <see cref="ILogger"/> from
     /// the registered <see cref="ILoggerProvider"/>s.
     /// </summary>
-    public interface ILoggerFactory : IDisposable
+    public interface ILoggerFactory
     {
         /// <summary>
         /// Creates a new <see cref="ILogger"/> instance.
--- a/src/Microsoft.Extensions.Logging.Abstractions/ILoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/ILoggerProvider.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Extensions.Logging
     /// <summary>
     /// Represents a type that can create instances of <see cref="ILogger"/>.
     /// </summary>
-    public interface ILoggerProvider : IDisposable
+    public interface ILoggerProvider 
     {
         /// <summary>
         /// Creates a new <see cref="ILogger"/> instance.
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
@@ -108,9 +108,5 @@ namespace Microsoft.Extensions.Logging.Console
                 name = name.Substring(0, lastIndexOfDot);
             }
         }
-
-        public void Dispose()
-        {
-        }
     }
 }
--- a/src/Microsoft.Extensions.Logging.Debug/DebugLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Debug/DebugLoggerProvider.cs
@@ -26,9 +26,5 @@ namespace Microsoft.Extensions.Logging.Debug
         {
             return new DebugLogger(name, _filter);
         }
-
-        public void Dispose()
-        {            
-        }
     }
 }
--- a/src/Microsoft.Extensions.Logging.EventLog/EventLogLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.EventLog/EventLogLoggerProvider.cs
@@ -34,9 +34,5 @@ namespace Microsoft.Extensions.Logging.EventLog
         {
             return new EventLogLogger(name, _settings ?? new EventLogSettings());
         }
-
-        public void Dispose()
-        {
-        }
     }
 }
--- a/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceLoggerProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Logging.TraceSource
     /// <summary>
     /// Provides an ILoggerFactory based on System.Diagnostics.TraceSource.
     /// </summary>
-    public class TraceSourceLoggerProvider : ILoggerProvider
+    public class TraceSourceLoggerProvider : ILoggerProvider, IDisposable
     {
         private readonly SourceSwitch _rootSourceSwitch;
         private readonly TraceListener _rootTraceListener;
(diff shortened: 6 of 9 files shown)


Eilon:
@lodejard can you update this? Was this the PR where you missed adding back `IDisposable` to the concrete class?

lodejard:
Ah, yep! I need to push the 2nd commit.

davidfowl:
:shipit:

lodejard:
After implementing this and seeing the implications - I don't think it should be merged after all.

If you implement ILoggerProvider you aren't compelled to implement a Dispose method, which makes it slightly less discoverable.

On the other hand if an ILoggerProvider can even optionally rely on IDispose in order to correctly flush buffers on exit - it's only the need to implement it is optional. The need to call it is still mandatory. In other words the default ILoggerFactory implementation, and anything which replaces it, **must** call dispose to satisfy the requirements of using logger providers.

In other words - IDisposable *is* actually part of the contract of being an ILoggerFactory because gracefully flushing buffers by the ILoggerProvider is a required and relied upon feature.

So... I'm giving myself a -1 on my PR :)

Eilon:
@dotnetjunkie it would be good to get your thoughts on @lodejard 's latest comment. I talked to @lodejard and it does seem like `IDisposable` is a contractual part of the `ILoggerFactory` contract.

dotnetjunkie:
> IDisposable is actually part of the contract of being an ILoggerFactory because gracefully flushing buffers by the ILoggerProvider is a required and relied upon feature.

What you are basically saying is that since the default implementation needs flushing, so does the abstraction. This reasoning however is flawed. You are now leaking implementation details into the abstraction. With that reasoning, the abstraction still needs IDisposable, even if the default implementation doesn't need it, because there will always be an implementation that might need to dispose. With this reasoning, actually any abstraction we can think of will need to implement IDisposable, since we can always think of an implementation that needs to dispose.

So the right question is: does EVERY implementation of ILoggerFactory require disposing? If the answer is 'yes', we do need IDisposable on the abstraction.

Chances however that you can answer 'yes' are really slim, because we can always think of inplementations that don't need disposal. Take fake or mock implementations or we can easily build an implementation that doesn't require flushing at all.

So the Dependency Inversion Principle still stands: don't let abstractions depend on implementation details.

lodejard:
> So the right question is: does EVERY implementation of ILoggerFactory require disposing? If the answer is 'yes', we do need IDisposable on the abstraction.

Yes - EVERY implementation if ILoggerFactory does require disposing. And that is why IDisposable is on the abstraction.

The reason for this is that every implementation of ILoggerFactory will be given the same ILoggerProvider. There exist providers which MUST be disposed to function correctly. The ONLY way to execute code to dispose the providers is to be disposable.

Any ILoggerFactory which manages an ILoggerProvider collection must be IDisposable. All ILoggerFactory must manage an ILoggerProvider collection, because the AddProvider method MUST be implemented. Therefor ALL ILoggerFactory must be IDisposable.

This is not simply a detail of the default implementation.

> Take fake or mock implementations or we can easily build an implementation that doesn't require flushing at all.

You can always create an object which implements a subset of a service contract for unit test cases. Just because Dispose is not needed for all unit tests doesn't mean Dispose isn't mandatory for all ILoggerFactory implementations.

dotnetjunkie:
> The reason for this is that every implementation of ILoggerFactory will be given the same set of ILoggerProvider implementations. 

But if the factory is supplied with existing logging provider instances, why is it responsible of disposing them? This is not a typical practice, because the factory has no way of knowing what the lifetime is of its dependencies; they might outlive the factory (just image someone registering a scoped factory, or sharing providers across multiple factories). In other words, if someone wants to use his provider after the factory gets disposed, the application would break.

Instead, usual practice is to keep the ownership with the entity who is responsible of creating such instance, which in our case probably is the IoC container. It already has sophisticated handling of lifetimes (and wil likely guarantee that all disposables are disposed, even if there is some disposable that throws an exception from within its dispose method). So if the providers are registered within the container, there is no need for the factory to dispose them (they would be disposed twice) and the factory can (and probably should) do without a dispose method. And do note that anyone who implements a custom factory must implement the same behavior of guaranteed disposal of its dependencies (as the IoC container does), which makes it actually pretty hard to implement such a simple factory correctly.

davidfowl:
> Instead, usual practice is to keep the ownership with the entity who is responsible of creating such instance, which in our case probably is the IoC container.

Except the Logger system is decoupled from IoC so that argument doesn't hold. It works well with the DI system of course but it isn't dependent on it.

dotnetjunkie:
> Except the Logger system is decoupled from IoC so that argument doesn't hold

I think the argument still holds, because even without an IoC container, management of dependencies would be the same; the one who creates an object is responsible of destroying it. All the defects I described still exist if you don't have a container.

lodejard:
> the one who creates an object is responsible of destroying it

In other words - it is always incorrect for a component to dispose items in a collection it maintains if the items in the collection are passed in? That in all cases it is literally the class which called new that must call dispose? That clearly can't be the semantics of IDisposable.

In this case we will also need to change the default LoggerFactory to not dispose the providers which are added to it - instead the loggerFactory.AddFoo() extensions methods to add providers should return an IDisposable which the caller must dispose when the application shuts down.

davidfowl:
The object that calls new does not have to be the one to call dispose. Objects can delegate lifetime ownership to other objects. In this case, `ILoggerFactory` is responsible for the lifetime of `ILoggerProvider` since there's no easy way (right now) to register and dispose these objects.

dotnetjunkie:
> In other words - it is always incorrect for a component to dispose items in a collection it maintains if the items in the collection are passed in? That in all cases it is literally the class which called new that must call dispose?

No, it is not always incorrect, but you will have to be very careful nreaking this rule. Ownership can be transferred, but just think about how developers keep tripping over the StreamReader class that closes the underlying stream that is passed in from the outside. Its unintuitive for developers, and in .NET 1.0, developers had to jump through hoops to prevent their stream from being disposed. This has improved in v2, but the core problem still exists.

There are other exceptions and there are sometimes good reasons to break this guideline, but again, care must be taken. In most cases you will annoy and upset many developers and can cause tremendous costs because of the time they will spend in tracking down the source of the bug in their code.

> In this case we will also need to change the default LoggerFactory to not dispose the providers which are added to it

In case the factory creates providers itself, and keeps a hold of them, it would be natural for it to be responsible of disposing those created providers. And if this is difficult to achieve this with an extension method, perhaps an extension method is not the best way to solve this problem.

I hope you see there's much more detail to consider than first meets the eye. Resource management is hard, even with IDosposable.

lodejard:
Well, in a way there is an easy way right now - calling AddProvider registers the object for disposal :)

Eilon:
See https://github.com/aspnet/Logging/issues/319#issuecomment-196121767.

PR #365, opened 2016-02-19
Enabled xml doc generation

ajaybhargavb:
@NTaylorMullen

### DIFF
--- a/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceScope.cs
+++ b/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceScope.cs
@@ -14,11 +14,18 @@ namespace Microsoft.Extensions.Logging.TraceSource
         // To detect redundant calls
         private bool _isDisposed;
 
+#if NET451
         /// <summary>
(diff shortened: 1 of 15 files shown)


src/Microsoft.Extensions.Logging.TraceSource/TraceSourceScope.cs
```diff
@@ -14,11 +14,13 @@ internal class TraceSourceScope : IDisposable
         // To detect redundant calls
         private bool _isDisposed;
 
+#if NET451
```
NTaylorMullen: Hmmm, should still have some sort of doc comments for non-net451. Ditto elsewhere

NTaylorMullen:
:watch:

src/Microsoft.Extensions.Logging.TraceSource/TraceSourceScope.cs
ajaybhargavb: Added.

ajaybhargavb:
:up: :date:

NTaylorMullen:
:shipit:

PR #372, opened 2016-02-25
Make ServiceCollectionExtensions consistent

javiercn:


src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
```diff
@@ -3,16 +3,28 @@
 
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using System;
```
BrennanConroy: order

javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

Eilon:
:shipit: after fixing @BrennanConroy 's item.

Issue #370, opened 2016-02-23
Keep it simple (abstraction only)


ejsmith:
I really like the ILogger and ILoggerFactory interfaces. They are great abstractions that the community can build upon. The problem is that this package includes some extra things that are opinionated and that not everyone agrees with. Can we please move those extension methods to another package and keep this package to exactly what it says it is?

ejsmith:
Any thoughts on this? I know you guys don't want to make any more breaking changes, but I personally think this is very important. Let's just keep this package extremely simple with just the abstractions. It should be pretty easy to migrate projects if another package is created with the extensions in this package. The migration would just be a matter of adding another package to the project.

niemyjski:
+1

herecydev:
> The problem is that this package includes some extra things that are opinionated and that not everyone agrees with

It's probably best to quantify that surely?

ejsmith:
So basically there are a ton of extension methods here:

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs

and here:

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerFactoryExtensions.cs

I just want this package to be the core interfaces only because a lot of people have differing opinions on what these extension method signatures should look like and there is also another whole use case of people creating strongly typed logger extension methods on top of the ILogger and not wanting to use any of these extension methods at all.

They can't make that choice because they have to bring this package in and use the namespace that they ILogger types are in. Which is also the same namespace as the ILogger extension methods are in. Which means I have no choice but to have all of them on my logger classes.

herecydev:
Then again, from my experience these are invaluable. I don't want to log out the `EventId` or worry about the formatter, which makes methods like [this](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L61) invaluable.

It's hard to make a one-size-fits-all package that isn't going to upset some people

niemyjski:
@herecydev if everything is a package we could just add this to a logger.extensions package and then you can bring it in and then it's a buy in and every one is happy.

ejsmith:
The extension methods would live in a different package. I'm talking about just simplifying the abstractions package. I think most people will install the Microsoft.Extensions.Logging package anyway and it could have the extension methods in it or it could have a dependency that brings them in. Most people would never know.

[...]

PR #373, opened 2016-02-26
Move interfaces into their own package.

ejsmith:
This is a small change and should not affect any of the projects using this library, but it could have a huge impact on the logging fragmentation on .NET. If we can get everyone to agree on these very simple interfaces ILogger and ILoggerFactory then we can avoid the hell of having to create a bunch of proxies that translate between the various logging frameworks.


olepichon-systran:
I know about https://github.com/aspnet/dnx/issues/2003 but this kind of interfaces would have been great candidates for AssemblyNeutral usage.

MuiBienCarlota:
I know about aspnet/dnx#2003 but this kind of interfaces would have been great candidates for AssemblyNeutral usage.

ejsmith:
Yeah, absolutely. Bummer that they got cut.

304NotModified:
@ejsmith is this for https://github.com/aspnet/Logging/issues/370?

ejsmith:
Yes it is

muratg:
We won't be making this change as it's a big breaking change.

PR #376, opened 2016-03-01
Don't reference design-time facades on net451

bricelam:
Without this, the packages won't install into "classic" Full .NET projects

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/project.json
+++ b/src/Microsoft.Extensions.Logging.Testing/project.json
@@ -20,7 +20,7 @@
   "frameworks": {
     "net451": {
       "frameworkAssemblies": {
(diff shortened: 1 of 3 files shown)


Eilon:
@davidfowl ?

davidfowl:
I think we should make these all as build dependencies so that DNX still works for now

Eilon:
@davidfowl that sounds like a :shipit: to me?

bricelam:
I don't think the two I removed were even used. There's nothing portable in the dependency closure. Concurrent types live in System and mscorlib on Full .NET.

davidfowl:
:shipit:

Eilon:
@bricelam I logged https://github.com/aspnet/DnxTools/issues/30, assigned to you.

PR #380, opened 2016-03-09
Changing BuildScopeImpl to BuildScope<TState>

### LINKED ISSUE #366: Should BeginScopeImpl be a generic method
For all the same reasons that `Log<T>`  is generic https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/ILogger.cs#L36.

/cc @lodejard @rynowak @Eilon

lodejard:
Reduces allocations when none of the registered loggers are tracking scope state. In particular when there are no loggers at all.

Closes #366

### DIFF
--- a/samples/SampleApp/Program.cs
+++ b/samples/SampleApp/Program.cs
@@ -129,7 +129,7 @@ namespace SampleApp
                     _logger.LogWarning("Unexpected warning", ex);
                 }
 
-                using (_logger.BeginScopeImpl("Main"))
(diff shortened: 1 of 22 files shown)


davidfowl:
@Eilon @muratg @lodejard Can we get this change in?

muratg:
@BrennanConroy Can you take this one?

BrennanConroy:
Rebasing then merging

PR #384, opened 2016-03-11
Return IServiceCollection from AddLogging so that additional calls can be chained

ajcvickers:
Previously the AddLogging method did this. It is also common in other builder methods like this. It doesn't do any harm for those that don't want to chain and allows cleaner code with fewer variable declarations for those that do.

More details:
We often write code like this:
```C#\
var appServiceProivder = new ServiceCollection()
    .AddLogging()
    .AddCaching()
    .BuildServiceProvider();
```
This is really clean and easy to read. When we are no longer able to chain we have to write this instead:
```C#
serviceCollection.AddLogging();
serviceCollection.AddCaching();
var appServiceProivder = serviceCollection.BuildServiceProvider();
```
This additional clutter detracts from the overall readability of the code.

Furthermore, the benefits of this kind of pattern are cascading. For example, it lets us use expression body syntax to again create very clear, very readable code:
```C#
protected virtual void ConfigureServices([NotNull] IServiceCollection services)
    => services
        .AddSingleton<CSharpHelper>()
        .AddSingleton<CSharpMigrationOperationGenerator>()
        .AddSingleton<CSharpSnapshotGenerator>()
        .AddSingleton<MigrationsCodeGenerator, CSharpMigrationsGenerator>()
        .AddScaffolding()
        .AddLogging();
```
This was the reason when these methods where designed a couple of years ago they allowed chaining. I'm not sure why we think this is no longer relevant. Also, returning the service collection as opposed to void doesn't force anybody to use chaining if they don't want to, so what is the downside?

### DIFF
--- a/src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
@@ -16,7 +16,8 @@ namespace Microsoft.Extensions.DependencyInjection
         /// Adds logging services to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
-        public static void AddLogging(this IServiceCollection services)
+        /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+        public static IServiceCollection AddLogging(this IServiceCollection services)
         {
             if (services == null)
             {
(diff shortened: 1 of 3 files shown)


davidfowl:
Pretty sure we did this across the board, not just for logging. At least that's what is hope

Eilon:
@javiercn 

BTW we decided to bring back returning the IServiceCollection (or, at least, no one said not to bring it back :smile: ).

ajcvickers:
@Eilon Can I go ahead and merge this?

Eilon:
:shipit:

PR #385, opened 2016-03-11
Target dnx451

bricelam:
Reverts f6eb9182c40a59712f4ab7de4aa9618bc93fe977 for DNX


bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

Issue #378, opened 2016-03-04
Please keep default foreground color, don't override with ConsoleColor.Grey


NinoFloris:
Please keep default foreground color, don't override with ConsoleColor.Grey
`Microsoft.Extensions.Logging.Console.ConsoleLogger:249`
```cs
 return new ConsoleColors(ConsoleColor.Gray, DefaultConsoleColor);
```
This grey really does not do wonders on default OSX terminals (white background)

Eilon:
Yeah we should set the foreground color only when we set the background color - that's needed to ensure readability. But for all other console writes it should be 100% default.

NinoFloris:
Exactly

davidfowl:
Fix for RC2? This is a pretty bad.

Eilon:
Yes RC2.

BrennanConroy:
So Error, Warning, and Information are now going to be the same color as Debug/Trace?

Eilon:
The place where it renders the log level should have explicit foreground and background colors set. But then it should completely unset all colors and use the console's default foreground/background.

BrennanConroy:
Alright, what background color would we like to have for Error/Warning/Information?

Eilon:
For the "labels" use whatever seems normal :smile: is what it's doing now for those acceptable?

BrennanConroy:
They use the default background color but set the foreground color

Eilon:
Chatted with @BrennanConroy , we're on the same page.

BrennanConroy:
bc8ba8aa66773afaab5a4698ecadf4cd5247c2b7

PR #386, opened 2016-03-11
Set background color with foreground color in ConsoleLogger 

BrennanConroy:
#378


Eilon:
:shipit:

davidfowl:
Screenshot on osx? :smile:

Eilon:
@davidfowl it looks like a clown ate a unicorn :smile:

davidfowl:
Is it going to put a black background on my white console with white text?

BrennanConroy:
It's going to use your default colors for everything except the log level text, logger name, and eventid

Sent from my HTC

----- Reply message -----

Eilon:
Yeah the colorization is only for the "header" of each log entry, for which we explicitly set both background and foreground color (to ensure good contrast). Then the log details are in whatever color you had in your console.

Issue #350, opened 2016-01-28
[Discussion] Logging API change


BrennanConroy:
Changes from https://github.com/aspnet/Logging/issues/331 see announcement https://github.com/aspnet/Announcements/issues/148

cwe1ss:
> The Log... functions are now in 3 flavors
>
> `Log....(this ILogger logger, EventId eventId, Exception exception, string message, params object[] args);`
> `Log....(this ILogger logger, EventId eventId, string message, params object[] args);`
> `Log....(this ILogger logger, string message, params object[] args);`

Wouldn't it be good to also have `Log....(this ILogger logger, Exception exception, string message, params object[] args);` to support writing Exceptions without an EventId? As far as I understood, those EventIds are optional, right?

tuespetre:
While I can understand the move from `ILogValues` to `IReadOnlyList<KeyValuePair<string,object>>`, I feel that the overloads should have been left in place. `params object[]` can feel very hokey and I would much prefer to define explicit classes that implement `IReadOnlyList<KeyValuePair<string,object>>` to represent the state of logged events.

muratg:
We are closing this issue because this was just an announcement issue. If you still have any issues or questions, please log a new issue with any additional details that you have.

Issue #294, opened 2015-11-16
No Overload for Exception, error format and args.


runxc1:
There are quite a few overloads but there aren't any that accept the Exception along with an error format string and args.  This seems to be how I most often want to use it.  Is there any reason not to add an overload for that?

muratg:
I think the main reason would be already crowded intellisense for these extension methods.

Would using ILogValues and have an entry for the Exception work?  It doesn't seem ideal but should be doable.

Of course you can add your own extension methods, just follow https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs

cwe1ss:
I've just seen that I posted the same question in the discussion about the recent API changes: https://github.com/aspnet/Logging/issues/350

It appears to be a very useful overload so I think this would be a great addition!

cwe1ss:
I could write a PR for this if you want this in the library. 

I'm sure there will be many more questions about this because many people won't use Event-Ids in their code. I don't agree that this makes the API noisy - i think it's the opposite, it makes it clear that event-ids are optional.

runxc1:
Seems odd to not have a logger that lets you easily write an exception message (using a format string) and pass the Exception in.... so I would use it if its included.   When I added this back in November I thought the code was close to being released so didn't want to add the overload but it seems they are farther from RC now then they were then...

tmds:
@muratg it is a common use-case to log exceptions, why make users add the magic constant ('0') or define their own extension methods?

runxc1:
Here I am a year later and actually using these Extensions.  What sucks is now they make you specify an EventId object anytime you want to include an Exception.

adamhathcock:
Added the overloads myself.  Yay open source.

BrennanConroy:
Thank you @cwe1ss!

BrennanConroy:
https://github.com/aspnet/Logging/commit/c149d48a52403759ef1e1ae932c5e0faf9e340d6

PR #392, opened 2016-03-19
Extension methods for Log*(exception, message, args)

cwe1ss:
This PR adds the following extension method: 

`Log....(this ILogger logger, Exception exception, string message, params object[] args);`

This has been requested in #350 and #294.

I think this would be a great addition for two reasons:
- There already are other extension methods without EventId - having this method makes it more consistent
- This makes EventIds completely optional for people who don't want to use them

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
@@ -56,6 +56,23 @@ namespace Microsoft.Extensions.Logging
         /// Formats and writes a debug log message.
         /// </summary>
         /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
+        /// <param name="exception">The exception to log.</param>
+        /// <param name="message">Format string of the log message.</param>
+        /// <param name="args">An object array that contains zero or more objects to format.</param>
+        public static void LogDebug(this ILogger logger, Exception exception, string message, params object[] args)
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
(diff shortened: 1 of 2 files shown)


cwe1ss:
Any update about this?

muratg:
We won't be changing the API for RC2. Putting this in Backlog for now for considering later.

cwe1ss:
any plans to get this in for RTM?

ThomasArdal:
:+1:

304NotModified:
:+1:

Tratcher:
Bumping for re-triage

sgryphon:
I would suggest keeping the public API narrow, and not proliferate with too many utility methods for all the variations. 

For low level messages, e.g. Verbose, sure, it is reasonable to not have event IDs, but for more significant issues (some Info, but definitely warning/error/etc), we should be encouraging best practice to assign event IDs. (And you can always create your own extension methods if you don't want to follow this practice).

cwe1ss:
The public API already has extension methods without event IDs so this would just fix an inconsistency.

I don't think it's an agreed upon best practice to use event IDs everywhere - there are many very popular loggers who don't use them (e.g. Serilog).

Nowadays almost every common logging system indexes every field anyway so you can achieve the same things with the "message template". And changing message templates are hardly an issue in IMO.

304NotModified:
IMO fixing inconsistency is the strongest point here.

poke:
Is there any update on this? I would really like to see these overloads without the event id.

adamhathcock:
Ended up making my own PR before noticing this one: https://github.com/aspnet/Logging/issues/586

Can we get this merged?  This is only additive to the API and it makes dev life much nicer.

BrennanConroy:
Thank you @cwe1ss!

cwe1ss:
awesome! thank you! 👍

PR #395, opened 2016-03-23
Fix overload of Debug.WriteLine we are using

BrennanConroy:
https://github.com/aspnet/DataProtection/issues/129
@pranavkm @muratg @Eilon

### DIFF
--- a/test/Microsoft.Extensions.Logging.Test/DebugLoggerTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/DebugLoggerTest.cs
@@ -1,6 +1,7 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System;
 using Microsoft.Extensions.Logging.Debug;
 using Xunit;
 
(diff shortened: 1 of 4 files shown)


test/Microsoft.Extensions.Logging.Test/DebugLoggerTest.cs
```diff
@@ -36,5 +37,24 @@ public void CallingBeginScopeOnLogger_ReturnsNonNullableInstance()
...
+            // Arrange
+            var logger = new DebugLogger("Test");
+            var message = "{test string}";
+
+            // Act
+            try
+            {
+                logger.Log(LogLevel.Debug, 0, message, null, (s, e) => s);
+            }
+            catch(Exception e)
```
pranavkm: Why catch it? The test will fail if this method throws.

pranavkm:
:shipit:

test/Microsoft.Extensions.Logging.Test/DebugLoggerTest.cs
BrennanConroy: But then I have nothing to assert :(
```
Assert.True(true);
return;
```

PR #396, opened 2016-03-23
Add missing tests from DNX451 -> NET451 rename

BrennanConroy:
@JunTaoLuo @muratg


JunTaoLuo:
Thanks for catching this. We'll need to scan all the other repos as well. :shipit:

PR #398, opened 2016-03-25
[Fixes #369] Make filtering a first class citizen and remove provider specific filters #369

### LINKED ISSUE #369: Make filtering a first class citizen and remove provider specific filters
There is an ongoing discussion about the noise of log events here: #359 - I created this new issue because it discusses a specific solution - I hope that's ok! I will just provide an overview for people who haven't seen the other conversation.

I strongly believe that this ILogger will be used *a lot* in many libraries and applications in the future. For this reason, being able to filter logs based on their namespace will be a very important topic.
 
Right now, there is no clear recommendation. I read a few times that we should forward everything to another logger (Serilog, log4net, ...) and do the filtering there. However, there are also providers who already provide filtering mechanisms (e.g. ConsoleLogger). 
I'm sure this library will get many more first class providers (Application Insights, DB, ...) in the future, which would remove the need to have a 3rd party logger like Serilog/log4net. However, each provider has to decide about whether and how to implement filtering. This will result in a chaos where each provider must be configured differently or doesn't support filtering at all.

During the discussion in the other thread, @lodejard [created an example FilterLogger](https://github.com/aspnet/Logging/commit/bf2314a9a139ae04fd7310d5df99b0cbe968ae35#diff-459e7bd4d619a1efc2739dff9398cf91R28) which acts as a LoggerFactory-wrapper and can be used like this:

```csharp
/// How to wrap filtering provider around other providers factory
loggerFactory
	.WithFilter(new FilterLoggerSettings
	{
		{ "Microsoft", LogLevel.Warning },
		{ "System", LogLevel.Warning },
		{ "SampleApp.Program", LogLevel.Debug },
	})
	.AddEventLog();
```

I really like this concept and would love to see something like this included as a core concept for v1 because it has many important advantages:
* You could completely remove the provider specific filters from ConsoleLogger etc.
* Writing new providers would be very easy too (which leads to fast adaption)
* Every provider would support this feature automatically
* There would be a consistent API
* There's no cost to it if you don't want to use it.

I'm sure this would remove the need to have another logger (Serilog, ...) for most people! (since providers for most used targets will show up over time)

As an example, here's my real-life scenario:
* Console: We usually don't output anything in Production but we have it running in no-op mode so that we can turn it on in case we need to do some on-server troubshooting (shouldn't happen though :-) ...). In Development, this is the most important provider and we constantly change the levels and namespaces based on our needs.
* We log application startup errors in Production to EventLog in case our persistent store (e.g. Application Insights) isn't available (or didn't initialize correctly). This way we can still have some monitoring for startup crashes. We also have it enabled in Development to make sure it works. Right now we log "Critical" for everything and "Error" for Microsoft.AspNet.Hosting (because [startup errors are not considered critical right now](https://github.com/aspnet/Hosting/issues/612) )
* We use a persistent log store (e.g. Application Insights, Stackify, ...) for our production environment. We log almost everything for our namespace but only "Warning+" for Microsoft. We might have to change the level based on our needs. In Development, we only send a little bit to make sure it works.

kichalla:
@lodejard @muratg @Eilon


NuGetPackageVerifier.json
```diff
@@ -9,6 +9,7 @@
             "Microsoft.Extensions.Logging.Console": { },
             "Microsoft.Extensions.Logging.Debug": { },
             "Microsoft.Extensions.Logging.EventLog": { },
+            "Microsoft.Extensions.Logging.Filter": { },
```
kichalla: I will make sure to update the `csv` file

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -0,0 +1,161 @@
...
+using System.Linq;
+using Microsoft.Extensions.Logging.Test;
+using Xunit;
+
+namespace Microsoft.Extensions.Logging
+{
+    public class LoggerFilterTest
+    {
+        [Fact]
+        public void FiltersMessages_OnDefaultLogLevel_BeforeSendingTo_AllRegisteredLoggerProviders()
```
kichalla: I am working on adding a test where messages can further be filtered at individual logger provider levels too.

kichalla:
@rynowak

src/Microsoft.Extensions.Logging.Filter/project.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": "1.0.0-*",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/aspnet/logging"
+  },
+  "description": "Logging filters.",
```
kichalla: Needs to be updated.

src/Microsoft.Extensions.Logging.Filter/FilterLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// Registers a wrapper logger which provides a common way to filter log messages across all registered
+        ///  <see cref="ILoggerProvider"/>s.
+        /// </summary>
+        /// <param name="loggerFactory">The logger factory.</param>
+        /// <param name="settings">The filter settings which get applied to all registered logger providers.</param>
+        /// <returns>
+        /// A wrapped <see cref="ILoggerFactory"/> which provides common filtering across all registered
+        ///  logger providers.
+        /// </returns>
+        public static ILoggerFactory WithFilter(this ILoggerFactory loggerFactory, FilterLoggerSettings settings)
```
Eilon: Why is this overload needed? Won't the other overload suffice?

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        private Func<LogLevel, bool> GetFilter()
+        {
+            foreach (var prefix in GetKeyPrefixes(_categoryName))
+            {
+                LogLevel level;
+                if (_settings.TryGetSwitch(prefix, out level))
+                {
+                    return l => l >= level;
```
Eilon: `l`?

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLogger.cs
```diff
@@ -0,0 +1,78 @@
...
+                }
+            }
+
+            return l => true;
```
Eilon: Use `_` when the lambda parameter is of no significance. (Or a real parameter name if you like.)

src/Microsoft.Extensions.Logging.Filter/project.json
Eilon: And add `tags` too. I think just `logging` is enough (that's the common one all the logging-related projects have).

Eilon:
@cwe1ss if you don't mind, could you also have a look at this PR and see if it matches your expectations? This started from an issue you logged so I'd love to get your feedback on it :smile:

samples/SampleApp/Program.cs
```diff
@@ -16,8 +19,15 @@ public class Program
...
-            var factory = new LoggerFactory();
+            // A dependency injection based application would get ILoggerFactory injected instead.
+            // Create a logger factory with filter settings that can be applied across all logger providers.
+            var factory = new LoggerFactory()
+                .WithFilter(new FilterLoggerSettings
+                {
+                    { "Microsoft", LogLevel.Warning },
+                    { "System", LogLevel.Warning },
+                    { "SampleApp.Program", LogLevel.Debug }
+                });
```
rynowak: :heart_eyes_cat:

src/Microsoft.Extensions.Logging.Filter/IFilterLoggerSettings.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Primitives;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Filter settings for messages logged by an <see cref="ILogger"/>.
+    /// </summary>
+    public interface IFilterLoggerSettings
```
cwe1ss: Does the `IFilterLoggerSettings` interface really need the `ChangeToken` and `Reload` functionality? They are not used within this library. (`FilterLoggerSettings` only has a dummy implementation) 

Shouldn't this be part of implementations that actually support it - e.g. a `Configuration`-based implementation?

davidfowl:
How does this work in a web application? Who creates the filtered logger factory?

src/Microsoft.Extensions.Logging.Filter/FilterLoggerSettings.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Filter settings for messages logged by an <see cref="ILogger"/>.
+    /// </summary>
+    public class FilterLoggerSettings : IFilterLoggerSettings, IEnumerable<KeyValuePair<string, LogLevel>>
+    {
+        IChangeToken IFilterLoggerSettings.ChangeToken => null;
+
+        public IDictionary<string, LogLevel> Switches { get; set; } = new Dictionary<string, LogLevel>();
```
nblumhardt: With VB.NET using case-insensitive class/namespace names, an insensitive comparer might be called for here, just to improve usability. Maybe a nit, hard to change later however.

cwe1ss:
hmm... David's question made me think and now I'm having doubts about this. I'm using the following code in my project and I'm doing this to have different switches for every provider:
```csharp
var loggerFactory = new LoggerFactory(); // done by WebHostBuilder

loggerFactory.WithFilter(/*...*/).AddConsole();
loggerFactory.WithFilter(/*...*/).AddEventLog();
```
I just realized that this creates multiple `FilterLoggerFactory` instances which are never used and never moved back to the DI framework. So these instances are just a "hack" to make sure the original factory gets the wrapped provider when `AddProvider` is called.

Do you think that this is ok? How about the lifecycle of `FilterLoggerFactory`? I'm not sure if they get garbage collected immediately. Although AFAIK the GC does not call `Dispose` on it, it's still pretty hacky, right? 

If it is required to keep the `FilterLoggerFactory` instance and to add it to the DI, I'm afraid that we would have to use it differently and that we could end up with just one global filter instead of provider-specific filters - which would be a no-go in my opinion.

Wouldn't we have to do something like this in that case:
```csharp
// in Main() to make sure we can create the factory
var filterSettings = new FilterLoggerSettings(/*...*/);
var loggerFactory = new FilterLoggerFactory(filterSettings);

loggerFactory.AddConsole(); // How can we set filters just for this provider?
loggerFactory.AddEventLog();

webHostBuilder
  .UseLoggerFactory(loggerFactory) // it's great that we can do this now :-)
  // other stuff like UseServer(), ...
  .Build();
```

To keep provider-specific filters, `IFilterLoggerSettings` would need information about the provider. We would have to change `TryGetSwitch` to `bool TryGetSwitch(ILoggerProvider, string name, out LogLevel level)` and could do the checks internally. But I'm not sure if this is a good thing?!

The advantage of this would be that filtering is completely encapsulated though. and you wouldn't have to think about it when you add providers.

Your thoughts?

cwe1ss:
I think that the current architecture makes this more complex than it should be. I keep thinking that creating a FilterLoggerFactory which is not actually used is hacky. 

I think this could be simplified if we had an observable-like system. I tried to come up with a proposal for some changes in #400 . It would be great, if you could look through that and share your opinion about whether that would make global filtering easier.

cwe1ss:
I did further changes to my PR in #400 and added filtering as well. I think, integrating filtering into the core would be a cleaner solution than this additional library. 

I'm looking forward to your feedback!

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLoggerFactory.cs
```diff
@@ -1,36 +1,33 @@
...
 
         public ILogger CreateLogger(string categoryName)
         {
-            return _loggerFactory.CreateLogger(categoryName);
+            return _innerLoggerFactory.CreateLogger(categoryName);
         }
 
         public void Dispose()
         {
+            _innerLoggerFactory.Dispose();
```
kichalla: This is my bad. Didn't realize it was left out on purpose.

kichalla:
:up: :date:

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLoggerFactory.cs
```diff
@@ -22,12 +24,14 @@ public void AddProvider(ILoggerProvider provider)
 
         public ILogger CreateLogger(string categoryName)
         {
-            return _innerLoggerFactory.CreateLogger(categoryName);
+            throw new NotImplementedException();
```
kichalla: @lodejard I suppose this is OK as we do not want this logger factory instance to be used for anything other than creating wrapping providers?

cwe1ss: This is one of the main reasons why I think that loggers and providers should be split. I know my other PR is way too large - and probably even too large to just read through it. Sorry for that.

This whole thing just feels really weird to me now. I also know I started this, but I think we should discuss the general approach first before we continue with this.

*Is it ok, that someone can call .WithFilter() and then not use the return object? (see my example above)*

*How does it work if someone does use the returned FilterLoggerFactory and adds it to DI? It will then get resolved whenever someone wants an ILoggerFactory and it will blow up If we throw a NotImplementedException() in CreateLogger*

It just seems that there are too many ways how this could go wrong :-(

304NotModified: it should be `NotSupportedException` then?

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLogger.cs
```diff
@@ -0,0 +1,83 @@
...
+    public class FilterLogger : ILogger
+    {
+        private readonly ILogger _innerLogger;
+        private readonly string _categoryName;
+        private IFilterLoggerSettings _settings;
+        private Func<LogLevel, bool> _filter;
+
+        public FilterLogger(ILogger innerLogger, string categoryName, IFilterLoggerSettings settings)
+        {
+            _innerLogger = innerLogger;
```
javiercn: null check parameters

samples/SampleApp/Program.cs
lodejard: :laughing:

src/Microsoft.Extensions.Logging.Filter/FilterLoggerSettings.cs
lodejard: Well, there's always a "correct" case whether you're case sensitive or not... Plus insensitive comparisons have a long track record of burning us eventually (linux file systems, globalization, etc).

WRT changing later, we've found it's actually easier to go from exact to insensitive in the future than the other way around, because if you go from insensitive to exact folks existing data may no longer match.

src/Microsoft.Extensions.Logging.Filter/IFilterLoggerSettings.cs
lodejard: Being on the interface is probably okay - it's following the same pattern that the console's logger's build-in filtering+reloadablefile featues. Kicker is that the filter needs to know how to react to variable switches without "asking" for the switch literally every time and without being tightly coupled to a specific switch provider.

I guess this could be split into two interfaces, but that seems kind of excessive for fairly buried interface which site owners won't typically implement.

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLogger.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public IDisposable BeginScopeImpl(object state)
+        {
+            return _innerLogger.BeginScopeImpl(state);
+        }
+
+        private Func<LogLevel, bool> GetFilter()
+        {
+            foreach (var prefix in GetKeyPrefixes(_categoryName))
```
lodejard: nit: the `var prefix` is probably more like a `var category`. such a small point of feedback. feel free to ignore. :)

src/Microsoft.Extensions.Logging.Filter/Internal/FilterLoggerFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public void AddProvider(ILoggerProvider provider)
+        {
+            var wrappedProvider = new FilterLoggerProvider(provider, _settings);
+            _innerLoggerFactory.AddProvider(wrappedProvider);
+        }
+
+        public ILogger CreateLogger(string categoryName)
+        {
+            throw new NotImplementedException();
```
lodejard: oh, goodness... this is very unfortunate... are we certain this shouldn't just call through? seems like it would be fairly frustrating to get a NotImplemented from CreateLogger, and learn that you need to "go up" to the original logger yourself...

kichalla:
Got a verbal signoff from @lodejard

PR #399, opened 2016-03-26
PR for #398 (filtering PR): Removed dependency on Configuration; cleanup

cwe1ss:
I hope it's possible to do a PR for a PR?!? :)

This would remove the dependency on `Microsoft.Extensions.Configuration`. IMO the filter library, as it is implemented now, does not need to know anything about the `Configuration` system. This would only be required in a Configuration based implementation of `IFilterLoggerSettings`.

It also removes an unnecessary method and adds `readonly` to member variables.


cwe1ss:
no longer relevant since filtering has already been integrated.

PR #400, opened 2016-03-27
[Prototype] Split loggers and log providers/sinks (ILogger has too many responsibilities)

cwe1ss:
I think it is problematic that the `ILogger` interface is used for creating log entries **and** for writing them to some output. For that reason, I suggest that providers (e.g. Console, EventLog, ...) should not inherit `ILogger`. Instead, we should move towards an Observable-like system.

I will try my best to explain the benefits and possible future improvements, so please take a seat for the long story! :)

This is the current dependency structure:
* `LoggerFactory` creates and manages `Logger` instances - one per categoryName. They are responsible for **creating** log entries.
* `LoggerFactory` has one `ILoggerProvider` per provider
* Every `Logger` instance stores a list of `ILogger` instances - one per provider. They are responsible for **writing** log entries to outputs.
* Whenever a new provider is added, `LoggerFactory` adds the `ILogger` instance from the provider to every `Logger`-instance.
* Each `Logger` communicates directly with each provider-`ILogger`.

I think that this solution has a few disadvantages:
* Providers are forced to have `ILogger` instances per logger category. Requiring different `Logger` instances is a concept of log entry creation and is useful there. Why should there be many instances of e.g. an EventLog-provider?
* This also makes it harder for providers to batch log entries. If you want to send e.g. 50 log entries at once to some output system, you would either have to batch 50 per category or you would have to merge them somehow first.
* the direct relationship between `Logger` and the provider-`ILogger`s makes it harder to apply global patterns like filtering because you always have to broadcast everything to every `Logger` in the LoggerFactory. 
* Providers are forced to implement the same contract. If we would e.g. have global filtering, `IsEnabled` would only be interesting for the creation-side. Providers would not need the same method because they would not even receive the entry.
* Mixing the concepts makes this really hard to understand. Just look at the mix of `ILogger` and `Logger` here.

In my opinion, the logging system should work more like the observable pattern. `ILogger` instances should be responsible for creating log entries and they should send them to some central coordinator. This coordinator knows about subscribed providers and forwards the entries to each of them. 

#### what I would like to see
Ideally, I would completely split log creation and forwarding. This would begin with how providers are set up. I think it is weird, that you can add providers to a `LoggerFactory` because the factory is used in many locations. Why should you be able to do something like this:
```csharp
public SomeClass {
  public SomeClass(ILoggerFactory loggerFactory) {
    loggerFactory.AddConsole();
  }
}
```
I think it should only be possible to call `CreateLogger` here. I think this would come closest to what the name "factory" implies.

Ideally, providers would be observers - e.g. `ILogObserver`. There would be a corresponding `LogObservable` which acts as the described central coordinator. Only this class would need the `AddObserver` (aka `Subscribe`, aka `AddProvider`) method. So when you would create the application, you would configure `LogObservable` instead of `LoggerFactory`. This would look like this:
```csharp
var logObservable = new LogObservable()
  .AddEventLog()
  .AddConsole();

var loggerFactory = new LoggerFactory(logObservable);
// both will be added to DI as singleton

// There would be no AddProvider here
var logger = loggerFactory.CreateLogger<Program>();
```
Ideally, the `LogObservable` and `LogObserver` would actually implement `IObservable` and `IObserver` to allow us to use Rx components. However I'm not sure if this would play well with other required methods like `BeginScope`.

#### baby steps
Since this would be a quite big change, I'm proposing a compromise and a first step with this pull request:
* the meaning of `ILoggerProvider` will be changed (instead of renamig it to LogObserver). This interface will represent everything that is required for providers to **write** entries to some output. It will no longer have `CreateLogger`, instead it will have methods like `Log` and `BeginScope`. This will be a breaking change, however it only affects people who built their own providers.
* Providers will no longer have to inherit `ILogger` because everything they need is on `ILoggerProvider`.
* Providers will get the `categoryName` as part of their methods. This means, there may be just one instance of every provider. If the provider needs to have different instances, it can still do so based on this parameter.
* The `LoggerFactory` will still be responsible for adding providers to make sure there is no breaking change for client applications.
* The `LoggerFactory` and `Logger` instances will no longer maintain a direct relationship with providers. Instead, they will just forward everything to the new `LogObservable` class. This results in very easy to understand code in `Logger` and `LoggerFactory`
* The new `LogObservable` contains most of the code from `Logger` and `LoggerFactory`. It allows us to have one central location that talks to the providers.

This code contains a POC of how this would work. Please note the following:
* the existing providers still contain *LoggerProvider and *Logger classes. The *Logger class would not be necessary but I kept them to keep changes to a minimum.
* the providers still contain code for filtering. This could be moved to the LogObservable class.
* some unit tests fail - I just made sure everything compiles for now.

so don't merge it yet :-) :-)

I think that this solution would also make global filtering #398 easier. I think it would remove the need to create a separate LoggerFactory. I will try to come up with a good solution.

I'm looking forward to your feedback!


davidfowl:
I'm like 97% sure we won't take an architectural change this big, this late in the game. This not only affects us but the serilog and nlog implementations (which are already filterable and fully featured). At this point, it just seems like a non-starter IMO.

304NotModified:
Not sure if it's too late. It's a fact that rc2 will have breaking changes.

cwe1ss:
thanks for taking the time to read through this, David! I'm sorry for taking up so much of your time!

I understand that. Maybe this PR is just helpful to raise awareness for how this library is scoped and what it can/can't/shouldn't do. I wouldn't be offended if this PR doesn't go my way.

Most of my initial post tries to explain an ideal scenario which I'm sure shouldn't be fully implemented now. However, I tried to come up with a minimal-impact (but yes, still critical) solution. This would be the "management summary":

*A new central observable-like component will send log entries to all providers. Providers will inherit only `ILoggerProvider`, an observer-like modified version of `ILogger`. This simplifies provider implementations (they only need one instance) and allows for more flexibility.*

I don't think that breaking existing providers is a big issue because there already are other breaking changes for RC2 and this change wouldn't be very difficult. I'd be happy to send PRs.

Serilog and NLog are great products and I enjoy using them. However, people will try to have as few dependencies as possible, so you can't prevent people from making this library their first class logger solution. They will eventually build direct providers for most outputs. For that reason, I don't think that the architecture of this library should lack important extensibility points.

@nblumhardt, @304NotModified What's your opinion about this?

304NotModified:
Too bad it's also unlogic that consolelogger won't implement ilogger. The name of consolelogger is now also not correct. 

FYI they called for that reason "targets"  in NLog, so we have ilogger and consoletarget in NLog.

cwe1ss:
I agree, I kept that the same for now just to keep the changes minimal. They should not be called *Logger because they are not the same as instances that create log entries. (even with the current system IMO)

As I've said in my initial post, I also think that the name LoggerProvider is misleading. Something like *Sink, *Target, *Observer, *... would be better IMO.

cwe1ss:
I made further enhancements and I also added a POC for filtering. The changes to the core classes are now actually pretty small, however it's still a huge-impact PR. [it looks a lot less dramatic when you look at it locally instead of here on GitHub :-) ]

Here's the summary of changes in this PR. Stuff from my initial post is included here, so this summary is complete.

**Log creation and log writing has been separated - introducing `ILogSink`**
`ILogger` is now only used for creating log entries. The complete contract for providers is in the current `ILoggerProvider` interface. However, since this name no longer makes sense (it doesn't "provide loggers"), I've renamed it to `ILogSink`. Of course, this can change if anyone has better ideas.

**The Abstractions-package no longer has any knowledge about providers/sinks**
`ILogSink` was moved to `Microsoft.Extensions.Logging`. This makes it even clearer, that `M.E.L.Abstractions` is used for creating log entries and `M.E.L` for processing log entries.

IMO this also makes clear that `M.E.L` should be seen as a competitor to Serilog, NLog, etc and not as a supplier because it contains routing logic and has its own providers/sinks.

@nblumhardt @304NotModified With this change, it would now easily be possible for Serilog, NLog, etc to completely circumvent `Microsoft.Extensions.Logging`. Instead of implementing a provider/sink, you could just create your own `ILoggerFactory` and `ILogger` and have people inject it through `IWebHostBuilder.UseLoggerFactory`. This way, people who are using your frameworks would no longer have to use `Microsoft.Extensions.Logging` (although other libraries like `Hosting` will still bring in this DLL)! 

I think this alone is a huge win. It's really unnecessary to send all logs through `M.E.L` if all of them should end up in Serilog, NLog, etc anyway. Reducing dependencies FTW! :)

**Breaking changes for sinks/providers**
* Sinks must reference `Microsoft.Extensions.Logging` instead of `Microsoft.Extensions.Logging.Abstractions` (`ILogSink` is there)
* Sinks must only implement `ILogSink` instead of `ILoggerProvider` and `ILogger`.
* Sinks no longer have to maintain loggers per category. The `ILogSink` contains the `categoryName` in each method.
* the Add*-methods (e.g. AddEventLog) have to target `LoggerFactory` instead of `ILoggerFactory` because AddProvider/AddSink no longer exists on the `ILoggerFactory` interface.

This is the contract of `ILogSink`:
```csharp
public interface ILogSink: IDisposable {
  void Log<TState>(categoryName, logLevel, eventId, state, exception, formatter);
  bool IsEnabled(categoryName, logLevel);
  IDisposable BeginScope(categoryName, state);
}
```
**ILogSinkProvider / LogSinkProvider**
To make sure, not each category `Logger` in `M.E.L` has to maintain a list of sinks, there is a component called `ILogSinkProvider` which maintains sink-specific properties. It maintains *a list of sinks* and *a filter object* (see below). This component takes over the provider-specific logic of `LoggerFactory`. However, `LoggerFactory` still has `AddSink` for backwards-compatibility. (people can now still use e.g. AddEventLog on LoggerFactory if they want to)

**M.E.L sink implementations**
Console/Debug/EventLog/TraceSource have been renamed and changed to only implement ILogSink.

**Filtering POC**
Filtering is now built into the library. `ILogSinkProvider` can maintain a `ILogFilter` object which has a method `bool IsEnabled(ILogSink sink, string categoryName, LogLevel logLevel)`.

`M.E.L/Filtering` contains an implementation of `ILogFilter` which contains many parts of #398. It's a zero-overhead solution if you don't want to use it (at least I think so). 
There's not yet a Configuration-based implementation, but this could easily be added. The sinks also still have their own implementation.
.............
This code is still pretty much untested and some tests fail. I will continue to work on this, if you decide that we should proceed with it.

davidfowl:
> Not sure if it's too late. It's a fact that rc2 will have breaking changes.

It is too late. We're not going to take a change this big to the design of logging at this point. RC2 does have breaking changes but that's no excuse to make more. 

On top of that, I'm not even sure I like the proposed changes. Especially the fact that custom sinks now have to use Logging instead of abstractions. I also don't understand how the application wires up the sink since it doesn't control creating the logger factory.

cwe1ss:
I added [Prototype] to the title to make sure it's clear that I don't expect this to be merged like this. I think it's just easier to discuss stuff if there's some actual code. It's also important that the main reason why this PR is so big is that it shows multiple things. I probably shouldn't have done that.

> RC2 does have breaking changes but that's no excuse to make more.
In case of providers/sinks I think it actually is. There are probably less than 20 provider implementations in the world. Having the ILogSink interface would be a win for each of them. Maybe we get some feedback from those developers during the next days. 

@davidfowl what to you think about the core idea of `ILogSink`? how about not renaming providers to sinks and just changing `ILoggerProvider` to contain the methods from my `ILogSink` interface? The naming wouldn't be great but it's still acceptable IMO. As you can see in the Debug and EventLog provider, this hugely simplifies provider creation.
`Logger` would then just keep a list of `ILoggerProvider` instead of `ILogger` if you don't want to introduce the proposed `ILogSinkProvider` container. This would be a really small change and wouldn't have any breaking changes for application developers.

cwe1ss:
Regarding moving ILoggerProvider away from the Abstractions package:

People here have said many times that we should use Serilog, NLog, etc because they have more features. So why should we be required to use `Microsoft.Extensions.Logging` at all? It seems highly unlikely that anyone will want to combine providers from M.E.L with sinks from Serilog because Serilog completely covers the M.E.L providers.

IMO, "providers" should only be used for actual outputs like console, files, sql, etc. Logging frameworks like Serilog, NLog, etc should be seen at the same layer as `Microsoft.Extensions.Logging` and should therefore be implemented as a custom `ILoggerFactory`. This would completely remove one layer.

But this is where `ILoggerProvider` and `ILoggerFactory.AddProvider` start to get problematic. These are concepts of `Microsoft.Extensions.Logging`'s implementation. Serilog, NLog, etc don't need them. That's why I wanted to move it away!

However, I noticed that my idea of putting AddEventLog, AddConsole on `LoggerFactory` doesn't work. It must be an extension method to ILoggerFactory, because that's what's injected into `Startup`. So we would need a different solution. Maybe `ILoggerFactory.AddProvider` can be an extension method in `Microsoft.Extensions.Logging`.

> I also don't understand how the application wires up the sink since it doesn't control creating the logger factory.

Now that we have `IWebHostBuilder.UseLoggerFactory`, we can actually create the logger factory. People just have to move their logging setup into the `Main` method. You don't have DI there but logging should be configured before DI anyway IMO.

cwe1ss:
> On top of that, I'm not even sure I like the proposed changes. Especially the fact that custom sinks now have to use Logging instead of abstractions. 

We could also introduce another package called `Microsoft.Extensions.Logging.ProviderAbstractions` which contains the `ILoggerProvider`-interface. This just seemed like an overkill to me. 

The important thing for me was, that the interface shouldn't be in the regular Abstractions package. See previous post for the reasons.

cwe1ss:
I'm closing this as it contains too many things. Also, filtering has now been merged into dev. I hope I didn't annoy you with this PR... I'm pretty new to this open source thing. 

Maybe I will create a separate and smaller PR that shows just the changes to ILoggerProvider. Would still be a breaking change though.

PR #403, opened 2016-04-09
Cross-compile to netcore50 to support EF on UWP

natemcmaster:
Part of https://github.com/aspnet/EntityFramework/pull/5021


src/Microsoft.Extensions.Logging.Abstractions/project.json
```diff
@@ -28,6 +28,22 @@
         "System.Runtime.Extensions": "4.1.0-*",
         "System.Runtime.InteropServices": "4.1.0-*"
       }
+    },
+    "netcore50": {
+      "dependencies": {
+        "Microsoft.NETCore.Platforms": {
```
davidfowl: You shouldn't need this either

natemcmaster: Yup. Until we switch off `--infer-runtimes`, restore fails unless this Ms.NetCore.Platforms is listed as a build dependency.

PR #404, opened 2016-04-11
Cache the console log scope path if its already calculated

kichalla:
@rynowak


rynowak:
What's the purpose of this change?

kichalla:
@rynowak The current implementation is inefficient as we repeatedly calculate it for every log message written. BTW we already have tests covering the scope path scenarios so this change is safe.

rynowak:
Why is this closed?

Get some perf data, ping me if you need help measuring

304NotModified:
I don't like micro optimalisations, but doing expensive stuff for every log message sounds wrong. So I think this is a good PR.

PR #409, opened 2016-04-18
Only reset color if changed

### LINKED ISSUE #408: System.Console Color stuff is crazy expensive
Is calls System.Console.GetBufferInfo which is 1/3 as expensive as actually writing the messages (which is also relatively expensive)

benaadams:
Resolves #408

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/Internal/WindowsLogConsole.cs
+++ b/src/Microsoft.Extensions.Logging.Console/Internal/WindowsLogConsole.cs
@@ -7,19 +7,24 @@ namespace Microsoft.Extensions.Logging.Console.Internal
 {
     public class WindowsLogConsole : IConsole
     {
+        private ConsoleColor _background = System.Console.BackgroundColor;
+        private ConsoleColor _foreground = System.Console.ForegroundColor;
+
         public void Write(string message, ConsoleColor? background, ConsoleColor? foreground)
         {
             var changedColor = false;
-            if (background.HasValue)
+            if (background.HasValue && _background != background)
             {
                 System.Console.BackgroundColor = background.Value;
                 changedColor = true;
+                _background = background.Value;
             }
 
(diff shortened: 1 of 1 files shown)


benaadams:
This may be better #7823

stephentoub:
> This may be better

Not necessarily.  It'll help, but we can't do color caching in the Console class because at that level we shouldn't be making assumptions about what else in the process might be interacting with the OS and changing colors.  As a result, we're limited in the optimizations we can make.  A higher-level component like this one could choose to do such caching.  And as you've done here, you also know whether the calls to change the color, write, and reset the color are correlated, something we can't tell reasonably in the Console implementation itself.

stephentoub:
The first commit looks like pure goodness to me.

The second commit looks like it could be problematic, e.g. if something else in the process uses Console.*Color to change colors.

I'd also have expected the whole body here to be surrounded by a lock on Console.Out, so that anyone else in the process writing would have to synchronize with the color changes being done here.

benaadams:
> I'd also have expected the whole body here to be surrounded by a lock on Console.Out

The locking is handled up a level https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L122

davidfowl:
@benaadams do we still need this after the changes to corefx?

muratg:
@benaadams ping :)

benaadams:
As aside was wondering if... check if console is redirected if so drop colour and coalesce writes. Except; generally piping to file makes console instant compared to not 😄

BrennanConroy:
Considered that, no way of detecting redirects without pinvoke though :(
https://github.com/dotnet/corefx/issues/10428

stephentoub:
I'm confused.  What would you be checking with a P/Invoke that Console.IsOutputRedirected doesn't tell you?

PR #424, opened 2016-05-10
ILogger implementation that sends data to EventSource/EventListener

karolz-ms:
This implementation follows the standard "use extension method to enable" pattern. Nothing is on by default.

A separate test assembly was created because there are two slightly different implementations, one for .NET Core and one for .NET 4.6


src/Microsoft.Extensions.Logging.EventSource/project.json
```diff
@@ -0,0 +1,49 @@
...
+    "Newtonsoft.Json": "8.0.3"
+  },
+  "frameworks": {
+    "net46": {
+      "dependencies": {
+        "System.Runtime": "4.0.0"
+      },
+      "imports": [ "dotnet" ]
+    },
+    "netcoreapp1.0": {
```
Eilon: This should be a `netstandard` TFM, which is for library. `netcoreapp` is only for apps. E.g. see this other logger: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Debug/project.json#L25-L36.

karolz-ms: The reason why I went with netcoreapp is that when I am trying to consume Newtonsoft JSON from a project that targets netstandard, I am getting the following error:

```
Errors in C:\code\aspnet-logging-rc\src\Microsoft.Extensions.Logging.EventSource\project.json
    Package Newtonsoft.Json 8.0.3 is not compatible with netstandard1.3 (.NETStandard,Version=v1.3). Package Newtonsoft.Json 8.0.3 supports:
      - net20 (.NETFramework,Version=v2.0)
      - net35 (.NETFramework,Version=v3.5)
      - net40 (.NETFramework,Version=v4.0)
      - net45 (.NETFramework,Version=v4.5)
      - portable-dnxcore50+net45+win8+wp8+wpa81 (.NETPortable,Version=v0.0,Profile=net45+wp80+win8+wpa81+dnxcore50)
      - portable-net40+sl5+win8+wp8+wpa81 (.NETPortable,Version=v0.0,Profile=Profile328)
    One or more packages are incompatible with .NETStandard,Version=v1.3.
```

Now, I use Newtonsoft to serialize log data into JSON format. The serialization is quite simple, so I could do it manually and remove the dependency if overcoming the problem above is hard, but I would prefer to leverage Newtonsoft if possible

src/Microsoft.Extensions.Logging.EventSource/project.json
```diff
@@ -0,0 +1,49 @@
...
+      "xmlDoc": true
+  },
+  "dependencies": {
+    "Microsoft.Extensions.Logging.Abstractions": "1.0.0-*",
+    "System.Diagnostics.Tracing": "4.1.0-*",
+    "System.Threading": "4.0.11-*",
+    "Newtonsoft.Json": "8.0.3"
+  },
+  "frameworks": {
+    "net46": {
```
Eilon: This is why it's failing... this presumably doesn't exist when compiling on Mac (which is what Travis uses).

@davidfowl suggestions on how to get this working?

@karolz-ms for right now feel free to ignore this, but we'll have to come up with some solution for this.

src/Microsoft.Extensions.Logging.EventSource/project.json
Eilon: You still can't use `netcoreapp`, it's the wrong TFM. Use `imports` instead: https://github.com/aspnet/Mvc/blob/3d494fd7326d3e137fdb22b53f01c5a34866fbaa/src/Microsoft.AspNetCore.Mvc.ViewFeatures/project.json#L59-L61

(Mind you, this is all temporary; once a new build of Json.NET is out that supports `netstandard`, this will go away.)

karolz-ms:
Yes, that has helped, thanks--I can compile and run tests fine with the new logger targeting netstandard.

net46 issue still remains...

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,218 @@
...
+            return logLevel >= Level;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+            if (!IsEnabled(logLevel))
+                return;
+
+            // See if they want the formatted message
+            if (LoggingEventSource.Instance.IsEnabled(System.Diagnostics.Tracing.EventLevel.Critical, LoggingEventSource.Keywords.FormattedMessage))
```
davidfowl: nit: Why are we using the full namespace for the EventLevel?

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,218 @@
...
+                    eventId.ToString(),
+                    exceptionJson,
+                    ToJson(arguments));
+            }
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            if (!IsEnabled(LogLevel.Critical))
+                return NoopDisposable.Instance;
```
davidfowl: we don't use self closing ifs, add the braces

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Not sure what was original reason, but we do not need it. Fixed.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Sorry. Fixed

karolz-ms:
So @davidfowl : any ideas/recommendations on how to treat the .NET46 target?

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,226 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.Tracing;
+using System.Threading;
+using Newtonsoft.Json;
+using System.IO;
```
Eilon: Sort & Remove.

Eilon: (Apply to all files, as necessary. Also, ensure "System" namespaces are at the top - VS has a setting for this.)

src/Microsoft.Extensions.Logging.EventSource/ExceptionInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Diagnostics.Tracing;
+
+namespace Microsoft.Extensions.Logging.EventSourceLogger
+{
+    /// <summary>
+    /// Represents information about exceptions that is captured by EventSourceLogger
+    /// </summary>
+#if !NO_EVENTSOURCE_COMPLEX_TYPE_SUPPORT
+    [EventData(Name ="ExceptionInfo")]
+    #endif
```
Eilon: Style: Would be nice to balance the #if/#endif in terms of indentation.

src/Microsoft.Extensions.Logging.EventSource/project.json
```diff
@@ -0,0 +1,40 @@
+{
+  "version": "1.0.0-*",
+  "description": "A logger provider that sends data logged through ILogger interfaces to EventSource/EventListener infrastructure",
```
Eilon: Change to:
`EventSource/EventListener logger provider implementation for Microsoft.Extensions.Logging.`

(this matches the pattern we use in the descriptions of the other loggers)

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
```diff
@@ -0,0 +1,581 @@
...
+        {
+            // This is a workaround for a bug in EventListener where OnEventSourceCreated will be called before the listener
+            // is fully constructed. Unfortunately this bug has shipped and is now an "expected behavior".
+            public class ListenerSettings
+            {
+                public EventKeywords Keywords;
+                public EventLevel Level;
+                public string FilterSpec;
+            }
+            public static ListenerSettings Settings = new ListenerSettings();
```
Eilon: This code is not thread-safe. Multiple tests will execute in parallel and contend for this value. One test could affect another test.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,236 @@
...
+    ///     protected override void OnEventWritten(EventWrittenEventArgs eventData) {
+    ///         // Look for the formatted message event, which has the following argument layout (as defined in the LoggingEventSource.  
+    ///         // FormattedMessage(LogLevel Level, int FactoryID, string LoggerName, string EventId, string FormattedMessage);
+    ///         if (eventData.EventName == "FormattedMessage")
+    ///             Console.WriteLine("Logger {0}: {1}", eventData.Payload[2], eventData.Payload[4]);
+    ///     }
+    /// }
+    /// </summary>
+    [EventSource(Name = "Microsoft-Extensions-Logging")]
+    public class LoggingEventSource : System.Diagnostics.Tracing.EventSource
```
Eilon: Is the `System.Diagnostics.Tracing.` prefix needed? The namespace is already imported.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,236 @@
...
+            /// </summary>
+            public const EventKeywords FormattedMessage = (EventKeywords)4;
+            /// <summary>
+            /// Turns on the 'MessageJson' event when ILogger.Log() is called.   It gives  JSON representation of the Arguments. 
+            /// </summary>
+            public const EventKeywords JsonMessage = (EventKeywords)8;
+        }
+
+        /// <summary>
+        ///  The one and only instance of the LoggingEventSource.
```
Eilon: Why would there be only one? The typical pattern is to have everything be instance-based.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        /// <summary>
+        ///  The one and only instance of the LoggingEventSource.  
+        /// </summary>
+        internal static readonly LoggingEventSource Instance = new LoggingEventSource();
+
+        internal static readonly LogLevel LoggingDisabled = LogLevel.None + 1;
+
+        private string _filterSpec = null;
```
Eilon: No need to initialize default values of fields - they'll already be `null` anyway.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,34 @@
...
+            if (factory == null)
+            {
+                throw new ArgumentNullException(nameof(factory));
+            }
+
+            LoggingEventSource.Instance.CreateLoggerProvider(factory);
+
+            // Do not call factory.AddProvider() here. EventSourceLoggerProvider will do that automatically 
+            // when something starts using Microsoft-Extensions-Logging EventSource. This way there is no overhead
+            // from using the EventSourceLogger when no one is interested in the data it exposes.
```
Eilon: This concerns me. Why not follow the regular ILoggerFactory/ILoggerProvider pattern? I think it's fine if the logger *internally* has logic to detect that it has no listeners and no-op all its logic on a *per-instance* basis. But the logger provider must always already be on the ILoggerFactory.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        public readonly EventSourceLoggerProvider Next;
+
+        /// <summary>
+        /// <inheritdoc />
+        /// </summary>
+        public ILogger CreateLogger(string categoryName)
+        {
+            var newLogger = _loggers = new EventSourceLogger(categoryName, _factoryID, _loggers);
```
Eilon: This code needs to be thread-safe. Multiple requests can be occurring simultaneously and the head of the linked list will be corrupted. in case of a race.

Eilon: Though, ultimately, the real question is: what is special about this logger? I don't think any other loggers use this type of pattern.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -0,0 +1,238 @@
...
+            {
+                // Compute an ID for the Factory.  It is its position in the list (starting at 1, we reserve 0 to mean unstarted). 
+                _factoryID = 1;
+                for (var cur = Next; cur != null; cur = cur.Next)
+                {
+                    _factoryID++;
+                }
+
+                // Add myself to the factory.  Now my CreateLogger methods will be called.  
+                _loggerFactory.AddProvider(this);
```
Eilon: We should try to understand why this logger adds itself dynamically like this.

Eilon:
Added some comments 🕐

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,236 @@
...
+    ///     }
+    ///     protected override void OnEventWritten(EventWrittenEventArgs eventData) {
+    ///         // Look for the formatted message event, which has the following argument layout (as defined in the LoggingEventSource.  
+    ///         // FormattedMessage(LogLevel Level, int FactoryID, string LoggerName, string EventId, string FormattedMessage);
+    ///         if (eventData.EventName == "FormattedMessage")
+    ///             Console.WriteLine("Logger {0}: {1}", eventData.Payload[2], eventData.Payload[4]);
+    ///     }
+    /// }
+    /// </summary>
+    [EventSource(Name = "Microsoft-Extensions-Logging")]
```
davidfowl: How does this name show up?

src/Microsoft.Extensions.Logging.EventSource/project.json
davidfowl: We might have to skip this on mono with custom build logic

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
```diff
@@ -0,0 +1,581 @@
...
+                    "E7JS",
+                    "OuterScopeJsonStop",
+                    "E8JS");
+            }
+        }
+
+        [Fact]
+        public void Logs_OnlyMessage_IfKeywordSet()
+        {
+            TestEventListener.Settings.Keywords = LoggingEventSource.Keywords.Message;
```
davidfowl: Why are these static?

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
karolz-ms: Tests in this class cannot run in parallel, because there is only one instance of Microsoft-Extensions-Logging per process (or, more precisely, per app domain). If you attempted to run multiple tests in parallel, they would interfere with each other (every listener would receive all the events raised from all the concurrently running tests) and the expected results would not match actual results.

My understanding is that running tests in parallel requires an explicit opt-in in xunit. If that is not the case and if I should explicitly opt out, please advise how to do so.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: That is how EventSource works; there can be only one EventSource instance with a given name per app domain.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
karolz-ms: Adding @vancem

This follows the ETW pattern of ensuring absolutely minimum overhead in the case when no one is interested in the data exposed by Microsoft-Extensions-Logging EventSource.

Please explain why "the logger provider _must_ always already be on the ILoggerFactory". I do not see any contract on the ILoggerFactory that lets the consumer enumerate the ILoggerProviders, so what the extension method does is implementation detail as long as the expected result of event data appearing through Microsoft-Extensions-Logging EventSource is observed.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
karolz-ms: This code is thread safe because the only caller (LoggingEventSource) acquires a lock on the only instance before calling into EventSourceLoggerProvider.

The linked list is there because we need to respond to dynamic configuration change commands sent through EventSource/ETW infrastructure (it all starts at LoggingEventSource.OnEventCommand, which is also thread-safe BTW).

Bottom line, we are bridging ILogger world with EventSource/ETW world. The contract for this bridge should follow ILogger patterns. This is why Vance and I followed your request of making the bridge "off by default". However _the implementation and behavior_ of the bridge follows EventSource/ETW patterns. It is not realistic to expect that this logger implementation will look like any other logger, just pushing data into different place. I think this gives users the best of both worlds, as long as contracts & expectactions from both worlds are satisified

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
karolz-ms: Please see my reply under EventSourceLoggerFactoryExtensions

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: This name represents a well-known ID of the EventSource/ETW provider that exposes ILogger data to the ETW world. It will be used to subscribe to ILogger data by consumers who use EventSource/ETW to capture diagnostic traces.

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
karolz-ms: This is explained in the comment inside TestEventListener. It is a workaround. Basically, there is a bug in EventListener implementation, which causes it to call into one of the overloads (OnEventSourceCreated) before the instance is fully constructed. So we cannot use TestEventListener constructor parameters to initialize the instance; the initialization parameters must be supplied externally.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Fixed

src/Microsoft.Extensions.Logging.EventSource/ExceptionInfo.cs
karolz-ms: Fixed

src/Microsoft.Extensions.Logging.EventSource/project.json
karolz-ms: Fixed

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: Removed

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: Fixed

karolz-ms:
@Eilon , I think I have addressed all the code formatting issues.

As to why things are implemented the way they are, I tried to explain. Let me know if you need more info or want to discuss in person; please also add @VanceM to the conversation

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
davidfowl: That's fine but don't let it leak all over the code. Pass in the singleton instance in the ctors of the classes that consume it.

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
davidfowl: We should fix the code so that we can run in parallel. The fact that the event source is a singleton shouldn't leak throughout the code base.

karolz-ms: Let me make sure that we are on the same page.

The fact that an EventSource with a given name is a singleton is not a developer's choice, it is part of EventSource contract. For example

[EventSource(Name="MyCompany-ApplicationX-MyEventSource")]
public sealed class MyEventSource: EventSource { ... }

var es1 = new MyEventSource();
var es2 = new MyEventSource(); // This line will _always_ throw an exception--there can be only one MyEventSource in the process

I guess if we really need to make the tests run in parallel, we could do that by adding an extra tag to every trace indicating to which test the given trace belongs. Then at EventListener level we could filter out all traces that do not belong to given test. The problem with this approach is that it makes tests more likely to miss a case when the bridge passes too many thing through. I'd rather keep things as straightforward as possible. 

I have not measured it, but it seems that the speed of execution of this test suite is below discernable threshold. How much do we care about saving, say, 300 ms?

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: OK. Fixed

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,230 @@
...
+    internal class EventSourceLogger : ILogger
+    {
+        private LoggingEventSource _eventSource;
+        private int _factoryID;
+        private static int s_activityIds;
+
+        public EventSourceLogger(string categoryName, int factoryID, LoggingEventSource eventSource, EventSourceLogger next)
+        {
+            CategoryName = categoryName;
+            Level = LoggingEventSource.LoggingDisabled;     // Default is to turn off logging
```
davidfowl: nit: Can you put the comment on top instead of on the side?

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// ActivityScope is just a IDisposable that knows how to send the ActivityStop event when it is 
+        /// desposed.  It is part of the BeginScope() support.  
+        /// </summary>
+        private class ActivityScope : IDisposable
+        {
+            private string _categoryName;
+            private int _activityID;
+            private int _factoryID;
+            private bool _isJsonStop;
+            private LoggingEventSource _eventSource;
```
davidfowl: `readonly`

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <summary>
+        /// Converts an ILogger state object into a set of key-value pairs (That can be send to a EventSource) 
+        /// </summary>
+        private IEnumerable<KeyValuePair<string, string>> GetProperties(object state)
+        {
+            var arguments = new List<KeyValuePair<string, string>>();
+            var asKeyValues = state as IEnumerable<KeyValuePair<string, object>>;
+            if (asKeyValues != null)
+            {
+                foreach (var keyValue in asKeyValues)
```
davidfowl: nit: always use braces.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,230 @@
...
+                    arguments.Add(new KeyValuePair<string, string>(keyValue.Key, keyValue.Value.ToString()));
```
davidfowl: Can value be null?

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -0,0 +1,230 @@
...
+            }
+            return arguments;
+        }
+
+        private string ToJson(IEnumerable<KeyValuePair<string, string>> keyValues)
+        {
+            StringWriter sw = new StringWriter();
```
davidfowl: nit: use more var

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
karolz-ms: I have measured--the execution of these 11 tests takes less than 300 ms, and that includes Console.WriteLine calls to report the timing. I suggest we leave them as they are.

karolz-ms: Plus of course tests in this class _can_ be executed in parallel with tests from other test classes

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Will do

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: OK

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Thx, missed this spot in the previous round of fixes

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: Will check for that, thx

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
karolz-ms: OK

test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
karolz-ms: David and I discussed in person. We will keep the tests as they are, in a separate assembly, with disabled parallelism and a comment explaining why this is necessary

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
davidfowl: I agree with @Eilon here. We shouldn't be dynamically adding a provider to the factory when a message comes up. Just no-op until it's turned on.

karolz-ms: All right. I made that change

vancem: LoggingEventSource is perhaps different from most loggers in that the creation of a LoggingEvnetSource indicates the POTENTIAL DESIRE for logging (you want to allow users to get at your events via EventListeners / ETW).   It is only later, when someone actually turns on logging using EventListener/ETW that you actually log anything.   It would be a shame to have to call into the LoggingEventSource on every event just to drop the data.   

Logically what we are going is dynamically adding a LoggingProvider when an EventListener tells us to.  That seems like a perfectly valid thing to do, and it avoids ANY overhead when it the EventSource is off.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,234 @@
...
+        internal static readonly LoggingEventSource Instance = new LoggingEventSource();
+
+        internal static readonly LogLevel LoggingDisabled = LogLevel.None + 1;
+
+        private string _filterSpec;
+        private EventSourceLoggerProvider _loggingProviders;
+
+        internal EventSourceLoggerProvider CreateLoggerProvider(ILoggerFactory factory)
+        {
+            lock (this)
```
davidfowl: it's it better to lock a lock object?

karolz-ms: It is a concern if the object that you are locking on is publicly available (because someone else might take a lock on it from outside, screwing up your logic). That is not the case here.

davidfowl: Just other bad code in the same assembly 😄

karolz-ms: OK, fair enough :-)

src/Microsoft.Extensions.Logging.EventSource/project.json
davidfowl: @pranavkm can you help out here?

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
davidfowl: It does avoid some overhead but we'd have to measure to see just how much. For now, better to follow the pattern that the other logger providers are following. If things are *very* expensive then we can revisit.

src/Microsoft.Extensions.Logging.EventSource/project.json
pranavkm: Could this target netstandard1.3 or higher instead?.

davidfowl: I just noticed the dependencies here looked messed up as well:

- Move "System.Threading": "4.0.11-*" to the netstandard1.5 dependencies node.
- Remove the "System.Runtime": "4.0.0" from dependencies under net46 and.
- If possible, target netstandard1.3 and remove the net46 framework altogether...

karolz-ms:
After my last checkin

1.	The “Travis” build failed complaining about references to types related to EventSource infrastructure (Keywords, Level). I strongly suspect this is a Mono issue
2.	The “AppVeyor” build succeeded, although all tests (not just mine) failed due to “cannot load GCHandle” issue

After the build issues are resolved this request should be ready to merge. I believe I have addressed all issues raised during the review

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,235 @@
...
+
+#if !NO_EVENTSOURCE_COMPLEX_TYPE_SUPPORT
+        private LoggingEventSource() : base(EventSourceSettings.EtwSelfDescribingEventFormat) { }
+#endif
+
+        /// <summary>
+        /// FormattedMessage() is called when ILogger.Log() is called. and the FormattedMessage keyword is active
+        /// This only gives you the human reasable formatted message. 
+        /// </summary>
+        [Event(1, Keywords = Keywords.FormattedMessage, Level = EventLevel.LogAlways)]
```
iron9light: [Event(1, Keywords = Keywords.FormattedMessage, Level = EventLevel.LogAlways, **Message = "{2} - {4}"**)]
How about add a formatter like this?
Indeed, I tried it, but it does not work. I'm using azure fabric and tried it in local.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,235 @@
...
+        internal void FormattedMessage(LogLevel Level, int FactoryID, string LoggerName, string EventId, string FormattedMessage)
```
iron9light: Should we add more method for different levels, like **FormattedMassageDebug**, **FormattedMessageInformation**, so we can use different level for Event attribute.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// <summary>
+    /// Extension methods for the <see cref="ILoggerFactory"/> class.
+    /// </summary>
+    public static class EventSourceLoggerFactoryExtensions
+    {
+        /// <summary>
+        /// Adds an event logger that is enabled for <see cref="LogLevel"/>.Information or higher.
+        /// </summary>
+        /// <param name="factory">The extension method argument.</param>
+        public static ILoggerFactory AddEventSourceLogger(this ILoggerFactory factory)
```
iron9light: public static ILoggerFactory AddEventSourceLogger(this ILoggerFactory factory**, string spec = null**)
How about we pass spec as a parameter here?

karolz-ms: I assume you are talking about "filter spec" that governs what data goes through the bridge from ILogger to EventSource. 

I am not sure about this suggestion. I think it would be confusing for the API user to tell the difference between the spec passed through EventSource commands vs. the spec passed here. Which one wins if there is a conflict? Having just one place to specify the spec is beneficial. Event more importantly, the spec is given by the consumer of the data. The AddEventSourceLogger() API is for the producer of the data. The producer should not care what spec is used.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: FormattedMessage is precisely what the producer of the data intended to expose. It is readily available from the event payload. I do not see an added benefit of using the Message for this event, it would be less flexible and redundant.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
karolz-ms: While we use a single ETW Event, the LogLevel is part of event payload. From practical standpoint (of indexing and querying the event data) it makes very little difference whether the event level information is expressed through ETW event level or through custom payload property (like here). So I suggest that what we have is "good enough" for now. We can certainly revisit this suggestion if we get more feedback about this.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
iron9light: I'm using EventSource in Azure fabric. Azure fabric provide a ETW viewer (as a VS extension) for local fabric services. It will show Timestamp, Event name and Message. You need to click one row to see the detail info. Anyway, I agree with you.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerFactoryExtensions.cs
iron9light: I'm a newbie to EventSource. I think you are right, I should use `EventSource.SendCommand` to set the filter spec.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -0,0 +1,241 @@
...
+    /// <summary>
+    /// The provider for the <see cref="EventSourceLogger"/>.
+    /// </summary>
+    internal class EventSourceLoggerProvider : ILoggerProvider
+    {
+        // A small integer that uniquely identifies the LoggerFactory assoicated with this LoggingProvider.
+        // Zero is illegal (it means we are uninitialized), and have to be added to the factory. 
+        private int _factoryID;
+
+        private readonly ILoggerFactory _loggerFactory;
```
davidfowl: You don't need to pass this anymore.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -0,0 +1,235 @@
...
+        /// </summary>
+        internal static readonly LoggingEventSource Instance = new LoggingEventSource();
+
+        internal static readonly LogLevel LoggingDisabled = LogLevel.None + 1;
+
+        private string _filterSpec;
+        private EventSourceLoggerProvider _loggingProviders;
+        private object _lockObj = new object();
+
+        internal EventSourceLoggerProvider CreateLoggerProvider(ILoggerFactory factory)
```
davidfowl: You don't have to pass the logger factory anymore.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
vancem: Can you be more specific?   Are you talking about the factoryID?     Can you explain your rationale for not needing it more?   It seems like EventListeners/ETW still have a model where there is only one logical instance of the logger, and thus if you are monitoring a scenario that has created several LoggerFactories (e.g. multi-tenant case), you still need to have the factory ID if you want to avoid ambiguity.

karolz-ms: I think he is talking about the loggerFactory, it is not used anymore. I will make the suggested change.

Eilon:
Looks good. I'm closing this because we'll end up merging https://github.com/aspnet/Logging/pull/448, which has some formatting/administrative fixes (e.g. version number updates). Thanks!!!

PR #426, opened 2016-05-13
Reducing the amount of time ConsoleLogger locks

DavidObando:
The code has been improved by:
 - Pre-computing the state before locking
 - Reducing the amount of allocations of color objects
 - Reducing the amount of calls to Console.Write to the minimum, while still supporting colors.

Tests have been adjusted after the changes, and a couple new tests were added. This PR is intended to be merged into the 'dev' branch.

/cc @rynowak @sivagms @anpete @Eilon @SajayAntony @mnltejaswini @muratg @DamianEdwards @danroth27 @divega

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -15,11 +15,10 @@ namespace Microsoft.Extensions.Logging.Console
         private static readonly object _lock = new object();
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
+        private static readonly string _newLineWithMessagePadding;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
-
-        private const int _indentation = 2;
 
         private IConsole _console;
         private Func<string, LogLevel, bool> _filter;
@@ -28,6 +27,7 @@ namespace Microsoft.Extensions.Logging.Console
         {
             var logLevelString = GetLogLevelString(LogLevel.Information);
             _messagePadding = new string(' ', logLevelString.Length + _loglevelPadding.Length);
+            _newLineWithMessagePadding = Environment.NewLine + _messagePadding;
         }
 
         public ConsoleLogger(string name, Func<string, LogLevel, bool> filter, bool includeScopes)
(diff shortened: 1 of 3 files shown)


rynowak:
ConsoleColors is a struct.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -15,10 +15,20 @@ public class ConsoleLogger : ILogger
...
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
+        private static readonly string _newLineWithMessagePadding;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
 
+        // do not change user's background color except for Critical
+        private readonly ConsoleColors CriticalConsoleColors = new ConsoleColors(ConsoleColor.White, ConsoleColor.Red);
+        private readonly ConsoleColors ErrorConsoleColors = new ConsoleColors(ConsoleColor.Red, ConsoleColor.Black);
```
BrennanConroy: ConsoleColor.Black != user's background color

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,101 +108,116 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-        public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message)
+        public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message, Exception exception)
         {
-            // check if the message has any new line characters in it and provide the padding if necessary
-            message = ReplaceMessageNewLinesWithPadding(message);
-            var logLevelColors = GetLogLevelConsoleColors(logLevel);
-            var loglevelString = GetLogLevelString(logLevel);
+            var fullMessageBuilder = new StringBuilder();
+
+            ConsoleColors logLevelColors = DefaultConsoleColors;
```
Eilon: `var`, `var`, `var`...

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,101 +108,116 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+            string loglevelString = string.Empty;
+            bool loglevelUsesDefaultColors = true;
 
             // Example:
             // INFO: ConsoleApp.Program[10]
             //       Request received
-
-            lock (_lock)
+            if(!string.IsNullOrEmpty(message))
```
Eilon: Formatting

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
BrennanConroy: Oh, I think the comment is outdated, we made a decision to set black as background for everything to make sure the foreground color was visible

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,101 +108,116 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+                loglevelUsesDefaultColors = DefaultConsoleColors.Equals(logLevelColors);
 
-                // category and event id
-                // use default colors
-                WriteWithColor(
-                    DefaultConsoleColor,
-                    DefaultConsoleColor,
-                    _loglevelPadding + logName + $"[{eventId}]",
-                    newLine: true);
+                if (loglevelUsesDefaultColors)
```
Eilon: Why is this `if` here? It wasn't there before, was it?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,101 +108,116 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+        public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message, Exception exception)
         {
-            // check if the message has any new line characters in it and provide the padding if necessary
-            message = ReplaceMessageNewLinesWithPadding(message);
-            var logLevelColors = GetLogLevelConsoleColors(logLevel);
-            var loglevelString = GetLogLevelString(logLevel);
+            var fullMessageBuilder = new StringBuilder();
+
+            ConsoleColors logLevelColors = DefaultConsoleColors;
+            string loglevelString = string.Empty;
```
Eilon: The var names should be `logLevelXyz` (`Level` starts with capital `L`)

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
BrennanConroy: I understand it now, however it seems you don't need an if and can just assume the logLevel doesn't use default colors?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: @BrennanConroy 

> ConsoleColor.Black != user's background color

That's what was there before. I just made them instance members instead of producing a new ConsoleColors instance each time it was needed. (see [here](https://github.com/aspnet/Logging/blob/3d853c72eb9fb7e6182840d5c7811769dd431c62/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L239))

> Oh, I think the comment is outdated, we made a decision to set black as background for everything to make sure the foreground color was visible

Idem. This comment was simply moved with the colors. (see [here](https://github.com/aspnet/Logging/blob/3d853c72eb9fb7e6182840d5c7811769dd431c62/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L233))

rynowak:
Have any profiling/allocation data for this change?

It's definitely a good change to move all the logic outside of the lock, but there's a tradeoff being made here that I don't think we want to make with the use of the `StringBuilder`. That's going to allocate roughly 2x the amount of memory needed to represent the message (`char[]` in the `StringBuilder` + the final string for the whole message). Most of the things we need to write are already strings.

Breaking this up into a series of individual writes can avoid that, at the cost of taking the underlying console lock multiple times (which should be uncontended).

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: You're right. I was trying to fold the log level into the full message when it used default colors, but that case appears to not be true ever under the current implementation. I'll change this to always print the log level (special colors) separately from the full message (default colors).

Eilon:
I think one must-have is that we lock the console for an entire log message. Otherwise it will interleave multiple log calls and inter-mix them. While it would be "thread-safe," it would bring back the exact problem we must avoid.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
BrennanConroy: Yes I realize that, but the comment should have been updated back when I made some changes, hence the comment is outdated

rynowak:
>I think one must-have is that we lock the console for an entire log message.

Yep, I agree.

My comments refer to the locking done inside the console APIs. Every call to `Console.Write` does a [volatile read](https://github.com/dotnet/corefx/blob/master/src/System.Console/src/System/Console.cs#L133) and then calls into a [synchronized textwriter](https://github.com/dotnet/corefx/blob/master/src/System.Console/src/System/Console.cs#L433).

The tradeoff is whether or not to build as large of chunks of text as possible (this PR), or to do smaller writes to the console based on the pieces we already have (my suggestion). This would all be done while holding 'our' lock regardless. We'd have to measure to know, but this PR as is will allocate a significant amount of memory that we're not allocating today.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: @BrennanConroy so does that mean that `CriticalConsoleColors = new ConsoleColors(ConsoleColor.White, ConsoleColor.Red);` is wrong? Should it be white on black instead of white on red?

BrennanConroy: They're all correct. Should change the comment to something like "We must explicitly set the background color if we are setting the foreground color otherwise just setting one can look bad on the users console"

DavidObando:
:up::date:

I went ahead and did an allocation analysis for this change and confirmed that the pressure added by using the StringBuilder is enough to make an impact, so I went ahead and removed it in favor of having multiple calls to Console.Write, as @rynowak suggested. I also ran a short haul stress test on it via the MusicStore end-to-end scenario, and the results are satisfactory.

I sent an update to this PR addressing the all feedback obtained so far, let me know if you have additional observations.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,99 +97,104 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+                    {
+                        // log level string
+                        Console.Write(
+                            logLevelString,
+                            logLevelColors.Background,
+                            logLevelColors.Foreground);
+                    }
+
+                    // use default colors from here on
+                    if (!string.IsNullOrEmpty(logLevelString))
```
BrennanConroy: Shouldn't this check be `logIdentifier`? And if yes, then it shouldn't need a check since it always has something.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,99 +97,104 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+            // System.InvalidOperationException
+            //    at Namespace.Class.Function() in File:line X
+            if (exception != null)
+            {
+                // exception message
+                exceptionText = exception.ToString();
+                printException = true;
+            }
+
+            if (printMessage || printException)
```
BrennanConroy: We could just use 1 bool "printLog"

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: Thanks. Yes, it should be `logIdentifier`, but it may be empty when the Log method is invoked with an Exception object and a null message string.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: Agreed.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
BrennanConroy: Oh sorry, it only has something if the message had something. But still I think the check is wrong?

BrennanConroy:
Looks good

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,99 +97,104 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
-                    DefaultConsoleColor,
-                    _loglevelPadding + logName + $"[{eventId}]",
-                    newLine: true);
-
+                logIdentifier = _loglevelPadding + logName + "[" + eventId + "]";
                 // scope information
                 if (IncludeScopes)
                 {
-                    var scopeInformation = GetScopeInformation();
+                    scopeInformation = GetScopeInformation();
```
kichalla: Since this call is now moved outside the lock, wondering if its going to mess up the data. There could be multiple threads each with its own scope tree, but while one of threads gains the lock and is trying to write to the console, another thread could modify the scope tree information giving misleading results. I think this should stay within the lock as before?

kichalla:
:watch:

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
rynowak: `scopeInformation` is going to access TLS/ExecutionContext. It's the opposite of the usual problem of threading. 

> another thread could modify the scope tree information giving 

This won't happen. The scope information is not accessible to another thread.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
DavidObando: It is wrong. Will fix.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
kichalla: Cool. Thanks.

DavidObando:
:up: :date: Let me know if this looks good. Thanks!

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -97,99 +97,103 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+                        Console.Write(
+                            logLevelString,
+                            logLevelColors.Background,
+                            logLevelColors.Foreground);
+                    }
+
+                    // use default colors from here on
+                    if (!string.IsNullOrEmpty(logIdentifier))
+                    {
+                        Console.WriteLine(
```
BrennanConroy: You **_could_** put this in the previous if check since if `logLevelString` is set then `logIdentifier` is set too, but that might be a bit too much :)

BrennanConroy:
Wait a few hours for others to have time to review, otherwise :shipit:

Eilon:
:shipit:

Issue #417, opened 2016-05-06
Console logger warnings aren't getting colorized on OSX


rynowak:
Console logger warnings aren't getting colorized on OSX. Was looking for yellow, couldn't find it.

muratg:
@BrennanConroy

BrennanConroy:
960e571ff7c735e461afe679eda87c750ced1451

PR #430, opened 2016-05-20
Fix unix Console colors

BrennanConroy:
#417 
@davidfowl @kichalla @muratg 
screen shot 2016-05-20 at 9 40 50 am

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -236,7 +236,7 @@ namespace Microsoft.Extensions.Logging.Console
                 case LogLevel.Critical:
                     return new ConsoleColors(ConsoleColor.White, ConsoleColor.Red);
                 case LogLevel.Error:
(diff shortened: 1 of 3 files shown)


kichalla:
`fail` is hard to read. 

:shipit:

BrennanConroy:
Agreed, which is why I got a screen shot. Should we do something about it?

muratg:
@BrennanConroy What are the other possible colors we can put instead of red? 

Also, did you try both types of reds?

BrennanConroy:
We decided to change fail to red background, black text

JunTaoLuo:
@BrennanConroy update screenshot? Also doesn't that make `crit` and `fail` look almost identical? Is there an orange color that may work better?

BrennanConroy:
Possible colors are in this color table https://en.wikipedia.org/wiki/ANSI_escape_code#Colors

BrennanConroy:
screen shot 2016-05-23 at 9 16 46 am

JunTaoLuo:
:shipit:

davidfowl:
:shipit:

PR #431, opened 2016-05-20
Allow configuring the trace source logger without a TraceListener ins…

moozzyk:
…tance

Addresses parking lot/I-#405


src/Microsoft.Extensions.Logging.TraceSource/TraceSourceLoggerProvider.cs
```diff
@@ -115,7 +119,10 @@ public void Dispose()
             {
                 _disposed = true;
                 _rootTraceListener.Flush();
-                _rootTraceListener.Dispose();
+                if (_rootTraceListener != null)
```
davidfowl: This is so old school

```C#
_rootTraceListener?.Dispose();
```

💃

davidfowl:
:shipit:

cesarbs:
:+1:

src/Microsoft.Extensions.Logging.TraceSource/TraceSourceLoggerProvider.cs
moozzyk:
```
10 PRINT "I AM NOT OLDSCHOOL"
20 GOTO 10
```

Issue #427, opened 2016-05-16
Update extension AddDebug() to accept an IConfiguration just like others (e.g. AddConsole()) 


muratg:
Following up on https://github.com/aspnet/Release/issues/31#issuecomment-219190918

pakrym:
Now all filters are consoled by configuration.

PR #434, opened 2016-05-22
Make DebugLogger and EventLogLogger externally configurable

cesarbs:
Addresses #427 and more.

Right now, only `ConsoleLogger` can be configured from an `IConfiguration` instance. This change makes `DebugLogger` and `EventLogLogger` configurable via `IConfiguration`, also supporting automatic reloading.

The changes are mostly additive, but there's a breaking change in that I've removed `ConfigurationConsoleLoggerSettings`, `ConsoleLoggerSettings` and `IConsoleLoggerSettings`.

To prevent duplicating logic across all three loggers, I've replaced `ConfigurationConsoleLoggerSettings` with `ConfigurableLoggerSettings` and put that in `Microsoft.Extensions.Logging.Abstractions`.

Along with that, I've introduced an abstract type named `ConfigurableLoggerProvider`, which `ConsoleLoggerProvider`, `DebugLoggerProvider` and `EventLogLoggerProvider` now subclass. `ConfigurableLoggerProvider` contains the configuration logic which used to reside in `ConsoleLoggerProvider`.

I don't think the filter logger and the TraceSource logger fit into this new model, so I've left them as is.

A few points pending:

* For now I've put `ConfigurableLoggerProvider` and `ConfigurableLoggerSettings` in `Microsoft.Exntesions.Logging.Abstractions`, but I think they might warrant their own package (`Microsoft.Extensions.Logging.Configuration`?). I also thought of putting them in `Microsoft.Extensions.Logging`, but that would change other loggers' dependencies from depending on that package instead of only depending on the `.Abstractions` one.

* `DebugLogger` and `EventLogLogger` now have an `IncludeScopes` property, but it doesn't do anything right now.

* More tests.

cc @kichalla @BrennanConroy @lodejard @davidfowl @Eilon @muratg


src/Microsoft.Extensions.Logging.Abstractions/ConfigurableLoggerProvider.cs
```diff
@@ -0,0 +1,121 @@
...
+                throw new ArgumentNullException(nameof(filter));
+            }
+
+            _filter = filter;
+            _settings = new ConfigurableLoggerSettings(new ConfigurationBuilder()
+                .Add(new MemoryConfigurationSource
+                {
+                    InitialData = new Dictionary<string, string>
+                    {
+                        ["IncludeScopes"] = "false"
```
BrennanConroy: Shouldn't you be using the value of `includeScopes` here?

src/Microsoft.Extensions.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -24,10 +26,24 @@ public static ILoggerFactory AddDebug(this ILoggerFactory factory)
         /// Adds a debug logger that is enabled as defined by the filter function.
         /// </summary>
         /// <param name="factory">The extension method argument.</param>
+        /// <param name="includeScopes">A value which indicates whether log scope information should be displayed
+        /// in the output.</param>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, bool includeScopes)
+        {
+            factory.AddDebug((n, l) => l >= LogLevel.Information, false);
```
BrennanConroy: Use `includeScopes`?

src/Microsoft.Extensions.Logging.Debug/DebugLoggerFactoryExtensions.cs
```diff
@@ -24,10 +26,24 @@ public static ILoggerFactory AddDebug(this ILoggerFactory factory)
...
+        /// </summary>
+        /// <param name="factory">The extension method argument.</param>
         /// <param name="filter">The function used to filter events based on the log level.</param>
-        public static ILoggerFactory AddDebug(this ILoggerFactory factory, Func<string, LogLevel, bool> filter)
+        /// <param name="includeScopes">A value which indicates whether log scope information should be displayed
+        /// in the output.</param>
+        public static ILoggerFactory AddDebug(this ILoggerFactory factory, Func<string, LogLevel, bool> filter, bool includeScopes)
         {
-            factory.AddProvider(new DebugLoggerProvider(filter));
+            factory.AddProvider(new DebugLoggerProvider(filter, false));
```
BrennanConroy: Again, use `includeScopes`?

Eilon:
Moving this to 1.0.1 as well.

muratg:
@cesarbs what are we doing with this one?

cesarbs:
@muratg I'll get back to it. I think I'll go over another "full iteration" on the idea since I'm not happy with a number of things in this first attempt.

muratg:
@cesarbs this one good to close then?

Issue #405, opened 2016-04-13
Allow configuring the trace source logger without having to flow a TraceListener instance


PinpointTownes:
Unlike Katana's `DiagnosticsLoggerFactory`, the vNext equivalent now requires a `TraceListener`, which makes the trace source logger a bit painful to use in practice, since you can't simply rely on the listeners defined at the static level (that may be registered via the .config file on .NET Desktop)

Please consider updating `TraceSourceLoggerProvider` to make this parameter optional and offering a simpler `AddTraceSource` that doesn't require a specific trace listener.

@davidfowl do you remember why you had removed the parameterless constructor? https://github.com/aspnet/Logging/pull/143/files#diff-ca74d694fa53328497b21b4a9528dae5L28

/cc @muratg

muratg:
@davidfowl why did we remove this constructor?

muratg:
Tentatively putting it in 1.0.0.  We'll discuss after RC2 ships.

moozzyk:
Fixed in 5019dd22aa

PinpointTownes:
@moozzyk thanks!

Quick question: any particular reason to not provide an overload of `AddTraceSource` taking no `TraceListener` parameter? Because it's great to make it optional in the provider class itself, but currently, an exception will be thrown if you try to use the `AddTraceSource` extensions:

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceFactoryExtensions.cs#L27
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.TraceSource/TraceSourceFactoryExtensions.cs#L50

moozzyk:
@PinpointTownes - thanks for pointing this out. I will look into this.

moozzyk:
Added overloads that don't take TraceListener e6af0dc

PR #435, opened 2016-05-23
Addendum to #405

moozzyk:
We enabled passing null TraceListener but extension methods were not updated accordingly and would still not allow to pass a null TraceListner.


davidfowl:
Don't flow null, make new overloads.

moozzyk:
:up::date:

Issue #445, opened 2016-06-05
Problems building sample app


kimbell:
This little journey started when I was trying to figure out why Exception information wasn't being output on the console logger. I downloaded the latest code and tried building it. I encountered two problems. 

Microsoft.Extensions.Logging.Filter.xproj had the following OutputPath
`<OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\</OutputPath>`

This caused other projects from building since it couldn't find the assembly. Other projects in the solution use
`<OutputPath Condition="'$(OutputPath)'=='' ">..\..\artifacts\bin\</OutputPath>`

adding in the missing bin folder fixed that issue. The second issue was that logging.json isn't being copied to the output folder, and an exception is thrown when it cannot be found. I haven't found the new way of 'copy to output folder'

muratg:
@BrennanConroy Could you take a look?

BrennanConroy:
Fixed in https://github.com/aspnet/Logging/commit/1a9808518ec58bef7f4afbbcab76719a30cb6b33

PR #451, opened 2016-06-10
Fix .xproj to build in VS

BrennanConroy:
#445


moozzyk:
How does this address the second issue mentioned in the bug: 
 _The second issue was that logging.json isn't being copied to the output folder, and an exception is thrown when it cannot be found. I haven't found the new way of 'copy to output folder'_

BrennanConroy:
I already fixed that when I was making the sample run in a different issue

moozzyk:
:ship::it:

Issue #455, opened 2016-06-21
AnsiLogConsole uses non-standard background values


BrennanConroy:
In https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/Internal/AnsiLogConsole.cs#L104 we use the 100-107 range when we should be using 40-47 see https://en.wikipedia.org/wiki/ANSI_escape_code#graphics

BrennanConroy:
https://github.com/aspnet/Logging/commit/7457578257cfb1eb2940b26818798a017efc1921

PR #456, opened 2016-06-24
Use standard ansi color codes

BrennanConroy:
#455


BrennanConroy:
Before on CentOS:
before-centos

After on CentOS:
centos

After on Ubuntu:
ubuntu

BrennanConroy:
@cesarbs @Eilon @kichalla

Eilon:
:shipit:

PR #458, opened 2016-07-01
Fixed ArgumentNullException on ConsoleLoggerProvider

marcogrcr:
`ConsoleLoggerProvider.OnConfigurationReload(object)` is supposed to support that `IConsoleLoggerSettings.Reload()` might return null.

This fix prevents an ArgumentNullException to be thrown when this happens.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Extensions.Logging.Console
             foreach (var logger in _loggers.Values)
             {
                 logger.Filter = GetFilter(logger.Name, _settings);
-                logger.IncludeScopes = _settings.IncludeScopes;
+                logger.IncludeScopes = _settings?.IncludeScopes ?? false;
             }
 
             // The token will change each time it reloads, so we need to register again.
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -52,7 +52,7 @@ private void OnConfigurationReload(object state)
             foreach (var logger in _loggers.Values)
             {
                 logger.Filter = GetFilter(logger.Name, _settings);
-                logger.IncludeScopes = _settings.IncludeScopes;
+                logger.IncludeScopes = _settings?.IncludeScopes ?? default(bool);
```
davidfowl: instead of `default(bool)` can you just make it false

marcogrcr: Done, changed!

Just out of curiosity. Was the request made for coding style reasons, or are they any drawbacks for using `default(bool)` instead of `false`? (I suspect that theoretically `default(bool)` could change to `true`, but we both know that such breaking change would never be allowed).

BrennanConroy:
I'm fine with the fix, but is this actually a scenario people will run into? As soon as `Reload` returns `null` the logger becomes useless because it will never get another `_settings` object and the `GetFilter` function will return a filter than is always `null`

BrennanConroy:
Thanks for your contribution! https://github.com/aspnet/Logging/commit/b5f8f914898d38e6f2740616516eea28cd1c514f

PR #401, opened 2016-04-06
Console Logger Settings should not have to respect Enum case

ORuban:
See https://github.com/aspnet/Logging/issues/381

Issue #381, opened 2016-03-10
Console Logger Settings should not have to respect Enum case


Yves57:
The configuration parameter "LogLevel" respects the LogLevel Enum case when parsed (see `Enum.TryParse<LogLevel>(value, out level)` in `ConfigurationConsoleLoggerSettings.TryGetSwitch()`. It is easy for a user to make a mistake when typing a value in a configuration file.
I think the code should be `Enum.TryParse<LogLevel>(value, true, out level)`.

PR #460, opened 2016-07-07
Ignore case when parsing LogLevel string in the Console Logger Settings.

Yves57:


Yves57:
See #381

cesarbs:
I like it. cc @BrennanConroy @kichalla

BrennanConroy:
Yeah, it looks good

BrennanConroy:
Although looks like some tests are failing

src/Microsoft.Extensions.Logging.Console/ConfigurationConsoleLoggerSettings.cs
```diff
@@ -62,7 +62,7 @@ public bool TryGetSwitch(string name, out LogLevel level)
                 level = LogLevel.None;
                 return false;
             }
-            else if (Enum.TryParse<LogLevel>(value, out level))
+            else if (Enum.TryParse<LogLevel>(value, true, out level))
```
pranavkm: Could you use a named param here - `Enum.TryParse<LogLevel>(value, ignoreCase: true, result: out level)`?

pranavkm:
@BrennanConroy those tests seem pretty unrelated.

moozzyk:
I am not sure if it adds any value. You can make typos everywhere... The exception message when there is a mismatch could be better though - it could list all the possible values.

hishamco:
This is a duplicate of PR #401

muratg:
Dup of #401

PR #461, opened 2016-07-08
Switched URL to elmah.io extensions logger

ThomasArdal:
We've created a new logger for elmah.io based on .NET Core.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This project is part of ASP.NET Core. You can find samples, documentation and ge
 Community projects adapt _Microsoft.Extensions.Logging_ for use with different back-ends.
 
  * [Serilog](https://github.com/serilog/serilog-framework-logging) - provider for the Serilog library
- * [elmah.io](https://github.com/elmahio/Elmah.Io.Framework.Logging) - provider for the elmah.io service
+ * [elmah.io](https://github.com/elmahio/Elmah.Io.Extensions.Logging) - provider for the elmah.io service
  * [Loggr](https://github.com/imobile3/Loggr.Extensions.Logging) - provider for the Loggr service
  * [NLog](https://github.com/NLog/NLog.Extensions.Logging) - provider for the NLog library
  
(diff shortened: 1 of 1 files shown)


ThomasArdal:
Seems like there's an existing problem with the continuous build. This PR contains markdown changes only :smile:

304NotModified:
To bad it isn't .Extensions.Logging like the others

ThomasArdal:
@304NotModified Very interested in hearing your throughs on why? The AspNetCore package contains everything to support ASP.NET Core. We could split it into Elmah.Io.Extensions.Logging and make Elmah.Io.AspNetCore dependent of that, but I think that it would confuse people and open up for dll-dependency-hell :smile:

ThomasArdal:
@304NotModified Now you got me all in doubt :smiley:. Having two NuGet packages could make sense. If you "just" want to use [elmah.io](https://elmah.io) as a logging framework for ASP.NET Logging, you could use Elmah.Io.Extensions.Logging and if you wan't the full "ELMAH" experience, you would install Elmah.Io.AspNetCore (as well as Elmah.Io.Extensions.Logging).

304NotModified:
yes, we (NLog) have also two packages, NLog.Extensions.Logging & NLog.Web.AspNetCore (And NLog of course)

The confusing part is that Microsoft.Extensions.Logging isn't only for ASP.NET Core (but also .NET Core without ASP)

ThomasArdal:
Now I understand. The aspnet/Logging had me confused as well. If ASP.NET Logging is in fact .NET Core logging, Elmah.Io.Extensions.Logging makes total sense. Will start working on that. Thx!

ThomasArdal:
@304NotModified There's an [Elmah.Io.Extensions.Logging](https://github.com/elmahio/Elmah.Io.Extensions.Logging) now. Thanks for your input.

davidfowl:
@ThomasArdal 😄

PR #464, opened 2016-07-14
Add ITestSink interface.

cesarbs:
I want to write a different test sink and right now there's no simple way to it. Introducing this interfaces allows for different test sinks to be implement according to testing needs.

cc @kichalla @BrennanConroy @pranavkm

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/TestLoggerFactory.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/TestLoggerFactory.cs
@@ -7,10 +7,10 @@ namespace Microsoft.Extensions.Logging.Testing
 {
     public class TestLoggerFactory : ILoggerFactory
     {
-        private readonly TestSink _sink;
+        private readonly ITestSink _sink;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Logging.Testing/TestLoggerFactory.cs
```diff
@@ -7,10 +7,10 @@ namespace Microsoft.Extensions.Logging.Testing
 {
     public class TestLoggerFactory : ILoggerFactory
     {
-        private readonly TestSink _sink;
+        private readonly ITestSink _sink;
         private readonly bool _enabled;
 
-        public TestLoggerFactory(TestSink sink, bool enabled)
+        public TestLoggerFactory(ITestSink sink, bool enabled)
         {
```
moozzyk: Isn't it a breaking change though?

cesarbs: We don't ship this package.

PR #465, opened 2016-07-15
AppDomain safety

BrennanConroy:
@Tratcher @davidfowl @kichalla 

Related to https://github.com/aspnet/Caching/issues/203

### DIFF
--- a/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Extensions.Logging.Test
         private const string _state = "This is a test, and {curly braces} are just fine!";
         private Func<object, Exception, string> _defaultFormatter = (state, exception) => state.ToString();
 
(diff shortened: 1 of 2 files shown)


test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -642,6 +642,38 @@ public void CallingBeginScopeOnLogger_ReturnsNonNullableInstance()
...
+
+            // Act
+            domain.DoCallBack(DomainFunc);
+            AppDomain.Unload(domain);
+            var disposable = logger.BeginScope("Scope1");
+            logger.LogInformation("Test");
+
+            // Assert
+            Assert.NotNull(disposable);
+            Assert.Equal(4, sink.Writes.Count);
```
kichalla: I can't recall how these writes look like...how is the count 4 here? (for example, the DomainFunc creates a new logger and new sink, so we are including it's numbers right?)

kichalla: Ah scratch that. Got it.

BrennanConroy:
@kichalla Is it good?

kichalla:
:shipit:

Issue #328, opened 2015-12-29
EventSource implementation


cwe1ss:
Hi, I created a logger implementation which forwards messages to a System.Diagnostics.Tracing.EventSource and was wondering if this would be interesting for you since EventSource is one of the recommended ways to log in the .NET framework.

https://github.com/cwe1ss/Logging/tree/EventSource/src/Microsoft.Extensions.Logging.EventSource

it has the following features:
* it's a general purpose EventSource called "LoggingEventSource" with the name "Microsoft-Extensions-Logging"
* it has events for every log level (Trace, Debug, ...)
* ILogger.BeginScopeImpl is implemented by calling an ActivityStart and ActivityStop event on the EventSource (which automatically creates an ActivityId)
* ILogValues from the state parameter are stored as a JSON serialized dictionary called "Properties". Right now this uses Newtonsoft.Json - In .NET 4.6 this could be changed to the "rich payload" feature, however most tools seem to not yet support this (Semantic Logging Application Block; Visual Studio "Diagnostic Events" window [from the Azure Service Fabric SDK])
* Exception details are also added as separate properties to this dictionary

If this is interesting for you, I'd be happy to do necessary changes and create a pull request.

@vancem - feedback would be appreciated!

best regards, christian

muratg:
@lodejard @Eilon @davidfowl

muratg:
@vancem any updates?

vancem:
Sorry for the delay.   This last week I have been filling in for a lead who is on leave, and there simply are not enough hours in the day.   I will take a look at this later today.

vancem:
This is a straightforward hookup of ILogger to an EventSource.    There are many details of exactly how the transformation gets done that are interesting to vet (which is part of the reason it has been taking me a while to give any feedback).    Currently all events end up in one EventSource and there is no filtering control at the component level (only the level (volume) control).   It also dumps things as JSON which are probably better done as dynamic events.   In short, there are issues to resolve.    It is fine as something that unblocks someone in the short term, however as a larger standard, I am more concerned. 

My main issue is that I would like ILogger data to flow into EventSource automatically, without the need for application developers to explicitly wire up the ILogger to a particular EventSource.   There is work on such a thing, and ideally we would use that.

I realize that there is this tradeoff between having something vs waiting for something that is potentially better.  We can use this as a fallback, but I would like to see if we can build a more automatic alternative first.

cwe1ss:
[...]

PR #467, opened 2016-07-21
Fix EventLogger in .NET 4.5.2

BrennanConroy:
The behavior of `OnEventCommand` callback changed between `.NET 4.5.2` and `.NET 4.6`. In the latter the internal `EventSource` level has changed before calling the `OnEventCommand` callback which was behavior we relied on to set the loggers level. Because it isn't set before the callback in `.NET 4.5.2` we delay the original behavior of the callback until a time after the Event change occurs.
#328

### DIFF
--- a/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
@@ -37,6 +37,9 @@ namespace Microsoft.Extensions.Logging.EventSourceLogger
         /// </summary>
         public ILogger CreateLogger(string categoryName)
         {
+            // need to check if the filter spec and internal event source level has changed
+            // and update the _defaultLevel if it has
+            _eventSource.ApplyFilterSpec();
             var newLogger = _loggers = new EventSourceLogger(categoryName, _factoryID, _eventSource, _loggers);
             newLogger.Level = ParseLevelSpecs(_filterSpec, _defaultLevel, newLogger.CategoryName);
             return newLogger;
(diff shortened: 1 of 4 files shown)


BrennanConroy:
@karolz-ms @kichalla

karolz-ms:
Looks good, but I would add a comment explaining why SetFilterSpec is not immediately called on logging providers when EventSource.SetFilterSpec is called

BrennanConroy:
Good point

davidfowl:
Some explanation would be good in the PR description, what was wrong and what was fixed

BrennanConroy:
@davidfowl Added some description

BrennanConroy:
🆙 📅

BrennanConroy:
Ping

davidfowl:
:shipit:

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -226,9 +227,24 @@ protected override void OnEventCommand(EventCommandEventArgs command)
...
+
+            // In .NET 4.5.2 the internal EventSource level hasn't been correctly set
+            // when this callback is invoked. To still have the logger behave correctly
+            // in .NET 4.5.2 we delay checking the level until the logger is used the first
+            // time after this callback.
+            _checkLevel = true;
+        }
+
+        [NonEvent]
+        internal void VerifyLevel()
```
davidfowl: Any concurrency concerns? Can this be called in parallel?

BrennanConroy:
🆙 📅 
Added lock and added another location to check level

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -37,6 +37,7 @@ public EventSourceLoggerProvider(LoggingEventSource eventSource, EventSourceLogg
         /// </summary>
         public ILogger CreateLogger(string categoryName)
         {
+            _eventSource.VerifyLevel();
```
davidfowl: Add a comment wherever you call VerifyLevel() (which is the wrong name)

Issue #466, opened 2016-07-21
Implement a logger for Azure WebApps


victorhurdugaci:
It needs to work with the configuration options in the Azure portal (file, blob, levels) and respond to changes in the `settings.json` file that the portal modifies.

It must no-op when running outside of WebApps.

PR #468, opened 2016-07-21
A logger for Azure WebApps

victorhurdugaci:
WIP implementation for https://github.com/aspnet/Logging/issues/466. Sending to get some early feedback. Tests are on the way.

Please review @pakrym @kichalla @davidfowl

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLogger.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.Abstractions.Internal
+{
+    public class NullLogger : ILogger
+    {
+        public static NullLogger Instance { get; } = new NullLogger();
+
+        private NullLogger()
+        {
+        }
+
+        public IDisposable BeginScope<TState>(TState state)
+        {
+            return NullScope.Instance;
+        }
+
+        public bool IsEnabled(LogLevel logLevel)
+        {
+            return false;
+        }
+
+        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+        {
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLoggerProvider.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.Abstractions.Internal
+{
+    public class NullLoggerProvider : ILoggerProvider
+    {
+        public static NullLoggerProvider Instance { get; } = new NullLoggerProvider();
+
+        private NullLoggerProvider()
+        {
+        }
+
+        public ILogger CreateLogger(string categoryName)
+        {
+            return NullLogger.Instance;
+        }
+
+        public void Dispose()
+        {
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsFactoryExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsFactoryExtensions.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.AzureWebAppDiagnostics;
+using Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Extension methods for <see cref="AzureWebAppDiagnosticsLoggerProvider"/>.
+    /// </summary>
+    public static class AzureWebAppDiagnosticsFactoryExtensions
+    {
+        /// <summary>
+        /// Adds an Azure Web Apps diagnostics logger.
+        /// </summary>
+        /// <param name="factory">The extension method argument</param>
+        /// <param name="fileSizeLimitMb">A strictly positive value representing the maximum log size in megabytes. Once the log is full, no more message will be appended</param>
+        public static ILoggerFactory AddAzureWebAppDiagnostics(this ILoggerFactory factory, int fileSizeLimitMb = FileLoggerProvider.DefaultFileSizeLimitMb)
+        {
+            if (WebAppContext.Default.IsRunningInAzureWebApp)
+            {
+                // Only add the provider if we're in Azure WebApp. That cannot change once the apps started
+                factory.AddProvider(new AzureWebAppDiagnosticsLoggerProvider(WebAppContext.Default, fileSizeLimitMb));
+            }
+            return factory;
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsLoggerProvider.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.Abstractions.Internal;
+using Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal;
+
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics
+{
+    /// <summary>
+    /// Logger provider for Azure WebApp.
+    /// </summary>
+    public class AzureWebAppDiagnosticsLoggerProvider : ILoggerProvider
+    {
+        private readonly IWebAppLogConfigurationReader _configurationReader;
+
+        private readonly ILoggerProvider _innerLoggerProvider;
+        private readonly bool _runningInWebApp;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="AzureWebAppDiagnosticsLoggerProvider"/> class.
+        /// </summary>
+        public AzureWebAppDiagnosticsLoggerProvider(WebAppContext context, int fileSizeLimitMb)
+        {
+            _configurationReader  = new WebAppLogConfigurationReader(context);
+
+            var config = _configurationReader.Current;
+            _runningInWebApp = config.IsRunningInWebApp;
+
+            if (!_runningInWebApp)
+            {
+                _innerLoggerProvider = NullLoggerProvider.Instance;
+            }
+            else
+            {
+                _innerLoggerProvider = new FileLoggerProvider(_configurationReader, fileSizeLimitMb);
+
+                if (!string.IsNullOrEmpty(config.BlobContainerUrl))
+                {
+                    // TODO: Add the blob logger by creating a composite inner logger which calls
+                    // both loggers
+                }
+            }
+        }
+
+        /// <inheritdoc />
+        public ILogger CreateLogger(string categoryName)
+        {
+            return _innerLoggerProvider.CreateLogger(categoryName);
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            _innerLoggerProvider.Dispose();
+            _configurationReader.Dispose();
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
@@ -1,0 +1,110 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using Serilog.Core;
+using Serilog.Events;
+
+// TODO: Might want to consider using https://github.com/jezzsantos/Serilog.Sinks.Async 
+// instead of this, once that supports netstandard
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+    /// <summary>
+    /// A background sink for Serilog.
+    /// </summary>
+    public class BackgroundSink : ILogEventSink, IDisposable
+    {
+        /// <summary>
+        /// The default queue size.
+        /// </summary>
+        public const int DefaultLogMessagesQueueSize = 1024;
+
+        private readonly CancellationTokenSource _disposedTokenSource = new CancellationTokenSource();
+        private readonly CancellationToken _disposedToken;
+
+        private readonly BlockingCollection<LogEvent> _messages;
+        private readonly Thread _workerThread;
+
+        private ILogEventSink _innerSink;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="BackgroundSink"/> class.
+        /// </summary>
+        /// <param name="innerSink">The inner sink which does the actual logging</param>
+        /// <param name="maxQueueSize">The maximum size of the background queue</param>
+        public BackgroundSink(ILogEventSink innerSink, int? maxQueueSize)
+        {
+            if (innerSink == null)
+            {
+                throw new ArgumentNullException(nameof(innerSink));
+            }
+
+            _disposedToken = _disposedTokenSource.Token;
+
+            if (maxQueueSize == null || maxQueueSize <= 0)
+            {
+                _messages = new BlockingCollection<LogEvent>(new ConcurrentQueue<LogEvent>());
+            }
+            else
+            {
+                _messages = new BlockingCollection<LogEvent>(new ConcurrentQueue<LogEvent>(), maxQueueSize.Value);
+            }
+
+            _innerSink = innerSink;
+
+            _workerThread = new Thread(Worker);
+            _workerThread.Name = GetType().Name;
+            _workerThread.IsBackground = true;
+            _workerThread.Start();
+        }
+
+        /// <inheritdoc />
+        public void Emit(LogEvent logEvent)
+        {
+            if (!_disposedToken.IsCancellationRequested)
+            {
+                _messages.Add(logEvent);
+            }
+        }
+
+        /// <summary>
+        /// Disposes this object instance.
+        /// </summary>
+        public virtual void Dispose()
+        {
+            lock (_disposedTokenSource)
+            {
+                if (!_disposedTokenSource.IsCancellationRequested)
+                {
+                    _disposedTokenSource.Cancel();
+                }
+
+                // Wait for the thread to complete before disposing the resources
+                _workerThread.Join(5 /*seconds */ * 1000);
+                _messages.Dispose();
+            }
+        }
+
+        private void Worker()
+        {
+            try
+            {
+                foreach (var logEvent in _messages.GetConsumingEnumerable(_disposedToken))
+                {
+                    PassLogEventToInnerSink(logEvent);
+                }
+            }
+            catch (OperationCanceledException)
+            {
+                // Do nothing, we just cancelled the task
+            }
+        }
+
+        private void PassLogEventToInnerSink(LogEvent logEvent)
+        {
+            _innerSink.Emit(logEvent);
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
@@ -1,0 +1,72 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.IO;
+using Serilog;
+using Serilog.Formatting.Display;
+using Serilog.Sinks.RollingFile;
+
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+    /// <summary>
+    /// A file logger for Azure WebApp.
+    /// </summary>
+    public class FileLoggerProvider : SerilogLoggerProvider
+    {
+        /// <summary>
+        /// The default file size limit in megabytes
+        /// </summary>
+        public const int DefaultFileSizeLimitMb = 10;
+
+        // Two days retention limit is okay because the file logger turns itself off after 12 hours (portal feature)
+        private const int RetainedFileCountLimit = 2; // Days (also number of files because we have 1 file/day)
+
+        private const string OutputTemplate = "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level}] {Message}{NewLine}{Exception}";
+        private const string FileNamePattern = "diagnostics-{Date}.txt";
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="FileLoggerProvider"/> class.
+        /// </summary>
+        /// <param name="configReader">A configuration reader</param>
+        /// <param name="fileSizeLimit">A strictly positive value representing the maximum log size in megabytes. Once the log is full, no more message will be appended</param>
+        public FileLoggerProvider(IWebAppLogConfigurationReader configReader, int fileSizeLimit)
+            : base(configReader, (loggerConfiguration, webAppConfiguration) =>
+            {
+                if (string.IsNullOrEmpty(webAppConfiguration.FileLoggingFolder))
+                {
+                    throw new ArgumentNullException(nameof(webAppConfiguration.FileLoggingFolder), "The file logger path cannot be null or empty.");
+                }
+
+                var logsFolder = webAppConfiguration.FileLoggingFolder;
+                if (!Directory.Exists(logsFolder))
+                {
+                    Directory.CreateDirectory(logsFolder);
+                }
+                var logsFilePattern = Path.Combine(logsFolder, FileNamePattern);
+
+                var fileSizeLimitBytes = fileSizeLimit * 1024 * 1024;
+
+                var messageFormatter = new MessageTemplateTextFormatter(OutputTemplate, null);
+                var rollingFileSink = new RollingFileSink(logsFilePattern, messageFormatter, fileSizeLimitBytes, RetainedFileCountLimit);
+                var backgroundSink = new BackgroundSink(rollingFileSink, BackgroundSink.DefaultLogMessagesQueueSize);
+
+                loggerConfiguration.WriteTo.Sink(backgroundSink);
+            })
+        {
+        }
+
+        /// <inheritdoc />
+        protected override void OnConfigurationChanged(WebAppLogConfiguration newConfiguration)
+        {
+            if (!newConfiguration.FileLoggingEnabled)
+            {
+                LevelSwitcher.MinimumLevel = LogLevelDisabled;
+            }
+            else
+            {
+                LevelSwitcher.MinimumLevel = LogLevelToLogEventLevel(newConfiguration.FileLoggingLevel);
+            }
+        }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/IWebAppLogConfigurationReader.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/IWebAppLogConfigurationReader.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+
+    /// <summary>
+    /// The contract for a WebApp configuration reader.
+    /// </summary>
+    public interface IWebAppLogConfigurationReader : IDisposable
+    {
+        /// <summary>
+        /// Triggers when the configuration has changed.
+        /// </summary>
+        event EventHandler<WebAppLogConfiguration> OnConfigurationChanged;
+
+        /// <summary>
+        /// The current value of the configuration.
+        /// </summary>
+        WebAppLogConfiguration Current { get; }
+    }
+}
--- a/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
@@ -1,0 +1,125 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Serilog;
+using Serilog.Core;
+using Serilog.Events;
+
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+    /// <summary>
+    /// Represents a Serilog logger provider use for Azure WebApp.
+    /// </summary>
+    public abstract class SerilogLoggerProvider : ILoggerProvider
+    {
+        // Solution suggested by the Serilog creator http://stackoverflow.com/questions/30849166/how-to-turn-off-serilog
+        /// <summary>
+        /// The log level at which the logger is disabled.
+        /// </summary>
+        protected static LogEventLevel LogLevelDisabled = ((LogEventLevel)1 + (int)LogEventLevel.Fatal);
+
+        private readonly LoggingLevelSwitch _levelSwitch = new LoggingLevelSwitch();
+
+        private readonly IWebAppLogConfigurationReader _configReader;
+        private readonly ILoggerFactory _loggerFactory;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="SerilogLoggerProvider"/> class.
+        /// </summary>
+        /// <param name="configReader">The configuration reader</param>
+        /// <param name="configureLogger">The actions required to configure the logger</param>
+        public SerilogLoggerProvider(IWebAppLogConfigurationReader configReader, Action<LoggerConfiguration, WebAppLogConfiguration> configureLogger)
+        {
+            if (configReader == null)
+            {
+                throw new ArgumentNullException(nameof(configReader));
+            }
+            if (configureLogger == null)
+            {
+                throw new ArgumentNullException(nameof(configureLogger));
+            }
+
+            _configReader = configReader;
+            var webAppsConfiguration = configReader.Current;
+
+            configReader.OnConfigurationChanged += OnConfigurationChanged;
+
+            var loggerConfiguration = new LoggerConfiguration()
+                .MinimumLevel.ControlledBy(_levelSwitch);
+            configureLogger(loggerConfiguration, webAppsConfiguration);
+            var serilogLogger = loggerConfiguration.CreateLogger();
+            
+            OnConfigurationChanged(webAppsConfiguration);
+
+            _loggerFactory = new LoggerFactory();
+            _loggerFactory.AddSerilog(serilogLogger);
+        }
+
+        /// <summary>
+        /// The switch used the modify the logging level.
+        /// </summary>
+        protected LoggingLevelSwitch LevelSwitcher => _levelSwitch;
+
+        /// <summary>
+        /// Called when the configuration changes
+        /// </summary>
+        /// <param name="newConfiguration">The new configuration values</param>
+        protected abstract void OnConfigurationChanged(WebAppLogConfiguration newConfiguration);
+
+        /// <inheritdoc />
+        public ILogger CreateLogger(string categoryName)
+        {
+            return _loggerFactory.CreateLogger(categoryName);
+        }
+
+        /// <summary>
+        /// Disposes this object.
+        /// </summary>
+        public void Dispose()
+        {
+            _configReader.OnConfigurationChanged -= OnConfigurationChanged;
+            _loggerFactory.Dispose();
+        }
+
+        private void OnConfigurationChanged(object sender, WebAppLogConfiguration newConfiguration)
+        {
+            OnConfigurationChanged(newConfiguration);
+        }
+
+        /// <summary>
+        /// Converts a <see cref="LogLevel"/> object to <see cref="LogEventLevel"/>.
+        /// </summary>
+        /// <param name="logLevel">The log level to convert</param>
+        /// <returns>A <see cref="LogEventLevel"/> instance</returns>
+        protected static LogEventLevel LogLevelToLogEventLevel(LogLevel logLevel)
+        {
+            switch (logLevel)
+            {
+                case LogLevel.Trace:
+                    return LogEventLevel.Verbose;
+
+                case LogLevel.Debug:
+                    return LogEventLevel.Debug;
+
+                case LogLevel.Information:
+                    return LogEventLevel.Information;
+
+                case LogLevel.Warning:
+                    return LogEventLevel.Warning;
+
+                case LogLevel.Error:
+                    return LogEventLevel.Error;
+
+                case LogLevel.Critical:
+                    return LogEventLevel.Fatal;
+
+                case LogLevel.None:
+                    return LogLevelDisabled;
+
+                default:
+                    throw new ArgumentOutOfRangeException($"Unknown log level: {logLevel}");
+            }
+        }
+    }
+}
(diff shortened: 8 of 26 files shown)


src/Microsoft.Extensions.Logging.Abstractions/Internal/CompositeScope.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public CompositeScope(int count)
+        {
+            if (count > 2)
+            {
+                _disposable = new IDisposable[count - 2];
+            }
+        }
+
+        public void SetDisposable(int index, IDisposable disposable)
```
pakrym: Indexer?

src/Microsoft.Extensions.Logging.Abstractions/Internal/CompositeScope.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                _disposable0 = disposable;
+            }
+            else if (index == 1)
+            {
+                _disposable1 = disposable;
+            }
+            else
+            {
+                _disposable[index - 2] = disposable;
```
pakrym: check index or it will nullref is constructor had 1 or 2 and `SetDisposable` is called for index >2

src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLoggerProvider.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        public static NullLoggerProvider Instance => new NullLoggerProvider();
+
+        private NullLoggerProvider()
+        {
+        }
+
+        public ILogger CreateLogger(string categoryName)
+        {
+            return NullLogger.Instance;
```
pakrym: You can cache `NullLoggerProvider` instance inside `NullLoggerProvider`, it's enough to have one singleton.

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using Microsoft.Extensions.Logging.Abstractions.Internal;
+using Microsoft.Extensions.Logging.AzureWebApps.Internal;
+
+namespace Microsoft.Extensions.Logging.AzureWebApps
+{
+    /// <summary>
+    /// Logger provider for Azure WebApps
+    /// </summary>
+    public class AzureWebAppsLoggerProvider : ILoggerProvider
```
pakrym: Configuration reload?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+
+        private void Worker()
+        {
+            while (!_disposedToken.IsCancellationRequested)
+            {
+                // This block until either an item is available
+                // or the token is cancelled 
+                LogEvent logEvent = _messages.Take(_disposedToken);
```
pakrym: Why not `_messages.GetConsumingEnumerable`?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// <param name="fileLoggingEnabled">True if file logging is enabled, false otherwise</param>
+        /// <param name="fileLoggingLevel">The logging level for the file logger</param>
+        /// <param name="fileLoggingFolder">The folder where the file logger stores the logs</param>
+        /// <param name="blobLoggingEnabled">True if blob logging is enabled, false otherwise</param>
+        /// <param name="blobLoggingLevel">The logging level for the blob logger</param>
+        /// <param name="blobContainerUrl">The SAS endpoint where blob logs are stored</param>
+        public WebAppsLogConfiguration(
+            bool fileLoggingEnabled,
+            LogLevel fileLoggingLevel,
+            string fileLoggingFolder,
```
pakrym: In .NET its "directory" not "folder"

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfigurationReader.cs
```diff
@@ -0,0 +1,132 @@
...
+        private readonly IConfigurationRoot _configuration;
+        private readonly string _fileLogFolder;
+
+        private WebAppsLogConfiguration _latestConfiguration;
+        private IDisposable _changeSubscription;
+
+        /// <inheritdoc />
+        public event EventHandler<WebAppsLogConfiguration> OnConfigurationChanged;
+
+
```
pakrym: nit: space

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            if (string.IsNullOrEmpty(webAppsConfiguration.FileLoggingFolder))
+            {
+                throw new ArgumentNullException(nameof(webAppsConfiguration.FileLoggingFolder), "The file logger path cannot be null");
+            }
+
+            var logsFolder = webAppsConfiguration.FileLoggingFolder;
+            if (!Directory.Exists(logsFolder))
+            {
+                Directory.CreateDirectory(logsFolder);
```
pakrym: Would we end up creating a directory even if logger is disabled?

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
victorhurdugaci: ?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
victorhurdugaci: Because `GetConsumingEnumerable` doesn't block if there are no items

pakrym: I'm quite certain it does.

davidfowl: It does

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+using Serilog.Events;
+
+// TODO: Might want to consider using https://github.com/jezzsantos/Serilog.Sinks.Async 
+// instead of this, once that supports netstandard
+namespace Microsoft.Extensions.Logging.AzureWebApps.Internal
+{
+    /// <summary>
+    /// A background sink for Serilog
+    /// </summary>
+    public class BackgroundSink : ILogEventSink
```
kichalla: Serilog logger has been available for a while and is probably used for web apps frequently, so wondering...do they have not have this kind of support already built in? (like I am a bit surprised if they go to the local disk for every write)

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfigurationReader.cs
```diff
@@ -0,0 +1,132 @@
...
+                _latestConfiguration = WebAppsLogConfiguration.Disabled;
+            }
+            else
+            {
+                _fileLogFolder = Path.Combine(context.HomeFolder, @"LogFiles\Application");
+                var settingsFile = Path.Combine(context.HomeFolder, @"Diagnostics\settings.json");
+
+                _configuration = new ConfigurationBuilder()
+                    .AddEnvironmentVariables()
+                    .AddJsonFile(settingsFile, optional: true, reloadOnChange: true)
```
kichalla: environment input after file input

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfigurationReader.cs
```diff
@@ -0,0 +1,132 @@
...
+        public WebAppsLogConfigurationReader(IWebAppsContext context)
+        {
+            if (!context.IsRunningInAzureWebApp)
+            {
+                _latestConfiguration = WebAppsLogConfiguration.Disabled;
+            }
+            else
+            {
+                _fileLogFolder = Path.Combine(context.HomeFolder, @"LogFiles\Application");
+                var settingsFile = Path.Combine(context.HomeFolder, @"Diagnostics\settings.json");
```
kichalla: Path.Combine

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/IWebAppsContext.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Gets the path to the home folder if running in Azure WebApps
+        /// </summary>
+        string HomeFolder { get; }
+
+        /// <summary>
+        /// Gets a value indicating whether or new we're in an Azure WebApp
+        /// </summary>
+        bool IsRunningInAzureWebApp { get; }
```
kichalla: Why do we want such a restriction? (example if a user wants to use this in their application hosting provider outside azure, they are not supposed to use it?) is it to discourage users from using this logger for non-web apps scenario?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
```diff
@@ -0,0 +1,88 @@
...
+            bool fileLoggingEnabled,
+            LogLevel fileLoggingLevel, 
+            string fileLoggingFolder,
+            bool blobLoggingEnabled, 
+            LogLevel blobLoggingLevel, 
+            string blobContainerUrl)
+        {
+            IsRunningInWebApps = isRunningInWebApps;
+
+            FileLoggingEnabled = fileLoggingEnabled;
```
kichalla: Should these be pluggable instead? for example, since we support structured logging one could log to Azure's DocumentDb (example: PackageSearch does this). Also not sure why we are wrapping around Serilog when an end user can directly use it...will talk offline...

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/FileLoggerProvider.cs
victorhurdugaci: Yes but this code is "public internal" and should not be called directly by the outside. Out "public public" API don't call it

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
victorhurdugaci: See my comment above the namespace

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfigurationReader.cs
victorhurdugaci: Not in this case. We don't want an environment variable to be able to overwrite the config file.

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/IWebAppsContext.cs
victorhurdugaci: This logger is WebApps specific. It will only work in azure because it's tied to the structure and location of the config file that the portal changes

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
victorhurdugaci: That's a different feature. This feature is strictly about logging into the predefined locations that the portal allow you to set. It's not a generic (blob/file/document db/etc) logger.

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfigurationReader.cs
```diff
@@ -0,0 +1,132 @@
...
+            var changeToken = _configuration.GetReloadToken();
+            _changeSubscription = changeToken.RegisterChangeCallback(OnConfigurationTokenChange, null);
+        }
+
+        private void ReloadConfiguration()
+        {
+            // Don't use the binder because of all the defaults that we want in place
+            _latestConfiguration = new WebAppsLogConfiguration(
+                fileLoggingEnabled: TextToBoolean(_configuration.GetSection("AzureDriveEnabled")?.Value),
+                fileLoggingLevel: TextToLogLevel(_configuration.GetSection("AzureDriveTraceLevel")?.Value),
```
kichalla: AzureDriveTraceLevel => AzureDriveLogLevel ? (similarly for AzureBlobTraceLevel)

victorhurdugaci: These come from the portal, we can't change them :)

src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLogger.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
```
cesarbs: Add copyright notice.

src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLoggerProvider.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Microsoft.Extensions.Logging.Abstractions.Internal
```
cesarbs: Add copyright notice.

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsFactoryExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// <summary>
+    /// Extension methods for <see cref="AzureWebAppsLoggerProvider"/>
+    /// </summary>
+    public static class AzureWebAppsFactoryExtensions
+    {
+        /// <summary>
+        /// Adds an Azure Web Apps logger.
+        /// </summary>
+        /// <param name="factory">The extension method argument.</param>
+        public static ILoggerFactory AddAzureWebApps(this ILoggerFactory factory)
```
cesarbs: Why plural?

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                _innerLoggerProvider = NullLoggerProvider.Instance;
+            }
+            else
+            {
+                _innerLoggerProvider = new FileLoggerProvider(_configurationReader);
+
+                if (!string.IsNullOrEmpty(config.BlobContainerUrl))
+                {
+                    // TODO: Add the blob logger by creating a composite inner logger which calls
```
cesarbs: When will this TODO be addressed?

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
cesarbs: I think he means there's no logic here to reload config when it's changed, like the console provider has.

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        /// <summary>
+        /// The default queue size
```
cesarbs: I think the proper style is to end summary sentences with a period.

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+        public void Emit(LogEvent logEvent)
+        {
+            if (!_disposedToken.IsCancellationRequested)
+            {
+                _messages.Add(logEvent);
+            }
+        }
+
+        /// <summary>
+        /// Disposes this object instance
```
cesarbs: end with period

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
cesarbs: `var`

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// A file logger for Azure WebApps
+    /// </summary>
+    public class FileLoggerProvider : WebAppsSerilogLoggerProvider
+    {
+        private const int RetainedFileCountLimit = 31; // Days
+        private const long FileSizeLimitBytes = 1L * 1024 * 1024 * 1024;
+        private const string OutputTemplate = "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{Level}] {Message}{NewLine}{Exception}";
+        private const string FileNamePattern = "log-{Date}.txt";
+
+
```
cesarbs: remove extra blank line

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// <summary>
+        /// Creates a new instance of the <see cref="FileLoggerProvider"/> class
```
cesarbs: end with period

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+            : base(configReader)
+        {
+        }
+
+        /// <inheritdoc />
+        protected override void ConfigureLogger(LoggerConfiguration loggerConfiguration, WebAppsLogConfiguration webAppsConfiguration)
+        {
+            if (string.IsNullOrEmpty(webAppsConfiguration.FileLoggingFolder))
+            {
+                throw new ArgumentNullException(nameof(webAppsConfiguration.FileLoggingFolder), "The file logger path cannot be null");
```
cesarbs: "...cannot be null *or empty*"?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsContext.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        private WebAppsContext() { }
+
+        /// <inheritdoc />
+        public string HomeFolder => Environment.GetEnvironmentVariable("HOME");
+
+        /// <inheritdoc />
+        public bool IsRunningInAzureWebApp =>
+               !string.IsNullOrEmpty(HomeFolder) &&
+               !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("WEBSITE_SITE_NAME"));
```
cesarbs: why 8 space indent?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
```diff
@@ -0,0 +1,88 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.AzureWebApps.Internal
+{
+
```
cesarbs: remove blank line

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// <summary>
+    /// Represents the configuration of the logger from Azure WebApps
+    /// </summary>
+    public class WebAppsLogConfiguration
+    {
+        /// <summary>
+        /// The configuration that has all loggers disabled 
+        /// </summary>
+        public static WebAppsLogConfiguration Disabled => new WebAppsLogConfiguration(false, false, LogLevel.None, null, false, LogLevel.None, null);
```
cesarbs: use named params, one per line

test/Microsoft.Extensions.Logging.AzureWebApps.Test/BackgroundSinkTests.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.AzureWebApps.Test
+{
+    public class BackgroundSinkTests
+    {
```
cesarbs: ?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                _messages = new BlockingCollection<LogEvent>(new ConcurrentQueue<LogEvent>(), maxQueueSize);
+            }
+
+            _innerSink = innerSink;
+
+            _workerThread = new Thread(Worker);
+            _workerThread.Name = GetType().Name;
+            _workerThread.IsBackground = true;
+            _workerThread.Start();
```
cesarbs: Do you really need/want a dedicated thread for this?

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,114 @@
...
+            while (!_disposedToken.IsCancellationRequested)
+            {
+                // This block until either an item is available
+                // or the token is cancelled 
+                LogEvent logEvent = _messages.Take(_disposedToken);
+
+                if (!_disposedToken.IsCancellationRequested &&
+                    logEvent != null)
+                {
+                    PassLogEventToInnerSink(logEvent);
```
kichalla: This sounds like we are flushing the queued log events one at a time? we ideally want to avoid frequent disk access...not sure if I am missing something...

I was thinking we should flush log events when: queue is full or the logger provider is being disposed etc.

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
victorhurdugaci: Correct, the provider is not responsible for reloading the logger. Each internal logger (file, blob, etc) will be responsible for that.

src/Microsoft.Extensions.Logging.AzureWebApps/AzureWebAppsLoggerProvider.cs
victorhurdugaci: @pakrym will add that feature soon

test/Microsoft.Extensions.Logging.AzureWebApps.Test/BackgroundSinkTests.cs
victorhurdugaci: WIP :)

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
victorhurdugaci: We send the events to the rolling file sync. If that one flushes all the time, then we do it to. One of the reasone why we have a background thread doing the work is that

victorhurdugaci:
Updated

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/WebAppsLogConfiguration.cs
ryanerdmann: I'm also curious about the reasoning for using Serilog to implement these loggers.  (Not that it's necessarily a bad thing, but as a likely consumer of this package, other things being equal I'd prefer not to take the transitive dependency).

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -92,9 +92,16 @@ public virtual void Dispose()
...
-            foreach (var logEvent in _messages.GetConsumingEnumerable(_disposedToken))
+            try
             {
-                PassLogEventToInnerSink(logEvent);
+                foreach (var logEvent in _messages.GetConsumingEnumerable(_disposedToken))
+                {
+                    PassLogEventToInnerSink(logEvent);
+                }
+            }
+            catch(OperationCanceledException)
```
cesarbs: nit: space after `catch`

victorhurdugaci:
Updated with some tests

src/Microsoft.Extensions.Logging.AzureWebApps/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        private void Worker()
+        {
+            try
+            {
+                foreach (var logEvent in _messages.GetConsumingEnumerable(_disposedToken))
+                {
+                    PassLogEventToInnerSink(logEvent);
```
kichalla: Ignore any exceptions thrown from here too? so that we can proceed with further log events...

kichalla:
:shipit: from me, but wait for other reviewers too

victorhurdugaci:
Updated (hopefully final form). It has the new names as suggested by @Eilon

src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLogger.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.Abstractions.Internal
+{
+    public sealed class NullLogger : ILogger
```
davidfowl: We shouldn't encourage APIs in the `.Internal` folder referenced by other projects in the same solution. This either needs to be proper public API or shared source

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    /// <summary>
+    /// A background sink for Serilog.
+    /// </summary>
+    public class BackgroundSink : ILogEventSink, IDisposable
+    {
+        /// <summary>
+        /// The default queue size.
+        /// </summary>
+        public const int DefaultQueueSize = 1024;
```
davidfowl: Number of log messages?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,113 @@
...
+                _messages.Add(logEvent);
+            }
+        }
+
+        /// <summary>
+        /// Disposes this object instance.
+        /// </summary>
+        public virtual void Dispose()
+        {
+            if (!_disposedTokenSource.IsCancellationRequested)
```
davidfowl: Get rid of the double check lock and just lock.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,113 @@
...
+        public BackgroundSink(ILogEventSink innerSink, int maxQueueSize)
+        {
+            if (innerSink == null)
+            {
+                throw new ArgumentNullException(nameof(innerSink));
+            }
+
+            _disposedToken = _disposedTokenSource.Token;
+
+            if (maxQueueSize <= 0)
```
davidfowl: Should < 0 blow up?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
davidfowl: Is this just trying to block concurrent calls to dispose?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/IWebAppLogConfigurationReader.cs
```diff
@@ -0,0 +1,24 @@
...
+    {
+        /// <summary>
+        /// Triggers when the configuration has changed.
+        /// </summary>
+        event EventHandler<WebAppLogConfiguration> OnConfigurationChanged;
+
+        /// <summary>
+        /// The latest/current value of the configuration.
+        /// </summary>
+        WebAppLogConfiguration Latest { get; }
```
davidfowl: Current?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
```diff
@@ -0,0 +1,125 @@
...
+            configReader.OnConfigurationChanged += OnConfigurationChanged;
+
+            var loggerConfiguration = new LoggerConfiguration()
+                .MinimumLevel.ControlledBy(_levelSwitch);
+            configureLogger(loggerConfiguration, webAppsConfiguration);
+            var serilogLogger = loggerConfiguration.CreateLogger();
+            
+            OnConfigurationChanged(webAppsConfiguration);
+
+            _loggerFactory = new LoggerFactory();
```
davidfowl: What's the deal with this?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
victorhurdugaci: Yes

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
victorhurdugaci: No. It supports unbounded queues

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
victorhurdugaci: I didn't want to reimplement the Serilog logger logic and the only type they expose is the logger factory extension, not the logger itself. So, in order to create an instance of their logger, I need a logger facotyr

victorhurdugaci:
Updated. I would like to push this before my vacation. If you have more comments, please add them, otherwise sign off

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
davidfowl: There's a thing for this in C# `int?` 😄

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,110 @@
...
+        {
+            lock (_disposedTokenSource)
+            {
+                if (!_disposedTokenSource.IsCancellationRequested)
+                {
+                    _disposedTokenSource.Cancel();
+                }
+
+                // Wait for the thread to complete before disposing the resources
+                _workerThread.Join();
```
davidfowl: You might want to be defensive here and add a timeout.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/WebAppLogConfiguration.cs
```diff
@@ -0,0 +1,87 @@
...
+        /// <summary>
+        /// Creates a new instance of the <see cref="WebAppLogConfiguration"/> class.
+        /// </summary>
+        /// <param name="fileLoggingEnabled">True if file logging is enabled, false otherwise</param>
+        /// <param name="fileLoggingLevel">The logging level for the file logger</param>
+        /// <param name="fileLoggingFolder">The folder where the file logger stores the logs</param>
+        /// <param name="blobLoggingEnabled">True if blob logging is enabled, false otherwise</param>
+        /// <param name="blobLoggingLevel">The logging level for the blob logger</param>
+        /// <param name="blobContainerUrl">The SAS endpoint where blob logs are stored</param>
+        public WebAppLogConfiguration(
```
davidfowl: We don;t use default parameters in any public APIs because of versioning issues. Change this to a builder or just an options object with properties.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
victorhurdugaci: Opened https://github.com/serilog/serilog-extensions-logging/issues/45 to track this

victorhurdugaci:
Updated

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsFactoryExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// Extension methods for <see cref="AzureWebAppDiagnosticsLoggerProvider"/>.
+    /// </summary>
+    public static class AzureWebAppDiagnosticsFactoryExtensions
+    {
+        /// <summary>
+        /// Adds an Azure Web Apps diagnostics logger.
+        /// </summary>
+        /// <param name="factory">The extension method argument</param>
+        /// <param name="fileSizeLimitMb">A strictly positive value representing the maximum log size in megabytes. Once the log is full, no more message will be appended</param>
+        public static ILoggerFactory AddAzureWebAppDiagnostics(this ILoggerFactory factory, int fileSizeLimitMb = FileLoggerProvider.DefaultFileSizeLimitMb)
```
victorhurdugaci: The code configuration will be addressed in https://github.com/aspnet/Logging/issues/470

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,110 @@
...
+        private readonly Thread _workerThread;
+
+        private ILogEventSink _innerSink;
+
+        /// <summary>
+        /// Creates a new instance of the <see cref="BackgroundSink"/> class.
+        /// </summary>
+        /// <param name="innerSink">The inner sink which does the actual logging</param>
+        /// <param name="maxQueueSize">The maximum size of the background queue</param>
+        public BackgroundSink(ILogEventSink innerSink, int maxQueueSize)
```
davidfowl: make maxQueueSize nullable

davidfowl:
@nblumhardt can you review?

src/Microsoft.Extensions.Logging.Abstractions/Internal/NullLogger.cs
davidfowl: @victorhurdugaci ?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/AzureWebAppDiagnosticsLoggerProvider.cs
```diff
@@ -0,0 +1,58 @@
...
+                _innerLoggerProvider = NullLoggerProvider.Instance;
+            }
+            else
+            {
+                _innerLoggerProvider = new FileLoggerProvider(_configurationReader, fileSizeLimitMb);
+
+                if (!string.IsNullOrEmpty(config.BlobContainerUrl))
+                {
+                    // TODO: Add the blob logger by creating a composite inner logger which calls
+                    // both loggers
```
nblumhardt: Might be good to reference another ticket somewhere for the TODO.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
```diff
@@ -0,0 +1,110 @@
...
+using Serilog.Events;
+
+// TODO: Might want to consider using https://github.com/jezzsantos/Serilog.Sinks.Async 
+// instead of this, once that supports netstandard
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+    /// <summary>
+    /// A background sink for Serilog.
+    /// </summary>
+    public class BackgroundSink : ILogEventSink, IDisposable
```
nblumhardt: I just spiked this up a few days ago as well: https://github.com/jezzsantos/Serilog.Sinks.Async/blob/master/src/Serilog.Sinks.Async/AsyncWorkerSink.cs - given there's a high probability of either Serilog.Sinks.Async, or eventually Serilog itself, shipping a type equivalent to this, it may be better `internal` at this stage?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+using Serilog;
+using Serilog.Formatting.Display;
+using Serilog.Sinks.RollingFile;
+
+namespace Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal
+{
+    /// <summary>
+    /// A file logger for Azure WebApp.
+    /// </summary>
+    public class FileLoggerProvider : SerilogLoggerProvider
```
nblumhardt: Minor thing, but instead of subclassing `SerilogLoggerProvider`, using an instance variable and delegating to it will keep `SLP` off of the public API surface of this type (in case a different implementation is desired in future - changing base classes is a breaking change).

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
```diff
@@ -0,0 +1,72 @@
...
+                if (!Directory.Exists(logsFolder))
+                {
+                    Directory.CreateDirectory(logsFolder);
+                }
+                var logsFilePattern = Path.Combine(logsFolder, FileNamePattern);
+
+                var fileSizeLimitBytes = fileSizeLimit * 1024 * 1024;
+
+                var messageFormatter = new MessageTemplateTextFormatter(OutputTemplate, null);
+                var rollingFileSink = new RollingFileSink(logsFilePattern, messageFormatter, fileSizeLimitBytes, RetainedFileCountLimit);
```
nblumhardt: `RollingFileSink` will handle app pool recycling reasonably, but won't coordinate changes by multiple processes (it opens files write-exclusive). Just a quick sanity check - is the hosting environment targeted by this going to spin up multiple processes that would share the same log path/filename?

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/SerilogLoggerProvider.cs
nblumhardt: Easy change for us, unless there's a catch I haven't spotted we should be able to make the `SerilogLogger` public in a point release in a few days.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
victorhurdugaci: Good point but in WebApps you only have one application running at a time. So, unless you add the logger multiple times, there shouldn't be an issue. Outside of WebApps, this logger is disabled.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/BackgroundSink.cs
victorhurdugaci: We have this convention that anything in the `Internal` namespace can be public but it's still internal (so we can change it). We do that to avoid InternalsVisibleTo and still be able to test internal code

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/FileLoggerProvider.cs
victorhurdugaci: I like that :)

nblumhardt:
Thanks for the loop-in guys. I had a quick run through it (sunny Saturday morning here, so on my way out :-)) and everything looks fine. Added what comments I could.

One thing I'd give serious thought to - when the buffer in `BackgroundSink` is full, dropping events in `Emit()` rather than blocking might be a better strategy. I know the idea of blocking is to provide some back-pressure, but when work is generated externally to the app this won't result in the incoming requests being throttled - it'll more likely eat RAM quickly as threads back up blocked on the `Add()` call.

Given events are already going to be dropped at shut-down if they can't be flushed within 5 seconds, dropping them proactively when the buffer fills up seems reasonable. It may not always make a difference, but seems like it might help save an app that's under heavy load (or faulty I/O) once in a while.

I had a shot at implementing this strategy [here](https://github.com/jezzsantos/Serilog.Sinks.Async/blob/master/src/Serilog.Sinks.Async/AsyncWorkerSink.cs#L34). Let me know if I'm off the mark.

Have a great weekend,
Nick

victorhurdugaci:
@nblumhardt thanks for talking a look!

victorhurdugaci:
Merging this as it is - it works. Opened a few bugs that we'll address but I don't have time to do it now. I can do it when I return form vacation.

Here are the issues:

- https://github.com/aspnet/Logging/issues/472
- https://github.com/aspnet/Logging/issues/471
- https://github.com/aspnet/Logging/issues/470

cc @muratg

Issue #473, opened 2016-07-30
Move NullLogger out of the .Internal namespace


davidfowl:
It's public API, not internal API.

/cc @victorhurdugaci

cesarbs:
Should it have it's own package? Seems weird to have it in Logging.Abstractions.

pranavkm:
We have precedent for putting the `Null*` type in the Abstractions package. See https://github.com/aspnet/FileSystem/blob/dev/src/Microsoft.Extensions.FileProviders.Abstractions/NullFileProvider.cs

BrennanConroy:
https://github.com/aspnet/Logging/commit/2affd93f72139efbfd8b6f359e357f371903688a

PR #478, opened 2016-08-22
Move NullLogger out of Internal

BrennanConroy:
#473 
Should NullScope also be moved?

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs
@@ -2,9 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Extensions.Logging.Abstractions.Internal;
 
-namespace Microsoft.Extensions.Logging.Abstractions.Internal
+namespace Microsoft.Extensions.Logging.Abstractions
 {
+    /// <summary>
+    /// Minimalistic logger that does nothing.
+    /// </summary>
     public class NullLogger : ILogger
     {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs
```diff
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.Extensions.Logging.Abstractions.Internal;
 
-namespace Microsoft.Extensions.Logging.Abstractions.Internal
+namespace Microsoft.Extensions.Logging.Abstractions
 {
     public class NullLogger : ILogger
```
pranavkm: Docs since they're now public types.

pranavkm:
:watch: for the docs.

pranavkm:
> Should NullScope also be moved? 

Doesn't appear that it's publicly exposed. Plus it's more of a `NullDisposable` rather than a scope with the NullLogger, so leaving it internal seems ok to me..

BrennanConroy:
Added some docs

src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// </summary>
+    public class NullLogger : ILogger
+    {
+        public static NullLogger Instance { get; } = new NullLogger();
+
+        private NullLogger()
+        {
+        }
+
+        /// <summary>
```
pranavkm: You could use `<inheritdoc />` for implementations. Should suffice in this case.

pranavkm:
:shipit:

PR #479, opened 2016-08-25
LoggerFactory public methods throw ObjectDisposedException after dispose

### LINKED ISSUE #418: 'AddProvider' on LoggerFactory should throw ObjectDisposedException after being disposed
Simple scenario:

    loggerFactory.Dispose();
    loggerFactory.AddProvider(new LoggerProviderWithUnmanagedResources());
    loggerFactory.Dispose();

The `Dispose()` on each `ILoggerProvider` is never being called a second time because the private member `_disposed` in `LoggerFactory` prevents a second call. An `ObjectDisposedException` is expected here.

Worth considering?

natemcmaster:
Resolve #418 

cc @muratg @BrennanConroy

### DIFF
--- a/src/Microsoft.Extensions.Logging/LoggerFactory.cs
+++ b/src/Microsoft.Extensions.Logging/LoggerFactory.cs
@@ -15,10 +15,15 @@ namespace Microsoft.Extensions.Logging
         private readonly Dictionary<string, Logger> _loggers = new Dictionary<string, Logger>(StringComparer.Ordinal);
         private ILoggerProvider[] _providers = new ILoggerProvider[0];
         private readonly object _sync = new object();
-        private bool _disposed = false;
+        private volatile bool _disposed;
 
         public ILogger CreateLogger(string categoryName)
         {
+            if (CheckDisposed())
+            {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -19,6 +19,11 @@ public class LoggerFactory : ILoggerFactory
 
         public ILogger CreateLogger(string categoryName)
         {
+            if (_disposed)
```
BrennanConroy: Extract this out into a `CheckDisposed` method?

BrennanConroy:
Should we throw if you try to dispose a second time also?

natemcmaster:
No, the second dispose should just noop.

src/Microsoft.Extensions.Logging/LoggerFactory.cs
natemcmaster: Good thinking. I'll add.

BrennanConroy:
:shipit:

muratg:
:shipit:

Issue #422, opened 2016-05-09
Loggers throw on null message or arguments


cesarbs:
Part of https://github.com/aspnet/Release/issues/31

`Log*(null)` or `Log*("{0}", null)` makes the logger throw:

    Unhandled Exception: System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.NullReferenceException: Object reference not set to an instance of an object.
       at Microsoft.Extensions.Logging.Internal.FormattedLogValues..ctor(String format, Object[] values)
       at Microsoft.Extensions.Logging.LoggerExtensions.LogTrace(ILogger logger, String message, Object[] args)
       at WebApplicationLoggingNetFramework.Startup.Configure(IApplicationBuilder app, ILoggerFactory loggerFactory) in C:\Users\cesars\Documents\Visual Studio 2015\Projects\WebApplicationLoggingNetFramework\src\WebApplicationLoggingNetFramework\Startup.cs:line 32
       --- End of inner exception stack trace ---
       at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
       at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
       at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
       at Microsoft.AspNetCore.Hosting.Startup.ConfigureBuilder.Invoke(Object instance, IApplicationBuilder builder)
       at Microsoft.AspNetCore.Hosting.Internal.WebHost.BuildApplication()
       at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()

Logging should be more resilient to this. Print `(null)` or something similar when a null is logged.

muratg:
I think this is a dup of https://github.com/aspnet/Logging/issues/335

@cesarbs is this implementation specific or does it repro regardless of the provider?

cesarbs:
@muratg Seems to repro regardless of provider. Notice there's no particular provider in the stack trace.

cesarbs:
[Current agreement](https://github.com/aspnet/Logging/issues/335#issuecomment-200669693) is that logging abstractions should not swallow logger exceptions.

We should still harden our loggers (Console, Debug, etc.) against `null`s.

PR #480, opened 2016-08-25
Harden log formatters to handle null formats, null arguments, and null arrays of arguments

natemcmaster:
Resolves https://github.com/aspnet/Logging/issues/422

cc @muratg @BrennanConroy

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
@@ -15,11 +15,10 @@ namespace Microsoft.Extensions.Logging.Internal
     /// </summary>
     public class LogValuesFormatter
     {
+        private const string NullValue = "(null)";
+        private static readonly object[] EmptyArray = new object[0];
         private readonly string _format;
         private readonly List<string> _valueNames = new List<string>();
-
-        public string OriginalFormat { get; private set; }
-        public List<string> ValueNames => _valueNames;
 
         public LogValuesFormatter(string format)
         {
@@ -59,6 +58,9 @@ namespace Microsoft.Extensions.Logging.Internal
 
             _format = sb.ToString();
         }
+
+        public string OriginalFormat { get; private set; }
+        public List<string> ValueNames => _valueNames;
 
         private static int FindBraceIndex(string format, char brace, int startIndex, int endIndex)
         {
@@ -124,6 +126,7 @@ namespace Microsoft.Extensions.Logging.Internal
 
                     if (value == null)
                     {
+                        values[i] = NullValue;
                         continue;
                     }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -114,6 +115,8 @@ private static int FindIndexOf(string format, char ch, int startIndex, int endIn
             return findIndex == -1 ? endIndex : findIndex;
         }
 
+        private static readonly object[] EmptyArray = new object[0];
```
pranavkm: Move this to the top of the files with the other fields.

pranavkm:
looks good to me. Have someone more familiar with this code sign off on this though.

muratg:
cc @victorhurdugaci

victorhurdugaci:
:shipit:

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -20,6 +20,7 @@ public class LogValuesFormatter
 
         public string OriginalFormat { get; private set; }
         public List<string> ValueNames => _valueNames;
+        private const string NullValue = "(null)";
```
cesarbs: nit: move to top of class, or below existing private members.

cesarbs: While you're at it, you could move `OriginalFormat` and `ValueNames` after the ctor

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -54,16 +54,27 @@ public ConsoleLoggerTest()
...
+            // Act
+            logger.LogCritical(eventId: 0, exception: null, message: null);
+            logger.LogCritical(eventId: 0, message: null);
+            logger.LogCritical(eventId: 0, message: null, exception: exception);
+
+            // Assert
+            Assert.Equal(expectedMessage, sink.Writes[2].Message);
+            Assert.Equal(expectedMessage, sink.Writes[5].Message);
+            Assert.Equal(expectedMessage, sink.Writes[8].Message);
+            Assert.Equal(expectedExceptionMessage, sink.Writes[9].Message);
```
cesarbs: Magic numbers 😁  Isn't there a less-potentially-flaky way to capture those messages?

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
moozzyk: Why not just `Array.Empty<object>()` ?

davidfowl: low netstandard version

moozzyk: I see.

natemcmaster: btw, is the `s_` variable name prefix for static/const privates `s_varName` a convention we do on this team?

PR #482, opened 2016-08-26
Faster Console Logger Mark 1

BrennanConroy:
@davidfowl @glennc @halter73 @rynowak 

~35% increase with an app hosted in azure.

Ignore test failures for now.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -23,6 +23,9 @@ namespace Microsoft.Extensions.Logging.Console
         private IConsole _console;
         private Func<string, LogLevel, bool> _filter;
 
+        [ThreadStatic]
+        private static StringBuilder _logBuilder;
+
         static ConsoleLogger()
         {
             var logLevelString = GetLogLevelString(LogLevel.Information);
@@ -105,12 +108,16 @@ namespace Microsoft.Extensions.Logging.Console
 
         public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message, Exception exception)
         {
+            var logBuilder = _logBuilder;
+            _logBuilder = null;
+
+            if (logBuilder == null)
+            {
+                logBuilder = new StringBuilder();
+            }
+
             var logLevelColors = default(ConsoleColors);
             var logLevelString = string.Empty;
-            var logIdentifier = string.Empty;
-            var scopeInformation = string.Empty;
-            var exceptionText = string.Empty;
-            var printLog = false;
 
             // Example:
             // INFO: ConsoleApp.Program[10]
@@ -120,15 +127,21 @@ namespace Microsoft.Extensions.Logging.Console
                 logLevelColors = GetLogLevelConsoleColors(logLevel);
                 logLevelString = GetLogLevelString(logLevel);
                 // category and event id
-                logIdentifier = _loglevelPadding + logName + "[" + eventId + "]";
+                logBuilder.Append(_loglevelPadding);
+                logBuilder.Append(logName);
+                logBuilder.Append("[");
+                logBuilder.Append(eventId);
+                logBuilder.AppendLine("]");
                 // scope information
                 if (IncludeScopes)
                 {
-                    scopeInformation = GetScopeInformation();
+                    GetScopeInformation(logBuilder);
                 }
                 // message
-                message = _messagePadding + ReplaceMessageNewLinesWithPadding(message);
-                printLog = true;
+                logBuilder.Append(_messagePadding);
+                var len = logBuilder.Length;
+                logBuilder.AppendLine(message);
+                logBuilder.Replace(Environment.NewLine, _newLineWithMessagePadding, len, message.Length);
             }
 
             // Example:
@@ -137,12 +150,12 @@ namespace Microsoft.Extensions.Logging.Console
             if (exception != null)
             {
                 // exception message
-                exceptionText = exception.ToString();
-                printLog = true;
+                logBuilder.AppendLine(exception.ToString());
             }
 
-            if (printLog)
+            if (logBuilder.Length > 0)
             {
+                var logMessage = logBuilder.ToString();
                 lock (_lock)
                 {
                     if (!string.IsNullOrEmpty(logLevelString))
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -120,14 +119,14 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
                 logLevelColors = GetLogLevelConsoleColors(logLevel);
                 logLevelString = GetLogLevelString(logLevel);
                 // category and event id
-                logIdentifier = _loglevelPadding + logName + "[" + eventId + "]";
+                sb.AppendLine(_loglevelPadding + logName + "[" + eventId + "]");
```
victorhurdugaci: Would this be even faster if you would `sb.Append` all the time instead of string concat?

BrennanConroy: If it was you would only save a couple milliseconds every few hundred thousand calls to log because most of the time is spent on the lock so any major perf gains will be gained from reducing time spent inside the lock.

moozzyk: Not sure how fast/slow it is but stringBuilder has the AppendFormat method....

rynowak: Don't use `AppendFormat` where concat will do, it results in running a lot more code.

The issue here is that `string.Concat` allocates, and allocating anything can result in taking a lock if the per-thread heap is exhausted.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -120,14 +119,14 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
                 // scope information
                 if (IncludeScopes)
                 {
-                    scopeInformation = GetScopeInformation();
+                    sb.AppendLine(GetScopeInformation());
```
rynowak: This is another opportunity to avoid allocations by passing in the builder - notice that scopes already use a stringbuilder inside the method, we can eliminate that.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -120,14 +119,14 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
                 }
                 // message
-                message = _messagePadding + ReplaceMessageNewLinesWithPadding(message);
+                sb.AppendLine(_messagePadding + ReplaceMessageNewLinesWithPadding(message));
```
rynowak: This is another opportunity to remove allocation, since we do a `string.replace` on the message

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -107,11 +107,10 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
         {
             var logLevelColors = default(ConsoleColors);
             var logLevelString = string.Empty;
-            var logIdentifier = string.Empty;
-            var scopeInformation = string.Empty;
-            var exceptionText = string.Empty;
             var printLog = false;
 
+            var sb = new StringBuilder();
```
rynowak: We could consider using a threadlocal to cache these if we need to. Logging is 99.99% likely to be non-reentrant.

BrennanConroy:
Added some changes for memory and went from `748,341` bytes for 1000 logs to `307,173` bytes
Will test RPS after lunch

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -105,11 +109,13 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
 
         public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message, Exception exception)
         {
+            if (_sb == null)
```
rynowak: You should cache this in a local, and document that this is safe while we're non-reentrant.

You should also probably go to the trouble of doing acquire/release in case someone subclasses this and does something crazy.

```
var sb = _sb;
_sb = null;

if (sb == null)
{
  _sb == new StringBuilder();
}

.... do logging stuff ...

_sb = sb;
```

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -120,14 +126,19 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
                 // scope information
                 if (IncludeScopes)
                 {
-                    scopeInformation = GetScopeInformation();
+                    _sb.AppendLine(GetScopeInformation());
                 }
                 // message
-                message = _messagePadding + ReplaceMessageNewLinesWithPadding(message);
+                _sb.Append(_messagePadding);
+                _sb.AppendLine(ReplaceMessageNewLinesWithPadding(message));
```
rynowak: https://msdn.microsoft.com/en-us/library/y1bxd041(v=vs.110).aspx

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -120,14 +126,19 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
+                _sb.Append(_loglevelPadding);
+                _sb.Append(logName);
+                _sb.Append("[");
+                _sb.Append(eventId);
+                _sb.AppendLine("]");
                 // scope information
                 if (IncludeScopes)
                 {
-                    scopeInformation = GetScopeInformation();
+                    _sb.AppendLine(GetScopeInformation());
```
rynowak: `GetScopeInformation` should be passed the current string builder rather creating a new one

BrennanConroy:
Perf decreased ~2% after the memory changes

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -255,14 +249,14 @@ private ConsoleColors GetLogLevelConsoleColors(LogLevel logLevel)
             }
         }
 
-        private string GetScopeInformation()
+        private void GetScopeInformation(StringBuilder builder)
```
kichalla: Not sure how much this would save, but since you are measuring perf, you could probably optimize here to avoid calculating the scope if its already calculated. (for reference...Long time back I had a PR for this change https://github.com/aspnet/Logging/pull/404/files)

BrennanConroy:
🆙 📅

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -255,14 +249,14 @@ private ConsoleColors GetLogLevelConsoleColors(LogLevel logLevel)
...
         {
             var current = ConsoleLogScope.Current;
-            var output = new StringBuilder();
             string scopeLog = string.Empty;
+            var length = builder.Length;
             while (current != null)
             {
-                if (output.Length == 0)
+                if (length == builder.Length)
```
victorhurdugaci: No need to have this in the while loop since you execute it only on the first iteration. It could be outside

BrennanConroy: Could be, but the code wouldn't look very good.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -137,12 +150,12 @@ public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId,
...
                 // exception message
-                exceptionText = exception.ToString();
-                printLog = true;
+                sb.AppendLine(exception.ToString());
             }
 
-            if (printLog)
+            if (sb.Length > 0)
             {
+                var msg = sb.ToString();
```
rynowak: give this a better name. `message` is already used, and this isn't semantically the same as `message`. This is a pretty good example of why we don't abbreviate :laughing:

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -105,12 +108,16 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
 
         public virtual void WriteMessage(LogLevel logLevel, string logName, int eventId, string message, Exception exception)
         {
+            var sb = _sb;
```
rynowak: don't abbreviate

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -255,14 +243,15 @@ private ConsoleColors GetLogLevelConsoleColors(LogLevel logLevel)
             }
         }
 
-        private string GetScopeInformation()
+        private void GetScopeInformation(StringBuilder builder)
         {
             var current = ConsoleLogScope.Current;
-            var output = new StringBuilder();
             string scopeLog = string.Empty;
+            var length = builder.Length;
```
rynowak: Since the tests also changed, please do a manual verification with scopes on that this behaves the same way

BrennanConroy: Tested with 360 no scope, single scope, and multi-scope, looks the same

rynowak:
:shipit:

PR #485, opened 2016-09-01
Fix blob sink tests

pakrym:
@victorhurdugaci

### DIFF
--- a/test/Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Test/AzureBlobSinkTests.cs
+++ b/test/Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Test/AzureBlobSinkTests.cs
@@ -7,14 +7,12 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging.AzureWebAppDiagnostics.Internal;
 using Microsoft.WindowsAzure.Storage;
 using Moq;
-using Serilog;
-using Serilog.Core;
 using Serilog.Events;
+using Serilog.Formatting;
 using Serilog.Formatting.Display;
 using Serilog.Parsing;
 using Xunit;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/AzureBlobSink.cs
```diff
@@ -70,6 +70,13 @@ public class AzureBlobSink : PeriodicBatchingSink
         /// <inheritdoc />
         protected override async Task EmitBatchAsync(IEnumerable<LogEvent> events)
         {
+            await EmitBatchInternalAsync(events);
+        }
+
+
+        // For unit testing
+        internal async Task EmitBatchInternalAsync(IEnumerable<LogEvent> events)
```
moozzyk: you could make EmitBatchAsync `protected internal`. Potentially you could avoid InternalsVisibleTo if you derived from AzureBlobSink in your tests and created a public EmitBatchAsyncPublich method that forwards to EmitBatchAsync...

victorhurdugaci:
What was the problem? Please add a few words when sending a PR

pakrym:
@victorhurdugaci problem was in timer inside `PeriodicBatchingSink` firing randomly producing inconsistent batches.

src/Microsoft.Extensions.Logging.AzureWebAppDiagnostics/Internal/AzureBlobSink.cs
```diff
@@ -70,6 +70,13 @@ public class AzureBlobSink : PeriodicBatchingSink
         /// <inheritdoc />
         protected override async Task EmitBatchAsync(IEnumerable<LogEvent> events)
         {
+            await EmitBatchInternalAsync(events);
+        }
+
```
cesarbs: nit: remove extra blank line

victorhurdugaci:
> @victorhurdugaci problem was in timer inside PeriodicBatchingSink firing randomly producing inconsistent batches.

Will this be an issue in production?

pakrym:
@victorhurdugaci we don't really care how exact batches would look like, only that most of them would have >1 message. Yes, no longer testing a base class.

pakrym:
:up::date:

PR #489, opened 2016-09-08
Implement add and subtract operators for EventId's

henkmollema:
- Implement `+` operator
- Implement `-` operator
- Added unit tests
- Use C# 6 read-only properties
- Use null coalescing in `ToString()` method

This allows for a event class like this:

``` csharp
public static class Events
{
    public static EventId Base = new EventId(100);
    public static EventId SomeEvent = Base + 1;
    public static EventId AnotherEvent = Base + 2;
    public static EventId MoreEvent = Base + 3;
}
```

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
@@ -5,29 +5,24 @@ namespace Microsoft.Extensions.Logging
 {
     public struct EventId
     {
-        private int _id;
-        private string _name;
-
         public EventId(int id, string name = null)
         {
-            _id = id;
-            _name = name;
+            Id = id;
+            Name = name;
         }
 
-        public int Id
+        public int Id { get; }
+
+        public string Name { get; }
+
+        public static EventId operator +(EventId left, EventId right)
         {
-            get
-            {
-                return _id;
-            }
+            return new EventId(left.Id + right.Id);
         }
 
-        public string Name
+        public static EventId operator -(EventId left, EventId right)
         {
-            get
-            {
-                return _name;
-            }
+            return new EventId(left.Id - right.Id);
         }
 
         public static implicit operator EventId(int i)
@@ -37,14 +32,7 @@ namespace Microsoft.Extensions.Logging
 
         public override string ToString()
         {
-            if (_name != null)
-            {
-                return _name;
-            }
-            else
-            {
-                return _id.ToString();
-            }
+            return Name ?? Id.ToString();
         }
     }
 }
--- a/test/Microsoft.Extensions.Logging.Test/EventIdTests.cs
+++ b/test/Microsoft.Extensions.Logging.Test/EventIdTests.cs
@@ -1,0 +1,67 @@
+﻿using Xunit;
+
+namespace Microsoft.Extensions.Logging.Test
+{
+    public class EventIdTests
+    {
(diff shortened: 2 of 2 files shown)


304NotModified:
In which case you like to do adding/subtracting of an identifier?

henkmollema:
@304NotModified see my example:
```csharp
public static class Events
{
    public static EventId Base = new EventId(100);
    public static EventId SomeEvent = Base + 1;
    public static EventId AnotherEvent = Base + 2;
    public static EventId MoreEvent = Base + 3;
}
```

Where
```
Base + 1
```

resolves in `new EventId(101)`.

304NotModified:
I was more looking for an use case. Why in earth should people do Id+1.

henkmollema:
Because I wan't to define a base event id for a component which is a starting point and start add a number for each event within that component.

304NotModified:
And what will happen of the id will have a collision with another component? And the id of the event could change every time?

I think IDs should be fixed, not dynamic.

henkmollema:
I don't see what that has to with this pull request. This is just about enabling some syntax, not new functionality.

304NotModified:
Well this PR makes it easy to change an id, which is strange as ids sound be fixed IMO - An id should identify something and its strange that this it could change. 

I don't see an useful use case to do this, and because this is also possible without the implicit conversions, I'm questioning this PR

henkmollema:
I agree that ID's should not change. But this is not about _changing_ ID's, it's about _defining_ them.

In the end I just want to be able to do:
```
public static PersonApiEvents
{
   public static EventId Base = new EventId(100);
   public static EventId PersonGetFailed = Base + 1;
   public static EventId PersonPostFailed = Base + 2;
   public static EventId PersonPutFailed = Base + 3;
   public static EventId PersonDeleteFailed = Base + 4;
}
```

Rather than:
```
public static PersonApiEvents
{
   public static int Base = 100;
   public static int PersonGetFailed = Base + 1;
   public static int PersonPostFailed = Base + 2;
   public static int PersonPutFailed = Base + 3;
   public static int PersonDeleteFailed = Base + 4;
}
```

It's a similar way of how [`WebEventCodes`](https://msdn.microsoft.com/en-us/library/system.web.management.webeventcodes(v=vs.110).aspx) are implemented in ASP.NET Health Monitoring.

304NotModified:
It a least makes it easier to change, which I think we should not encourage.

sgryphon:
The semantics of adding/subtracting don't make sense for EventIds. I agree with the other comments that it allows them to be changed to easily, when ideally they should be static. 

While it might work for the numeric parts, something like the following doesn't make sense: new EventId(1, "a") + new EventId(2, "b")

I understand the desire to have numbers derived from a base value (and have written about structured event IDs before: https://essentialdiagnostics.codeplex.com/wikipage?title=Event%20Ids&referringTitle=Guidance), but don't think math operators are the correct solution.

Note that there is an implicit cast from int to EventId, so you can just do the following, which is also less typing (it also facilitates if you have an error log with the number and do a text search for "103", then the corresponding definition can be easily found):

public static PersonApiEvents
{
   public static EventId PersonGetFailed = 101;
   public static EventId PersonPostFailed = 102;
   public static EventId PersonPutFailed = 103;
   public static EventId PersonDeleteFailed = 104;
}

henkmollema:
I guess I'll leave them be then.

PR #492, opened 2016-09-25
Remove Abstractions Linq dependency

benaadams:
Don't need to reference Linq for a tiny function and its a 125Kb dependency 

(As well as casting through object, to a `IEnumerator<object>`, to then Select for the `null`s back to `IEnumerator<object>` to then to `Join(String separator, IEnumerable<object> values)` rather than `Join(String separator, IEnumerable<String> values)` not being the greatest code path)

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using System.Globalization;
 using System.Text;
 
@@ -140,12 +139,20 @@ namespace Microsoft.Extensions.Logging.Internal
                     var enumerable = value as IEnumerable;
                     if (enumerable != null)
                     {
-                        values[i] = string.Join(", ", enumerable.Cast<object>().Select(o => o ?? NullValue));
+                        values[i] = string.Join(", ", CastIteratorToString(enumerable));
                     }
                 }
             }
 
             return string.Format(CultureInfo.InvariantCulture, _format, values ?? EmptyArray);
+        }
+
+        private static IEnumerable<string> CastIteratorToString(IEnumerable source)
+        {
+            foreach (var obj in source)
+            {
+                yield return obj?.ToString() ?? NullValue;
+            }
         }
 
         public KeyValuePair<string, object> GetValue(object[] values, int index)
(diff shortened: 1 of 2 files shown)


davidfowl:
LGTM, there is a question about dropping deps and how that relates to breaking changes but this looks fine /cc @Eilon

benaadams:
There are extra deps brought in for a single error message resource string; and some which aren't used but are used by sub packages:

Also `CultureInfo.InvariantCulture` => `(IFormatter)null`

Which would mean the following additional deps could be removed (also brought in by `ResourceManager`)

```
"System.Diagnostics.Debug": "4.0.11-*"
"System.Globalization": "4.0.11-*"
"System.Reflection": "4.1.0-*"
"System.Resources.ResourceManager": "4.0.1-*"
"System.Runtime.Extensions": "4.1.0-*"
```
But that is a behavioural change and they mostly exist in mscorlib anyway so don't think that would gain much 😞

Eilon:
We're going to add standard library references to everything anyway, so even if the refs are temporarily dropped from this project.json, they'll be effectively added back once we add standard lib.

Eilon:
The bigger question is just whether this change is valuable. And considering that as I mentioned the dependency won't *actually* end up getting dropped, LINQ will be available no matter what. And almost all apps reference LINQ eventually anyway, so most apps won't be getting any savings in practice.

benaadams:
> We're going to add standard library references to everything anyway, 

Whoa! Really? So if I'm building a standalone app its xcopy size has just jumped?

> And almost all apps reference LINQ eventually anyway,

I'd say under 50% of nuget libraries use Linq https://apisof.net/catalog/System.Linq and under 8% use XmlDoc which is another that the standard library brings in https://apisof.net/catalog/System.Xml.XmlDocument

benaadams:
Though Linq and Xml aren't the end of the world dependency-wise; however the code change is still valid as an improvement.

Eilon:
@benaadams there's a "dependency trimming" feature that is being worked on that will significantly help mitigate the "extraneous dependencies" problem.

Also, my comment was that *apps* end up using LINQ, not libraries. Ultimately apps are the unit of publishing, so it's only their dependencies that matter for that.

herecydev:
@Eilon Do you have a link for that issue?

Eilon:
@herecydev sorry, I don't. It's something I overheard that a team was working on. I was spying on them 😄 

Perhaps @davidfowl knows more about it?

davidfowl:
Closing this PR as we've pulled in netstandard.library

PR #494, opened 2016-09-26
Add Microsoft.Extensions.Logging.AzureWebAppDiagnostics to NuGetPackageVerifier.json

pakrym:
@Tratcher


Tratcher:
:shipit:

muratg:
Per @pranavkm, this is not required, as standard checks apply to all packages. Only if you need to have some exceptions or something, you can ignore this file when you have a new package.

I recommend closing this PR without merging.

Tratcher:
When did this change? And why do we still get CI warnings about missing packages?

muratg:
Not sure since when this is just a warning and not an error. @pranavkm is thinking about removing the warning.

PR #501, opened 2016-10-14
Accept existing ILoggerFactory in AddLogging()

cwe1ss:
Programs often initialize their own logger factory on startup before DI is initialized to get logging as soon as possible. To add the factory to DI, we have to manually call `AddSingleton(loggerFactory)`. 

However it's easy to get this wrong - the following code will not work properly because this will add the instance for the concrete type `LoggerFactory` instead of for the interface `ILoggerFactory`. This means, the existing factory will not be used, but instead AddLogging() will register a new logger factory instance.

``` csharp
services.AddSingleton(loggerFactory);
services.AddLogging();
```

This PR would allow us to pass an existing logger factory to `AddLogging` and therefore make sure that this problem can't occur.

This could then also be used in [aspnet/hosting -> WebHostBuilder](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/WebHostBuilder.cs#L189)

### DIFF
--- a/src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
+++ b/src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
@@ -16,15 +16,24 @@ namespace Microsoft.Extensions.DependencyInjection
         /// Adds logging services to the specified <see cref="IServiceCollection" />.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="loggerFactory">An optional existing logger factory.</param>
         /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
-        public static IServiceCollection AddLogging(this IServiceCollection services)
+        public static IServiceCollection AddLogging(this IServiceCollection services, ILoggerFactory loggerFactory = null)
         {
             if (services == null)
             {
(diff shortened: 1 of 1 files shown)


davidfowl:
> Programs often initialize their own logger factory on startup before DI is initialized to get logging as soon as possible. To add the factory to DI, we have to manually call AddSingleton(loggerFactory).

They do?

cwe1ss:
Yes - e.g. in all non-web projects.

davidfowl:
And the reason you're calling AddLogging is to get the ILogger<T> support?

cwe1ss:
After thinking about it, this is actually only meant for scenarios where you don't use WebHostBuilder. With WebHostBuilder, you would just call `UseLoggerFactory`.

Our console applications (message processors, workers, utils, ...) all use Microsoft.Extensions.Logging, Microsoft.Extensions.DependencyInjection, etc.
Until we get your "uber host builder" we manually have to setup these components. And we always setup Logging as early as possible - which means we do it before we have DI.

This PR would solve a small pain point with the current API in this scenario.

muratg:
Closing this one.

glennc:
@cwe1ss The discussion around a more generic HostBuilder for non-web apps that want to use ILogger, Config, and DI impact this. I think you've been following some of that and we should have something more coherent together soon.

PR #505, opened 2016-10-27
Cleanup EventSource logging provider

pakrym:
1. Change namespace to be the same as project name.
2. Make LoggingEventSource internal.
3. Minor code style fixes.

@pranavkm

### DIFF
--- a/src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
@@ -8,7 +8,7 @@ using System.IO;
 using System.Threading;
 using Newtonsoft.Json;
 
-namespace Microsoft.Extensions.Logging.EventSourceLogger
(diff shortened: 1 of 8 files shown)


BrennanConroy:
🐑 :it:

src/Microsoft.Extensions.Logging.EventSource/EventSourceLogger.cs
```diff
@@ -19,9 +19,9 @@ namespace Microsoft.Extensions.Logging.EventSourceLogger
     /// </remarks>
     internal class EventSourceLogger : ILogger
     {
+        private static int _activityIds;
```
pranavkm: `_activityId`? Also, should this be a non-static. cc @davidfowl for an opinion.

pakrym: It's used to get unique id's for scopes.

PR #512, opened 2016-11-04
Fix more comments and property validation in Microsoft.Extensions.Logging.AzureAppServices

pakrym:
@Eilon

### DIFF
--- a/src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesDiagnosticsSettings.cs
+++ b/src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesDiagnosticsSettings.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Extensions.Logging.AzureAppServices
             get { return _outputTemplate; }
             set
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesDiagnosticsSettings.cs
```diff
@@ -135,7 +135,7 @@ public int BackgroundQueueSize
             {
                 if (value < 0)
                 {
-                    throw new ArgumentOutOfRangeException(nameof(value), $"{nameof(BackgroundQueueSize)} must be positive or 0.");
+                    throw new ArgumentOutOfRangeException(nameof(value), $"{nameof(BackgroundQueueSize)} must be non-negative.");
```
pranavkm: ~non-negative~ positive integer?

pakrym: can be 0

Eilon:
Thanks! 😄 😄 😄

PR #521, opened 2016-11-17
DEBUG and TRACE should have CONDITIONAL attribute

### LINKED ISSUE #520: Add Conditional attribute to LogDebug and LogTrace
**LogDebug** and **LogTrace** should be decorated with `Conditional` attribute and tied with defined DEBUG & TRACE in order to not evaluate expensive objects on production.

#### project.json

```json
{
  "buildOptions": {
    "emitEntryPoint": true,
    "debugType": "portable",
    "define": [ ]
  }
}
```

#### appsettings.json
```json
  "Logging": {
    "IncludeScopes": true,
    "LogLevel": {
      "Default": "Information",
      "System": "Error",
      "Microsoft": "Error"
    }
  }
```

#### AddressRepository.cs 
```csharp
_logger.LogDebug(JsonConvert.SerializeObject(address));
_logger.LogTrace(JsonConvert.SerializeObject(address));
```

### Actual result
Evaluates `JsonConvert.SerializeObject(address)`

### Expected result
It should not evaluate `JsonConvert.SerializeObject(address)` unless DEBUG and/or TRACE is set on **appsettings.json**.

cilerler:
In order to eliminate DEBUG and TRACE evaluations on RELEASE those helper functions should have CONDITIONAL attribute.  Closes #520

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
@@ -13,6 +13,9 @@ namespace Microsoft.Extensions.Logging
     /// </summary>
     public static class LoggerExtensions
     {
+        private const string Debug = "DEBUG";
+        private const string Trace = "TRACE";
+        
         private static readonly Func<object, Exception, string> _messageFormatter = MessageFormatter;
 
         //------------------------------------------DEBUG------------------------------------------//
@@ -25,6 +28,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="exception">The exception to log.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Debug)]
         public static void LogDebug(this ILogger logger, EventId eventId, Exception exception, string message, params object[] args)
         {
             if (logger == null)
@@ -42,6 +46,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="eventId">The event id associated with the log.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Debug)]
         public static void LogDebug(this ILogger logger, EventId eventId, string message, params object[] args)
         {
             if (logger == null)
@@ -58,6 +63,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Debug)]
         public static void LogDebug(this ILogger logger, string message, params object[] args)
         {
             if (logger == null)
@@ -78,6 +84,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="exception">The exception to log.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Trace)]
         public static void LogTrace(this ILogger logger, EventId eventId, Exception exception, string message, params object[] args)
         {
             if (logger == null)
@@ -95,6 +102,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="eventId">The event id associated with the log.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Trace)]
         public static void LogTrace(this ILogger logger, EventId eventId, string message, params object[] args)
         {
             if (logger == null)
@@ -111,6 +119,7 @@ namespace Microsoft.Extensions.Logging
         /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
         /// <param name="message">Format string of the log message.</param>
         /// <param name="args">An object array that contains zero or more objects to format.</param>
+        [Conditional(Trace)]
         public static void LogTrace(this ILogger logger, string message, params object[] args)
         {
             if (logger == null)


cwe1ss:
Logging levels are completely independent of build targets. Even if I deploy with RELEASE, I must be able to log trace/debug messages in troubleshooting scenarios.

cilerler:
I understand your point but this doesn't mean that you shouldn't tie it up with definitions.  It is OK to not use default ones _(DEBUG, TRACE)_ but we can use _LOGDEBUG_, _LOGTRACE etc._ and it will gives us opportunity to remove the diagnostics code through conditional for production environment which will be a great advantage for performance perspective since those codes won't be evaluated.

If you still think this is not the way to do, would you please point me to the right direction how to implement logging and not evaluating the data I'm sending in on production without decorating entire code with `#if` and/or `isEnabled()`'s?

cwe1ss:
Non-default identifiers might be a possibility, yes. This would make it completely opt-in. Not sure if I like it though - it still results in lots of pain because you can't change your logging levels without a redeploy.

If you only need the message parameter you could create delegate extension methods that allow you to do something like `logger.LogDebug(() => JsonConvert.SerializeObject(address))`. If I'm correct, this would only allocate the delegate.

You can also move all logging code to a separate class, like they do in [aspnet/hosting](https://github.com/aspnet/Hosting/blob/dev/src/Microsoft.AspNetCore.Hosting/Internal/HostingLoggerExtensions.cs), but that's more work obviously.

cwe1ss:
Other extension methods - not 100% beautiful though :)
```csharp 
// string ComplexTransform() {}
logger.LogDebug(ComplexTransform);
// with regular args
logger.LogDebug(() => address.ToString(), customerId, orderId);
// with func args
logger.LogDebug(() => address.ToString(), () => new object[] { CalculateSum(), customerId });
// with exception
logger.LogDebug(ex, () => address.ToString());
```

cilerler:
> you can't change your logging levels without a redeploy

I don't agree with that, since it gives the opportunity to choose. By default all definitions should be added which means there will be no difference than what we already have right now.  

But it will gives us opportunity to implement such a powerful mechanism to remove unnecessary diagnostic code with removing single line of code.  I really think we should not pass this opportunity without proper thinking.

Delegation should not solve the issue since it will be called before filtering system.  _(have to check not sure 100%)_ but implementing extension class does serve the purpose.  _well it creates a [new problem](http://stackoverflow.com/questions/40660834/how-to-pass-compilation-directive-into-other-projects-in-dotnet-core)_  but almost there 😄 

I saw what they done in _aspnet/hosting_, as you mention it is more work.  So would you like me to go back and fix the project.json, tests stuff etc or should we pass that.

Should I continue to work on this and update files etc. or you are still not convinced to accept a pull request even everything will work just fine?

cilerler:
it will go to [here](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/Internal/FormattedLogValues.cs#L23) so it will evaluate even if we pass it as delegate.

cwe1ss:
I'm not working for MS - that was just an opinion from a random guy :man:. 

Would it even be possible to include build definitions by default? Or would it require negative definitions - e.g. NO_TRACE, NO_DEBUG, ...?

cilerler:
> "I'm not working for MS - that was just an opinion from a random guy"😂 

Nope, `Conditional` attribute does not support negatives.

cwe1ss:
My examples require new extension methods:
```csharp
public static void LogDebug(this ILogger logger, Func<string> message)
{
  if (logger.IsEnabled(LogLevel.Debug))
  {
    // delegate is only executed if enabled
    logger.LogDebug(message());
  }
}
```

sgryphon:
Part of the problem is that LogLevel.Trace and LogLevel.Debug are defined the opposite way around from how TRACE and DEBUG are usually configured.

For a default Visual Studio project, the Debug build defines DEBUG;TRACE, whereas the Release build only defines TRACE.

Obviously you can change this, but I've always interpreted it as the lowest level diagnostics statements are from System.Diagnostics.Debug.WriteLine(), i.e. DEBUG level, as they only appear (by default) in local Debug builds.

In contrast, System.Diagnostics.Trace.WriteLine(), and System.Diagnostics.TraceSource.TraceEvent(), which has levels of Verbose, Information, then through to Critical, are all enabled at TRACE level and appear (by default) in all builds, including Release builds (but can be turned off).

So, in old .NET, DEBUG turns on/off Debug, and TRACE turns on/off everything else.

To me, it appears the definitions in LogLevel are the wrong way around, which means mapping them to the existing symbols doesn't make sense.

cilerler:
@sgryphon First of all, I agree everything you said.  But focus here should be _"diagnostics logs should not effect the performance of the application.  And we have capability to write a single code and not decorate with `#IF`'s at all."_

Tie them up with DEBUG/TRACE is completely optional as I said above it can be LOGDEBUG/LOGTRACE but whole idea here is writing single code and having option to not compile certain parts without using `#IF` directives

Since feature is _logging_ in this subject and base is _compiling vs not compiling_ definitions are 100% related to the subject.

sgryphon:
@cilerler yes, logging should not affect the performance, although I would add the caveat "diagnostics logs should not affect the performance of the application *when disabled*".

If you want to generate 1 million verbose level messages and capture them in a log file, then the application performance will be affected as it writes to the file; however if you turn logging off (e.g. a config switch, conditional compilation, whatever), then the effect of 1 million verbose level messages that are *not* captured should be minimal.

I have previously done some performance analysis of several frameworks (pre-Core), scroll down to the performance section: https://essentialdiagnostics.codeplex.com/wikipage?title=Comparison&referringTitle=Guidance

In short, most of them have minimal impact when turned off. A part of this is lazy evaluation (more below), but note that the problems with the overhead of EntLib are not due to this (because I also compared with plain strings). In short, I don't think the overhead is something to worry about with a well designed framework and compilation switches should not be necessary.

I would do some comparisons and see just how much conditional compilation, vs turn a config switch off, is going to save you; some optimisations simply aren't worth it. I guess you are at most going to save 50ms per million messages not logged. 

Personally, I would never want to compile with tracing turned off because if you do, and there is a problem, then you are stuck with no way to investigate.


Work arounds:

The problem, as you pointed out, is that _logger.LogDebug(JsonConvert.SerializeObject(address)) in your example is not lazy evaluated. The expensive serialisation is done, even if the message is not actually logged.

A delegate/lambda (used in some frameworks) would solve the problem, i.e.  _logger.LogDebug(() => JsonConvert.SerializeObject(address))

Of course a lambda is just shorthand for a full class/method, and almost all logging frameworks (except EntLib) support lazy evaluation of args passed with a format string, so knowing that the ToString() method will be called we can do this (I wrote this by hand, so probably has syntax errors; treat as pseudocode):

public class LazyStringify
{
	Func<object> _item;
	public LazyStringify(Func<object> item) { _item = item; }
	public override string ToString() { return _item().ToString(); }
}

_logger.LogDebug("{0}", new LazyStringify(() => JsonConvert.SerializeObject(address)))

You can encapsulate this (or a check to IsEnabled()) in an extension method. I would probably be wary of expanding the API surface (e.g. with lambdas).

Where the thing to log is more extensive/requires external data, then a check to IsEnabled is the right way to go.

cilerler:
@sgryphon thank you for the details response and sorry about the delayed one, some how missed it.  😞  I hear you but in my case I'm logging [2GB/hour](http://docs.getseq.net/v3/discuss/5840887ce3e04b230055c703) data and really don't want that overhead in production.

Anyway, I already implemented it as external method and works just fine but I believe it should be in the framework since the problem I have is very common scenario.

PR #527, opened 2016-11-30
Very simple, basic "hello world", example for logging.

sgryphon:
This is a very basic example of logging, so developers can get started. Most logging frameworks (e.g. log4net, serilog, etc) start off with a very basic walkthrough, so this is a good place to start.


davidfowl:
@sgryphon I don't think we want this many sample projects in the logging repository. aspnet/Entropy is the good place to put samples like the ones you have been sending PRs for. Usually the samples in repros are projects used by the developers to do quick testing for things. They are not a showcase for all of features.

muratg:
@sgryphon What @davidfowl said :) Right place for the samples is Entropy repo.

Issue #532, opened 2016-12-13
ConsoleLogger deadlocked


milichev:
I found my POC project running overnight deadlocked with the same callstack on 4 threads were waiting for [_lock](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L159) to be free:

```
>	Microsoft.Extensions.Logging.Console.dll!Microsoft.Extensions.Logging.Console.ConsoleLogger.WriteMessage(Microsoft.Extensions.Logging.LogLevel logLevel, string logName, int eventId, string message, System.Exception exception)
 	Microsoft.Extensions.Logging.Console.dll!Microsoft.Extensions.Logging.Console.ConsoleLogger.Log<System.__Canon>(Microsoft.Extensions.Logging.LogLevel logLevel, Microsoft.Extensions.Logging.EventId eventId, System.__Canon state, System.Exception exception, System.Func<System.__Canon, System.Exception, string> formatter)
 	Microsoft.Extensions.Logging.dll!Microsoft.Extensions.Logging.Logger.Log<object>(Microsoft.Extensions.Logging.LogLevel logLevel, Microsoft.Extensions.Logging.EventId eventId, object state, System.Exception exception, System.Func<object, System.Exception, string> formatter)
 	Microsoft.Extensions.Logging.Abstractions.dll!Microsoft.Extensions.Logging.Logger<Microsoft.AspNetCore.Hosting.Internal.WebHost>.Log<object>(Microsoft.Extensions.Logging.LogLevel logLevel, Microsoft.Extensions.Logging.EventId eventId, object state, System.Exception exception, System.Func<object, System.Exception, string> formatter)
 	Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.Internal.HostingLoggerExtensions.RequestStarting(Microsoft.Extensions.Logging.ILogger logger, Microsoft.AspNetCore.Http.HttpContext httpContext)
 	Microsoft.AspNetCore.Hosting.dll!Microsoft.AspNetCore.Hosting.Internal.HostingApplication.CreateContext(Microsoft.AspNetCore.Http.Features.IFeatureCollection contextFeatures)
 	Microsoft.AspNetCore.Server.Kestrel.dll!Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame<Microsoft.AspNetCore.Hosting.Internal.HostingApplication.Context>.RequestProcessingAsync()
[...]

Issue #515, opened 2016-11-11
Console logger adds considerable latency to requests


mikeharder:
The console logger adds considerable latency to requests.  Here's a comparsion of the average and maximum latency with no logging, serilog to a file, and console logging:

Metric | No Logging | Serilog to File | Console Logging
--- | --- | --- | ---
Average (ms) | 6.46 | 6.79 | 11.19
Max (ms) | 7.08 | 8.45 | 67.63

https://github.com/mikeharder/AspNetLatency

A few options were discussed to improve console logging performance:

1. Don't block the request waiting for calls to `Console.WriteLine()` (or setting console colors).  Queue writes to something faster like `ConcurrentQueue`, and use a background thread to remove from queue and call `Console.WriteLine()`.  One downside is that log writes may appear out-of-order with respect to direct `Console.WriteLine()` calls.  Another downside is the server process may end before all logs have been flushed from the queue.

2. Still write to the console synchronously during the request, but avoid expensive operations like setting colors.  Could be an option on the existing console logger, or a new `FastConsoleLogger`.

BrennanConroy:
@glennc Colors!

BrennanConroy:
Just an FYI 1.0.0 doesn't have the ~35% perf improvement we added for Console Logging

muratg:
There are a few issues that we're tracking specifically, e.g. https://github.com/aspnet/Logging/issues/408

BrennanConroy:
Tested with #536 and 1.2.0 Console Logger

Metric | No Logging | Serilog to File | Console Logging (1.2.0-*) | Background Console Logging
--- | --- | --- | --- | ---
Average (ms) | 3.347 | 3.517 | 4.606 | 3.2992
Max (ms) | 5.1071 | 5.3236 | 12.5596 | 5.7356

PR #536, opened 2016-12-30
Move Console.Write out of critial path

benaadams:
Resolves https://github.com/aspnet/Logging/issues/532

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -2,20 +2,27 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
+using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Logging.Console.Internal;
 
 namespace Microsoft.Extensions.Logging.Console
 {
     public class ConsoleLogger : ILogger
     {
-        // Writing to console is not an atomic operation in the current implementation and since multiple logger
-        // instances are created with a different name. Also since Console is global, using a static lock is fine.
-        private static readonly object _lock = new object();
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
         private static readonly string _newLineWithMessagePadding;
+
+        // Async Semaphore so the Console is inactive when not actively logging
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(0);
+        private readonly ConcurrentQueue<LogMessageEntry> _messageQueue = new ConcurrentQueue<LogMessageEntry>();
+        private readonly CancellationTokenSource _cts = new CancellationTokenSource();
+        private readonly Task _outputTask;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
@@ -52,6 +59,11 @@ namespace Microsoft.Extensions.Logging.Console
             {
                 Console = new AnsiLogConsole(new AnsiSystemConsole());
             }
+
+            // Start Console message queue processor
+            _outputTask = ProcessLogQueue(_cts.Token);
+
+            RegisterForExit();
         }
 
         public IConsole Console
@@ -155,24 +167,20 @@ namespace Microsoft.Extensions.Logging.Console
 
             if (logBuilder.Length > 0)
             {
-                var logMessage = logBuilder.ToString();
-                lock (_lock)
+                var hasLevel = !string.IsNullOrEmpty(logLevelString);
+                // Queue log message
+                _messageQueue.Enqueue(new LogMessageEntry()
                 {
-                    if (!string.IsNullOrEmpty(logLevelString))
-                    {
-                        // log level string
-                        Console.Write(
-                            logLevelString,
-                            logLevelColors.Background,
-                            logLevelColors.Foreground);
-                    }
+                    Message = logBuilder.ToString(),
+                    LevelString = hasLevel ? logLevelString : null,
+                    LevelBackground = hasLevel ? logLevelColors.Background : null,
+                    LevelForeground = hasLevel ? logLevelColors.Foreground : null
+                });
 
-                    // use default colors from here on
-                    Console.Write(logMessage, DefaultConsoleColor, DefaultConsoleColor);
-
-                    // In case of AnsiLogConsole, the messages are not yet written to the console,
-                    // this would flush them instead.
-                    Console.Flush();
+                if (_semaphore.CurrentCount == 0)
+                {
+                    // Console output Task may be asleep, wake it up
+                    _semaphore.Release();
                 }
             }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -270,6 +309,51 @@ private void GetScopeInformation(StringBuilder builder)
             }
         }
 
+        private async Task ProcessLogQueue()
+        {
+            do
+            {
+                var isExiting = _isExiting;
+                await _semaphore.WaitAsync();
```
BrennanConroy: Shouldn't there be some skip condition here, like
```
if (_messageQueue.Count == 0)
{
    await _semaphore.WaitAsync();
}
```

davidfowl:
Did we decide that we want to do this?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: Added skip; though it should be `IsEmpty` not `Count == 0` as the count test is expensive.

BrennanConroy:
It's something we were interested in, but no one has made a decision here.

benaadams:
@davidfowl not sure; but there are so many user issues caused by this

https://github.com/aspnet/Logging/issues/532
https://github.com/aspnet/KestrelHttpServer/issues/1232
https://github.com/aspnet/KestrelHttpServer/issues/1259
https://github.com/aspnet/KestrelHttpServer/issues/1172

benaadams:
Even @DamianEdwards locks his console (and realises it) in his recent live stream https://youtu.be/_s1STP2-Xhc?t=6m22s ; its a very common head scratching issue...

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: Ref on `IsEmpty` vs `Count` https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/ConcurrentQueue.cs#L301-L303

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -68,6 +106,8 @@ public IConsole Console
             }
         }
 
+        public Action OnFlush { get; set; }
```
BrennanConroy: I don't think we should add this

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -2,21 +2,26 @@
...
     {
-        // Writing to console is not an atomic operation in the current implementation and since multiple logger
-        // instances are created with a different name. Also since Console is global, using a static lock is fine.
-        private static readonly object _lock = new object();
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
         private static readonly string _newLineWithMessagePadding;
 
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(0);
+        private readonly ConcurrentQueue<LogMessageEntry> _messageQueue = new ConcurrentQueue<LogMessageEntry>();
```
davidfowl: Use a blocking collection?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: Could extend `ConsoleLogger` in tests and add it there instead of in the lib? Open to other suggestions.

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: You thinking for upper bounds? Would have to spin up a dedicated thread to use it as its sync blocking.

davidfowl: What about BufferBlock?

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: Removed

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
benaadams: What's your concern? Backpressure, code complexity etc?

Could add backpressure; main complexity is from waiting for Console to output everything before exit; which would be similar with BufferBlock. Have added comments.

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -63,16 +63,19 @@ public void LogsWhenMessageIsNotProvided()
             var sink = t.Item2;
             var exception = new InvalidOperationException("Invalid value");
 
+            ((TestConsole)t.Item1.Console).OnFlush = () =>
```
BrennanConroy: Isn't there technically a race here, where not all `Log*` were called before `OnFlush` is called?

benaadams: Technically... though the semaphore dispatch is slower than the queue. Could add some `ManualResetEvent`s?

BrennanConroy: Another problem is that these Asserts don't cause the test to pass or fail since they run when the test method has finished,

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
davidfowl: Code complexity and scaling this when we need to eventually do someting with backpressure. I'm generally more comfortable with simpler code that achieves the same thing.

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
benaadams: Resolved

benaadams:
@davidfowl switched to cancellation token rather than MRE+volatile flag. Simpler?

benaadams:
Hmm... adding backpressure may mean it needs a dedicated thread otherwise it could deadlock

benaadams:
Changed it a bit to extract the queue/processor so the logger remains mostly the same; minus some code.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,197 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Reflection;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Logging.Console.Internal
+{
+    public class ConsoleLoggerProcessor
```
davidfowl: Avoiding BufferBlock because?

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,197 @@
...
+    {
+        private const int _maxQueuedMessages = 1024;
+
+        private IConsole _console;
+
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(0);
+        private readonly ConcurrentQueue<LogMessageEntry> _messageQueue = new ConcurrentQueue<LogMessageEntry>();
```
pakrym: Why not use `BlockingCollection<T>`? It has built in back-pressure.

benaadams:
Yeah think BlockingCollection now it uses another thread; have reviewed BufferBlock and I think it would make things more complicated; however also not sure I understand it 😆

BrennanConroy:
@benaadams Are you going to change to using a `BlockingCollection<T>`?

Did some latency testing in #515 and now logging has very little effect on latency.

Also, this doesn't fix the hanging issues you linked it would just delay them until the logging buffer got filled up.

benaadams:
@BrennanConroy yes will change. The hangs are generally with user interaction; debugging, local etc and shouldn't occur in prod; but do cause a lot of confusing in dev.

BrennanConroy:
Closing in favor of https://github.com/aspnet/Logging/pull/551

PR #538, opened 2016-12-30
AzureAppServices.Test correct Logging ref

benaadams:


BrennanConroy:
Using a project reference in csproj now

Issue #540, opened 2017-01-07
Remove duplicate TestLogger*


JunTaoLuo:
There are duplicate TestLogger* types in Microsoft.Extensions.Logging.Testing and Microsoft.Extensions.Logging.Test. They don't seem too different so maybe we can remove the duplication.

cc @BrennanConroy looks like when you moved these types to the testing repo, you didn't remove them from this repo so when @kichalla moved them back to this repo, we now have duplicates.

BrennanConroy:
Fixed in #542

PR #542, opened 2017-01-09
Remove duplicate TestLogger

BrennanConroy:
#540 
@JunTaoLuo @glennc 

Minor changes to the public `TestLogger`


test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -1,9 +1,9 @@
 ﻿// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
 using System.Linq;
 using Microsoft.Extensions.Logging.Test;
+using Microsoft.Extensions.Logging.Testing;
 using Xunit;
 
 namespace Microsoft.Extensions.Logging
```
JunTaoLuo: A few of these test classes should be in the `M.E.Logging.Test` namespace instead of `M.E.Logging`

PR #544, opened 2017-01-12
Read initial level values on application start

pakrym:
Fixes: https://github.com/aspnet/Logging/issues/517
Fixes: https://github.com/aspnet/AzureIntegration/issues/5

PR #545, opened 2017-01-19
Port "Read initial level values on application start (#544)" to 1.1.1

pakrym:
See https://github.com/aspnet/Logging/issues/544


BrennanConroy:
:shipit:

PR #548, opened 2017-02-02
Allow console logger settings case insensitive

DoCode:
Allow console logger settings define in case insensitive manner:
```json
  "logSettings": {
    "logLevel": {
      "default": "information",
      "Foo.Bar.ReaderHandler": "trace"
    }
```


BrennanConroy:
dupe of #401

Issue #526, opened 2016-11-30
Console Logger Filtering via IConfiguration - Issues


MichaCo:
There are two catches which might be intended or not

First, the values in the configuration are parsed case sensitive. If you put "error" in lower case, you'll get an exception.

This could easily be changed by calling Enum.Parse(<value>, true, ...) instead of 

```csharp
else if (Enum.TryParse<LogLevel>(value, out level))
{
    return true;
}
```
in ConfigurationConsoleLoggerSettings [line 65](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConfigurationConsoleLoggerSettings.cs#L65)

The second issue is a few lines later, it actually throws an exception whenever it cannot parse the enum.
```csharp
            else if (Enum.TryParse<LogLevel>(value, out level))
            {
                return true;
            }
            else
            {
                var message = $"Configuration value '{value}' for category '{name}' is not supported.";
                throw new InvalidOperationException(message);
            }
```

I think that's bad. 
For the initial application startup, this might be fine, but if you work with reload tokens, the whole logic blows up if you make one mistake one time:
If the configuration has been updated with a wrong configuration value, the token triggers but the code above throws an exception. The exception doesn't get logged or shown anywhere because it's an async "somewhere" task doing the checks...
Because of the exception, no new token and handler get created afterwards, and I have to restart the application to have it read the updated configuration again...

MichaCo:
@muratg @glennc Hey guys, any comment on this
Thanks, M

glennc:
We need to add a finally or something to the logger providers that handle change tokens (like Console). They appear to already have the logic to reset the token, it just gets skipped on an exception.

I think the ideal experience here would be that if any errors occur during reloading of configuration we leave everything as it was before the config was changed, so none of your changes apply until you fix whatever is breaking and then all of them take affect. The reason I think that is because we should also log the fact that an error happened and it seems most likely to be log successfully if we ignore all the changes that caused an exception.

BrennanConroy:
Fixed via https://github.com/aspnet/Logging/pull/549/commits/6525bffb58fe8035bc1604b4b32eece5000c5f22

MichaCo:
@BrennanConroy Great 👍

PR #549, opened 2017-02-06
Ensure change token is reset

BrennanConroy:
#526

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
@@ -47,20 +47,29 @@ namespace Microsoft.Extensions.Logging.Console
 
         private void OnConfigurationReload(object state)
         {
-            // The settings object needs to change here, because the old one is probably holding on
-            // to an old change token.
-            _settings = _settings.Reload();
-
-            foreach (var logger in _loggers.Values)
+            try
             {
-                logger.Filter = GetFilter(logger.Name, _settings);
-                logger.IncludeScopes = _settings.IncludeScopes;
+                // The settings object needs to change here, because the old one is probably holding on
+                // to an old change token.
+                _settings = _settings.Reload();
+
+                foreach (var logger in _loggers.Values)
+                {
+                    logger.Filter = GetFilter(logger.Name, _settings);
+                    logger.IncludeScopes = _settings.IncludeScopes;
+                }
             }
-
-            // The token will change each time it reloads, so we need to register again.
-            if (_settings?.ChangeToken != null)
+            catch (Exception ex)
             {
-                _settings.ChangeToken.RegisterChangeCallback(OnConfigurationReload, null);
+                System.Console.WriteLine($"Error while loading configuration changes.{Environment.NewLine}{ex}");
+            }
+            finally
+            {
+                // The token will change each time it reloads, so we need to register again.
+                if (_settings?.ChangeToken != null)
+                {
+                    _settings.ChangeToken.RegisterChangeCallback(OnConfigurationReload, null);
+                }
             }
         }
 
--- a/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
@@ -695,7 +695,7 @@ namespace Microsoft.Extensions.Logging.Test
                 Cancel = new CancellationTokenSource(),
                 Switches =
                 {
-                    ["Test"] = LogLevel.Information,
+                    ["Test"] = "Information",
                 }
             };
 
@@ -705,7 +705,7 @@ namespace Microsoft.Extensions.Logging.Test
             var logger = loggerFactory.CreateLogger("Test");
             Assert.False(logger.IsEnabled(LogLevel.Trace));
 
-            settings.Switches["Test"] = LogLevel.Trace;
+            settings.Switches["Test"] = "Trace";
 
             var cancellationTokenSource = settings.Cancel;
             settings.Cancel = new CancellationTokenSource();
@@ -726,7 +726,7 @@ namespace Microsoft.Extensions.Logging.Test
                 Cancel = new CancellationTokenSource(),
                 Switches =
                 {
-                    ["Test"] = LogLevel.Information,
+                    ["Test"] = "Information",
                 }
             };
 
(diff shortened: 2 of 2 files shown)


test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -749,6 +751,70 @@ public void ConsoleLogger_ReloadSettings_CanReloadMultipleTimes()
...
+            File.WriteAllText(fileName,
+                @"{
+                    ""LogLevel"": {
+                        ""Test"": ""Tracer""
+                    }
+                  }");
+
+            // Wait for reload to trigger
+            Assert.True(reset.WaitOne(1000));
+            Thread.Sleep(500);
```
BrennanConroy: Not a fan of this, but not sure how else to handle it.

BrennanConroy:
@glennc @kichalla @pakrym

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
pakrym: Write a custom configuration provider exposing reload token to the test.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -45,20 +45,32 @@ public ConsoleLoggerProvider(IConsoleLoggerSettings settings)
...
+            try
+            {
+                // The settings object needs to change here, because the old one is probably holding on
+                // to an old change token.
+                _settings = _settings.Reload();
 
-            foreach (var logger in _loggers.Values)
+                foreach (var logger in _loggers.Values)
+                {
+                    logger.Filter = GetFilter(logger.Name, _settings);
```
kichalla: This would not handle the following suggested by @glennc as you would have already modified state before hitting an exception

> I think the ideal experience here would be that if any errors occur during reloading of configuration we leave everything as it was before the config was changed, so none of your changes apply until you fix whatever is breaking and then all of them take affect. The reason I think that is because we should also log the fact that an error happened and it seems most likely to be log successfully if we ignore all the changes that caused an exception.

Our default implementation ([ConfigurationConsoleLoggerSettings](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Console/ConfigurationConsoleLoggerSettings.cs)) always seems to read from settings file without caching them because of which you would loose the earlier state of good settings.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -45,20 +45,32 @@ public ConsoleLoggerProvider(IConsoleLoggerSettings settings)
...
+                    logger.IncludeScopes = _settings.IncludeScopes;
+                }
+            }
+            catch (Exception ex)
             {
-                logger.Filter = GetFilter(logger.Name, _settings);
-                logger.IncludeScopes = _settings.IncludeScopes;
+                foreach (var logger in _loggers.Values)
+                {
+                    logger.LogError(0, ex, "Error while loading configuration changes.");
```
pakrym: I don't think this is a right place to log errors to.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
BrennanConroy: I don't think we should add extra stuff to rollback loggers, worst case is it sets some logger settings correctly, and the rest get skipped because of an error.

glennc: I am OK with that. My suggestion was "ideal world" adding more caching of data doesn't seem worth it.

IIRC the worst case here is that you end up with the first provider in a bad state, and probably broken. But for things like miss-typing an enum you wouldn't actually get any corruption of settings. Right?

BrennanConroy: Right, you can't corrupt any settings, the loggers will still be usable.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
BrennanConroy: @glennc What should we do here? Some kind of logger logger needed...

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
MichaCo: I agree. And as long as the reload token mechanic works the next time the configuration changes and eventually fixes the broken things, its all good...

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
MichaCo: @pakrym I think it is perfectly fine to log an exception if the configuration reload failed. 
I guess it would just be cleaner to create a logger for the LoggerProvider itself. Logging an exception on all attached loggers is definitely not a good idea, there could be quite a few right?

pakrym: Yes, my concern is that this would be logged into all loggers, and possibly multiple times because of how file system watcher works.

BrennanConroy:
Ping, will merge later today unless anyone has an issue with the way I log a configuration error.
@glennc @pakrym

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -45,20 +45,29 @@ public ConsoleLoggerProvider(IConsoleLoggerSettings settings)
...
+                {
+                    logger.Filter = GetFilter(logger.Name, _settings);
+                    logger.IncludeScopes = _settings.IncludeScopes;
+                }
+            }
+            catch (Exception ex)
             {
-                logger.Filter = GetFilter(logger.Name, _settings);
-                logger.IncludeScopes = _settings.IncludeScopes;
+                System.Console.WriteLine($"Error while loading configuration changes.{Environment.NewLine}{ex}");
```
pakrym: nit: Remove `System.`

BrennanConroy: Need System.Console otherwise it looks for WriteLine in Microsoft.Extensions.Logging.Console

HaoK:
Configuration had this same bug, its too bad we aren't able to use the same helpers

https://github.com/aspnet/Common/blob/dev/src/Microsoft.Extensions.Primitives/ChangeToken.cs

PR #551, opened 2017-02-09
Put ConsoleLogger writes on a background thread

BrennanConroy:
#532

Built on top of https://github.com/aspnet/Logging/pull/536 from @benaadams


src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,109 @@
...
+                if (value == null)
+                {
+                    throw new ArgumentNullException(nameof(value));
+                }
+
+                _console = value;
+            }
+        }
+
+        public bool HasQueuedMessages => _messageQueue.Count > 0;
```
benaadams: `BlockingCollection` doesn't expose `.IsEmpty` from the underlying `ConcurrentQueue`; but testing against `.Count > 0` is a lot more expensive.

Wonder if you should use the explicit ` BlockingCollection(IProducerConsumerCollection<T> collection, int boundedCapacity)` ctor and pass in as the [ctor you are calling does](https://github.com/dotnet/corefx/blob/master/src/System.Collections.Concurrent/src/System/Collections/Concurrent/BlockingCollection.cs#L160-L161); but keep a reference to a ConcurrentQueue for the fast `IsEmpty` test?

Or not expose the new `HasQueuedMessages` property on ConsoleLogger...

/cc @stephentoub

BrennanConroy: We could make it internal, its only used for testing right now

stephentoub: Or if it's really important, instead of using BlockingCollection, just use a ConcurrentQueue and a SemaphoreSlim.

benaadams: Would make it internal till someone wants it public; else its exposing a impl detail that someone will take a dependency on...

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        private static void ProcessLogQueue(object state)
+        {
+            var consoleLogger = (ConsoleLoggerProcessor)state;
+
+            consoleLogger.ProcessLogQueue();
+        }
+
+        private void RegisterForExit()
```
davidfowl: I'd remove all of this and just make this disposable. Then call that from the console logger's dispose. Logger providers are disposed "at the right time" when wired up correctly.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                _console = value;
+            }
+        }
+
+        public bool HasQueuedMessages => _messageQueue.Count > 0;
+
+        public void EnqueueMessage(LogMessageEntry message)
+        {
+            _messageQueue.Add(message);
```
davidfowl: Does this dispatch?

BrennanConroy: Just adds to the internal data structure and the `GetConsumingEnumerable` below is waiting for messages to be added on a separate thread

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -86,6 +75,8 @@ public IConsole Console
 
         public string Name { get; }
 
+        public bool HasQueuedMessages => _queueProcessor.HasQueuedMessages;
```
davidfowl: Remove this, I don't think we need to expose it.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Logging.Console.Internal
+{
+    public class ConsoleLoggerProcessor
+    {
+        private const int _maxQueuedMessages = 1024;
```
davidfowl: What happens when you hit the limit? Does it block the producer?

davidfowl:
We need some throughput tests before and after.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
BrennanConroy: yes

davidfowl: Nice

BrennanConroy:
I'll do some perf testing today

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
BrennanConroy: Unfortunately loggers aren't `IDisposable`

glennc:
Test results? :)

BrennanConroy:
Metric | Before | After |
 --- | --- | --- |
Average (RPS) | 4001.162 | 5007.562
Max (RPS) | 4618.2 | 6120.87
Standard deviation | 824.096 | 824.499

CPU was around 100% the whole time

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
davidfowl: Move this to the ConsoleLoggerProvider (those are disposable)

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -10,17 +10,14 @@ namespace Microsoft.Extensions.Logging.Console
...
-        private static readonly object _lock = new object();
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
         private static readonly string _newLineWithMessagePadding;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
 
-        private IConsole _console;
+        private readonly ConsoleLoggerProcessor _queueProcessor = new ConsoleLoggerProcessor();
```
davidfowl: This spins up a thread per category, we need to move this to the provider so there's a single thread and remeasure.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -74,26 +49,8 @@ private static void ProcessLogQueue(object state)
...
-        private void InitiateShutdown(object sender, EventArgs e)
-#elif NETSTANDARD1_5
-        private void InitiateShutdown(System.Runtime.Loader.AssemblyLoadContext obj)
-#else
-        private void InitiateShutdown()
-#endif
+        public void Dispose()
         {
-            // TODO: Do after _outputTask.Wait(...) in case there are items blocked on getting added?
             _messageQueue.CompleteAdding();
```
davidfowl: Why is Dispose throwing?

BrennanConroy: You mean the ones we catch and ignore?

davidfowl: Ah yes we just swallow 😄 .Good

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -10,18 +10,16 @@ namespace Microsoft.Extensions.Logging.Console
...
         private static readonly string _messagePadding;
         private static readonly string _newLineWithMessagePadding;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
 
-        private IConsole _console;
+        internal ConsoleLoggerProcessor QueueProcessor;
         private Func<string, LogLevel, bool> _filter;
+        private IConsole _console;
```
davidfowl: Why not just pass the `QueueProcessor` as a ctor argument and set the IConsole  on the `QueueProcessor` instead of on the message?

BrennanConroy: Public constructor, internal type. I can do the second one

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -10,17 +10,14 @@ namespace Microsoft.Extensions.Logging.Console
...
-        private static readonly object _lock = new object();
         private static readonly string _loglevelPadding = ": ";
         private static readonly string _messagePadding;
         private static readonly string _newLineWithMessagePadding;
 
         // ConsoleColor does not have a value to specify the 'Default' color
         private readonly ConsoleColor? DefaultConsoleColor = null;
 
-        private IConsole _console;
+        private ConsoleLoggerProcessor _queueProcessor;
```
davidfowl: readonly

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
+
+namespace Microsoft.Extensions.Logging.Console.Internal
+{
+    public class ConsoleLoggerProcessor : IDisposable
+    {
+        private const int _maxQueuedMessages = 1024;
```
davidfowl: File an issue to make this configurable.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        private readonly BlockingCollection<LogMessageEntry> _messageQueue = new BlockingCollection<LogMessageEntry>(_maxQueuedMessages);
+        private readonly Task _outputTask;
+
+        public IConsole Console;
+
+        public ConsoleLoggerProcessor()
+        {
+            // Start Console message queue processor
+            _outputTask = Task.Factory.StartNew(
```
davidfowl: I think we should file an issue to avoid using a dedicated thread. It's probably fine for the most part but it can be done in a way where we don't have a thread spun up all the time. I don't know how much of a problem it is in practice though.

PR #556, opened 2017-02-15
1.1.2 patch train updates

JunTaoLuo:


src/Microsoft.Extensions.Logging.AzureAppServices/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "1.0.0",
+  "version": "1.0.1",
```
BrennanConroy: Is this correct?

JunTaoLuo: We released for the first time in FTS so it's a 1.0.x version number

PR #557, opened 2017-02-15
Update Serilog.Extensions.Logging version

pakrym:
To get scopes capture by default


nblumhardt:
The failure on Travis/Mono is:

```
info: AzureBlobSinkTests.cs(41,26): error CS0433: The type 'Encoding' exists in both
  'System.Text.Encoding, Version=4.0.10.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' and
  'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' 
```

I don't think this is related to anything we've done in _Serilog.Extensions.Logging_, but please let me know if you need a hand to trace changes upstream.

pakrym:
@nblumhardt I think this was caused by a big move to Standard.Library 1.6.1 that was happening at that time in all of our packages.

Issue #516, opened 2016-11-15
Memory leak in Microsoft.Extensions.Logging.Internal.FormattedLogValues


meichtf:
The FormattedLogValues causes a serious memory leak. After a day about 1Gb of memory is wasted in my usage. The cause of the problem is:
private static ConcurrentDictionary<string, LogValuesFormatter> _formatters = new ConcurrentDictionary<string, LogValuesFormatter>();
...
...
_formatters.GetOrAdd(format, f => new LogValuesFormatter(f));

This code will just add objects and never clean them.

glennc:
@meichtf Do you have some examples of what you are logging that fill up the cache? I would like to get the scenario right when we are choosing a fix.

It looks like the only way for this to happen is if you are doing something like combining string interpolation and string formatting, like `_logger.LogInformation($"{DateTime.UtcNow} {{0}}", DateTime.UtcNow);` is something like that what you are doing?

meichtf:
Yes, this is a good example you just provided. I know this is not a great usage of the API, but it should still not cause a leak.

davidfowl:
That's really a bad use of the API, what are your actually doing in your application. There's obviously a trade off being made here. Regular usage of the API doesn't cause problem. It's possible we should just have an upper limit on the amount of values

glennc:
An upper limit on the collection is what I proposed when we first looked at this, and I expect we will just do that. But I would like to see exactly what you are logging that is giving you the problem. If you put the values that are changing frequently in arguments then this problem goes away, is there a reason you can't or don't want to do that? or was it just that it didn't occur to you at the time? or something else?

nblumhardt:
One other option to consider, but which I haven't tried it out, is to only cache formatters for interned strings (`string.IsInterned(s)`):

```csharp
// Literals are interned; would be cached
"Hello {Name}"

// Generated strings are not interned; not cached
"Hello {Name}" + DateTime.UtcNow
```

Suggested originally by @jods4 IIRC.

Serilog uses an upper bound on the cache size currently, but might switch to this strategy or a hybrid one (cap the size for non-interned formats, but always allow interned formats to be added). Curious whether there are any obvious holes in it.

The Roslyn-based analyzer in https://github.com/Suchiman/SerilogAnalyzer detects usage of logging methods with non-constant templates - doing something similar for _MEL_, or broadening the SerilogAnalyzer project to cover it, would be the optimal solution I think :-)

[...]

PR #561, opened 2017-02-21
Cap FormattedLogValues cache

BrennanConroy:
#516 
@glennc @davidfowl 

I tried a ConcurrentLru cache in the initial commit, but it looked like perf halved in that scenario.
Current thought is to just cap the cache to prevent memory leaks, and if perf is a concern then the `LoggerMessage.DefineScope` pattern can be used.

Will file an issue in docs to document that pattern: https://github.com/aspnet/Docs/issues/2810

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/FormattedLogValues.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/FormattedLogValues.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading;
 
 namespace Microsoft.Extensions.Logging.Internal
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Logging.Abstractions/Internal/FormattedLogValues.cs
```diff
@@ -19,12 +19,23 @@ public class FormattedLogValues : IReadOnlyList<KeyValuePair<string, object>>
...
         private readonly object[] _values;
         private readonly string _originalMessage;
+        private const int MaxCachedFormatters = 1024;
 
         public FormattedLogValues(string format, params object[] values)
         {
             if (values?.Length != 0 && format != null)
             {
-                _formatter = _formatters.GetOrAdd(format, f => new LogValuesFormatter(f));
+                if (_formatters.Count >= MaxCachedFormatters)
```
pakrym: ConcurrentDictionary.Count acquires all locks which makes this method slow we either need to do our own counter or switch to locking around `List<T>`

kichalla: You could also reorder the checks here to reduce the impact

pakrym: How?

kichalla: I mean currently the Count check would happen even after the cap is reached but if you were to get a successful hit you could avoid this check.

glennc: What is the impact right now vs not having the check?

niemyjski: So if we are over the limit we just create it? That seems bad, what's the cache eviction policy on this? Seems like frequently occurring formats should never be dropped.

kichalla: @niemyjski You could always use `LoggerMessage.Define` way for perf...Example, MVC usage below:
https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Core/Internal/MvcCoreLoggerExtensions.cs

BrennanConroy:
:up: 📅

davidfowl:
No tests?

BrennanConroy:
Ping

src/Microsoft.Extensions.Logging.Abstractions/Internal/FormattedLogValues.cs
```diff
@@ -14,17 +15,36 @@ namespace Microsoft.Extensions.Logging.Internal
...
+                    if (!_formatters.TryGetValue(format, out _formatter))
+                    {
+                        _formatter = new LogValuesFormatter(format);
+                    }
+                }
+                else
+                {
+                    _formatter = _formatters.GetOrAdd(format, f =>
+                    {
+                        Interlocked.Increment(ref _count);
```
kichalla: https://msdn.microsoft.com/en-us/library/ee378677(v=vs.110).aspx#Anchor_2

pakrym: I don't think 100% correctness is required here, we are preferring performance over it.

Issue #566, opened 2017-03-08
Move NullLoggerFactory to Microsoft.Extensions.Logging.Abstractions


ajaybhargavb:
From https://github.com/aspnet/DataProtection/pull/203#discussion_r104220371

We currently have `NullLogger` and `NullLoggerProvider` in Logging.Abstractions but `NullLoggerFactory` is in Logging.Testing. It should be moved to Logging.Abstractions so it can be used in other places like the one referenced above.

@Eilon

Eilon:
@muratg @davidfowl @DamianEdwards - any objections? This seems like good cleanup because we have a mess of NullLoggerXyz's scattered around.

davidfowl:
👍

PR #568, opened 2017-03-09
Added NullLoggerFactory to Logging.Abstractions

ajaybhargavb:
Issue - #566 
Removing `NullLogger*` types from Logging.Testing will break other repos. That should be done separately.

@Eilon @muratg

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
@@ -1,0 +1,35 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.Abstractions
+{
+    /// <summary>
+    /// An <see cref="ILoggerFactory"/> used to create instance of
+    /// <see cref="NullLogger"/> that logs nothing.
+    /// </summary>
+    public class NullLoggerFactory : ILoggerFactory
+    {
+        public static readonly NullLoggerFactory Instance = new NullLoggerFactory();
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// This returns a <see cref="NullLogger"/> instance which logs nothing.
+        /// </remarks>
+        public ILogger CreateLogger(string name)
+        {
+            return NullLogger.Instance;
+        }
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// This method ignores the parameter and does nothing.
+        /// </remarks>
+        public void AddProvider(ILoggerProvider provider)
+        {
+        }
+
+        public void Dispose()
+        {
+        }
+    }
+}


src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.Extensions.Logging.Abstractions
+{
+    public class NullLoggerFactory : ILoggerFactory
```
Eilon: Doc comments plz kthxbye!

src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
```diff
@@ -3,15 +3,20 @@
 
 namespace Microsoft.Extensions.Logging.Abstractions
 {
+    /// <summary>
+    /// An <see cref="ILoggerFactory"/> used to create instance of <see cref="NullLogger"/>.
```
Eilon: Might be worth adding some text that explicitly states that this logger never logs anything? Same goes for the methods below...

ajaybhargavb: It says that in the docs for `NullLogger`. Do we have to say that here too?

ajaybhargavb: I guess it doesn't hurt to add it here. Will update

ajaybhargavb:
🆙 📅

src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// This returns a <see cref="NullLogger"/> instance which explicitly logs nothing.
+        /// </remarks>
+        public ILogger CreateLogger(string name)
+        {
+            return NullLogger.Instance;
+        }
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// This method explicitly no ops.
```
Eilon: I think this phrasing can be improved, `no ops` is nerd-speak. How about "This method ignores the parameter and does nothing."

src/Microsoft.Extensions.Logging.Abstractions/NullLoggerFactory.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// An <see cref="ILoggerFactory"/> used to create instance of
+    /// <see cref="NullLogger"/> that logs nothing.
+    /// </summary>
+    public class NullLoggerFactory : ILoggerFactory
+    {
+        public static readonly NullLoggerFactory Instance = new NullLoggerFactory();
+
+        /// <inheritdoc />
+        /// <remarks>
+        /// This returns a <see cref="NullLogger"/> instance which explicitly logs nothing.
```
Eilon: Remove `explicitly`.

ajaybhargavb:
🆙 📅

davidfowl:
👏 That only took 8 hours

PR #570, opened 2017-03-11
Added NullLoggerOfT to Logging.Abstractions

ajaybhargavb:
Missed this type when I addressed #566 

@Eilon


davidfowl:
@ajaybhargavb You forgot to remove the NullLogger type from the testing assembly. I'm getting conflicts now 👎

ajaybhargavb:
@davidfowl, I wanted to do that separately. That way I can update other repos without all of them being broken at the same time

PR #575, opened 2017-03-16
add xunit logger provider

### LINKED ISSUE #574: Create "xunit" logger provider
Moved from https://github.com/aspnet/Testing/issues/243 since it fits better into the Microsoft.Extensions.Logging.Testing package. See the original issue for context.

anurse:
moved from https://github.com/aspnet/Testing/pull/260

fixes #574 

/cc @Eilon @moozzyk


src/Microsoft.Extensions.Logging.Testing/XUnitLoggerProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            return new XUnitLogger(_output, categoryName, _minLevel);
+        }
+
+        public void Dispose()
+        {
+        }
+    }
+
+    public class XUnitLogger : ILogger, IDisposable
```
BrennanConroy: Why `IDisposable`

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="xunit.extensibility.core" Version="$(XunitVersion)" />
```
natemcmaster: Can this be xunit.abstractions? We only need [ITestOutputHelper](https://github.com/xunit/abstractions.xunit/blob/master/src/xunit.abstractions/TestCases/ITestOutputHelper.cs), right?

src/Microsoft.Extensions.Logging.Testing/XUnitLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.Testing;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging
+{
+    public static class XUnitLoggerFactoryExtensions
+    {
+        public static void AddXUnit(this ILoggerFactory loggerFactory, ITestOutputHelper output)
```
BrennanConroy: Similar extension methods return the factory

test/Microsoft.Extensions.Logging.Testing.Tests/XUnitLoggerProviderTest.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        [Fact]
+        public void LoggerProviderWritesToTestOutputHelper()
+        {
+            var testTestOutputHelper = new TestTestOutputHelper();
+            var loggerFactory = new LoggerFactory();
+            loggerFactory.AddXUnit(testTestOutputHelper);
+
+            var logger = loggerFactory.CreateLogger("TestCategory");
+            logger.LogInformation("This is some {0} information", "great");
```
BrennanConroy: The `{0}` stuff isn't necessary

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -15,6 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
```
Eilon: Shouldn't this be its own package?

src/Microsoft.Extensions.Logging.Testing/XUnitLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.Testing;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging
+{
+    public static class XUnitLoggerFactoryExtensions
```
natemcmaster: Nit: lower case u in "Xunit".

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
anurse: So, yes it can... But `XunitVersion` is set to `2.2.0`, but `xunit.abstractions` is only version `2.0.1`. I wasn't sure which was less bad: 1) Adding an `XunitAbstractionsVersion` variable, 2) Setting `xunit.abstractions` to `2.0.1` without a variable or 3) Using `xunit.extensibility.core`. Very open to suggestions :)

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
anurse: It certainly could be. We could consider putting it in Common as a sources package.

Eilon: Hmm, you know, I didn't realize that this project already was xUnit-specific, which is very odd to me. It's so far just because of LogValuesAssert, which throws an xUnit-specific exception.

Is there any reason for this library to be tied to xUnit? The LogValuesAssert case could just change to throw a regular exception (not as nice as xUnit's exception, but it'll work just fine).

Then xUnit-specific stuff could go into a new library.

Thoughts?

anurse: I don't really have a problem keeping this library xUnit-specific. It's not a very big library. We could split into two libraries if there was demand for it, but I don't know that I'd front-load that work now. Meh, I'm good either way.

natemcmaster: I'm in favor of keeping this in the same library.

src/Microsoft.Extensions.Logging.Testing/XunitLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Logging.Testing;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging
+{
+    public static class XunitLoggerFactoryExtensions
+    {
+        public static ILoggerFactory AddXUnit(this ILoggerFactory loggerFactory, ITestOutputHelper output)
```
natemcmaster: Nit: lowercase u in 'Xunit'

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
Eilon: Yeah I'm ok with this because it doesn't really change the situation. It's just very odd that we would do this - I doubt we did it knowingly to begin with.

Issue #563, opened 2017-02-23
Logging after dispose issues?


davidfowl:
Seems like kestrel tries to log during shutdown if there's a critical error and that explodes in a weird way because we're trying to write to a blocking collection that has had CompleteAdding called on it.

```
C:\Users\dfowler\Documents\Visual Studio 2017\Projects\WebApplication24\WebApplication24>dotnet run
Hosting environment: Production
Content root path: C:\Users\dfowler\Documents\Visual Studio 2017\Projects\WebApplication24\WebApplication24
Now listening on: http://localhost:5000
Application started. Press Ctrl+C to shut down.
fail: Microsoft.AspNetCore.Server.Kestrel[0]
      UvTimerCb
System.MissingMethodException: Method not found: 'System.Span`1<Byte> System.Buffers.IOutput.get_Buffer()'.
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.<>c.<.cctor>b__74_0(IntPtr ptr, IntPtr arg)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.Libuv.NativeMethods.uv_walk(UvLoopHandle loop, uv_walk_cb walk_cb, IntPtr arg)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.KestrelThread.OnHeartbeat(UvTimerHandle timer)
   at Microsoft.AspNetCore.Server.Kestrel.Internal.Networking.UvTimerHandle.UvTimerCb(IntPtr handle)

Unhandled Exception: System.AggregateException: An error occurred while writing to logger(s). (The collection has been marked as complete with regards to additions.) ---> System.InvalidOperationException: The collection has been marked as complete with regards to additions.
   at System.Collections.Concurrent.BlockingCollection`1.TryAddWithNoTimeValidation(T item, Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at Microsoft.Extensions.Logging.Console.Internal.ConsoleLoggerProcessor.EnqueueMessage(LogMessageEntry message)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.WriteMessage(LogLevel logLevel, String logName, Int32 eventId, String message, Exception exception)
   at Microsoft.Extensions.Logging.Console.ConsoleLogger.Log[TState](LogLevel logLevel, EventId eventId, TState state, Exception exception, Func`3 formatter)
[...]

PR #578, opened 2017-03-24
Logging shouldn't throw after dispose

davidfowl:
- Check complete adding before adding new messages
to the list.
- Added a test

#563

PS: I didn't measure the performance of checking that bool on every write.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
+++ b/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
@@ -27,7 +27,10 @@ namespace Microsoft.Extensions.Logging.Console.Internal
 
         public virtual void EnqueueMessage(LogMessageEntry message)
         {
-            _messageQueue.Add(message);
+            if (!_messageQueue.IsAddingCompleted)
+            {
+                _messageQueue.Add(message);
+            }
         }
 
         // for testing
--- a/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
+++ b/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
@@ -29,8 +29,16 @@ namespace Microsoft.Extensions.Logging.Console.Internal
         {
             if (!_messageQueue.IsAddingCompleted)
             {
-                _messageQueue.Add(message);
+                try
+                {
+                    _messageQueue.Add(message);
+                    return;
+                }
+                catch (InvalidOperationException) { }
             }
+
+            // Adding is completed so just log the message
+            WriteMessage(message);
         }
 
         // for testing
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -27,7 +27,10 @@ public ConsoleLoggerProcessor()
 
         public virtual void EnqueueMessage(LogMessageEntry message)
         {
-            _messageQueue.Add(message);
+            if (!_messageQueue.IsAddingCompleted)
+            {
+                _messageQueue.Add(message);
```
BrennanConroy: I'm not sure how often this happens in production or if it ever does, but `Add` blocks if the queue is full, so this could throw still.

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -27,7 +27,10 @@ public ConsoleLoggerProcessor()
 
         public virtual void EnqueueMessage(LogMessageEntry message)
         {
-            _messageQueue.Add(message);
+            if (!_messageQueue.IsAddingCompleted)
```
BrennanConroy: Other threads can call `EnqueueMessage` while Dispose is being called, so this doesn't get rid of the issue, just decreases how often it happens.

pakrym:
We need to fix hosting issue first, maybe this exception is not that bad too, it prevents loosing log messages.

davidfowl:
@pakrym no we don't. We need to fix both bugs. Logging is one of those things that *really* should never throw. People don't put try catch statements around their logs.

davidfowl:
This is it still racey as @BrennanConroy points out. I thought I read that CompleteAdding waits on any outstanding adds to finishing before switching the state but it just waits on other calls to CompleteAdding to complete. Add does wait on CompleteAdding, but then it throws. 

I still think this should never throw after add and this lessens the chances of that (plus this *never* happened before since we wrote directly to the console). I'll see if there's any other collection we could use.

davidfowl:
Will revisit

pakrym:
I changed this PR to catch `InvalidOperationException` thrown by `.Add` and log messages even if processor is disposed.

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -863,17 +863,29 @@ public void WriteCore_NullMessageWithNullException()
...
+        [Fact]
+        public void LogAfterDisposeDoesNotThrow()
+        {
+            var sink = new ConsoleSink();
+            var console = new TestConsole(sink);
+            var processor = new ConsoleLoggerProcessor();
+            var logger = new ConsoleLogger(_loggerName, filter: null, includeScopes: false, loggerProcessor: processor);
+            logger.Console = console;
+            processor.Dispose();
+            logger.LogInformation("Logging after dispose");
```
BrennanConroy: Could we change this test to check that "Logging after dispose" gets logged

davidfowl:
👏

PR #579, opened 2017-03-24
Remove net451 as a cross-compile target

pranavkm:


pranavkm:
@karolz-ms - could you have a look at the event source changes? We're targeting net46 which seems to have complex type support (tests pass), but I'm not sure if there's more to it.

test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -649,7 +650,7 @@ public void CallingBeginScopeOnLogger_ReturnsNonNullableInstance()
             Assert.NotNull(disposable);
         }
 
-#if NET452
+#if NET46
```
BrennanConroy: This test shouldn't be needed anymore now that we are using `AsyncLocal` in .NET Framework

karolz-ms:
@pranavkm Looks good.

PR #581, opened 2017-03-26
Disable API Check for package that we do not push to NuGet

dougbu:
@Eilon FYI if we have plans to ship this on NuGet, we'll need a follow-up issue to create baselines and revert this change after that release.


dougbu:
2fa414a

PR #592, opened 2017-03-31
Xunit logger ignores exceptions from output helper

### LINKED ISSUE #591: Xunit Logger can crash on a background thread
If the Xunit Logger is called from a background thread and the test in question is no longer running, xunit throws an InvalidOperationException.

We still want to be able to log from potential background thread events if possible and we don't want crashes, so I suggest making the logger ignore this exception and simply drop the log message. We're going to be adding in file-based logging for tests as well, which **will** be able to capture these results.

/cc @natemcmaster

anurse:
fixes #591 

blocking aspnet/Hosting#999


anurse:
Because this is blocking things, I'm going to merge when I get green builds. Comments can be addressed afterwards.

PR #597, opened 2017-04-12
Make AddFilter chainable, add some doc comments, add UseConfiguration…

BrennanConroy:
… to LoggerFactory


samples/SampleApp/Program.cs
```diff
@@ -25,15 +21,16 @@ public Program()
...
             // Create a logger factory with filters that can be applied across all logger providers.
-            var factory = new LoggerFactory(loggingConfiguration.GetSection("Logging"));
-            factory.AddFilter(new Dictionary<string, LogLevel>
-            {
-                { "Microsoft", LogLevel.Warning },
-                { "System", LogLevel.Warning },
-                { "SampleApp.Program", LogLevel.Debug }
-            });
+            var factory = new LoggerFactory()
+                .UseConfiguration(loggingConfiguration.GetSection("Logging"))
```
glennc: Not sure of this name. It kind of implies that you are configuring the factory, like adding providers, instead of just filtering.

BrennanConroy:
@eilon Some doc comments

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -41,11 +42,35 @@ public LoggerFactory(IConfiguration configuration)
                 throw new ArgumentNullException(nameof(configuration));
             }
 
+            UseConfiguration(configuration);
+        }
+
+        /// <summary>
+        /// Replaces the <see cref="IConfiguration"/> used for filtering.
+        /// </summary>
+        /// <param name="configuration">The replacing configuration.</param>
```
Eilon: `The new configuration to use.` ?

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -127,7 +152,15 @@ public void AddProvider(string providerName, ILoggerProvider provider)
             }
         }
 
-        public void AddFilter(string providerName, string categoryName, Func<LogLevel, bool> filter)
+        /// <summary>
+        /// Adds a filter that applies to <paramref name="providerName"/> and <paramref name="categoryName"/> with the given
+        /// <paramref name="filter"/>, returning true means allow log through, false means reject log.
```
Eilon: Remove the `, returning ...` section because it's just describing the parameter. Put that text in the description of how ti implement the `filter` parameter.

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -41,11 +42,35 @@ public LoggerFactory(IConfiguration configuration)
                 throw new ArgumentNullException(nameof(configuration));
             }
 
+            UseConfiguration(configuration);
+        }
+
+        /// <summary>
+        /// Replaces the <see cref="IConfiguration"/> used for filtering.
```
Eilon: I'm a bit late on the design here, but why does this *replace* the configuration used for filtering? What if I have 2 sets of config that I want to use from 2 different sections? This seems to me like it should all be additive, always.

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -603,31 +603,6 @@ public void AddFilterForNonMatchingProviderDoesNotFilter()
         }
 
         [Fact]
```
Eilon: A bunch of code changed in this PR - should there not be more tests for this? E.g. the new extension methods, the new `UseConfiguration()` method, etc.

src/Microsoft.Extensions.Logging/LoggerFactory.cs
BrennanConroy: I think usage will mostly be `factory.UseConfiguration(config.GetSection("Logging"))`
with the config having:
```json
{
  "OtherStuff": {
  },
  "Logging": {
    "LogLevel": {
        "System": "Trace"
    }
  }
}
```

When would you want to use multiple sections?

Eilon: Why would we specifically disallow it?

Eilon: I'm ok punting on this now; let's log a bug to track this discussion.

BrennanConroy:
🆙 📅

BrennanConroy:
Need to do https://github.com/aspnet/MetaPackages/issues/45 and `Remove passing logging section by default in hosting` after this change.

src/Microsoft.Extensions.Logging/LoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// <summary>
+        /// Adds a filter that applies to <paramref name="providerName"/> and <paramref name="categoryName"/>, allowing logs with the given
+        /// minimum <see cref="LogLevel"/> or higher.
+        /// </summary>
+        /// <param name="factory">The <see cref="LoggerFactory" /> to apply the filter to.</param>
+        /// <param name="providerName">The name of the provider.</param>
+        /// <param name="categoryName">The name of the logger category.</param>
+        /// <param name="minLevel">The minimum <see cref="LogLevel"/> that logs from
+        /// <paramref name="providerName"/> and <paramref name="categoryName"/> are allowed.</param>
+        public static LoggerFactory AddFilter(this LoggerFactory factory, string providerName, string categoryName, LogLevel minLevel)
```
Eilon: General question: Why are some of these extension methods, but some are methods on LoggerFactory itself?

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -58,6 +58,96 @@ public void ChangingConfigReloadsDefaultFilter()
...
+
+            // Act
+            logger.LogTrace("Message");
+
+            // Assert
+            writes = loggerProvider.Sink.Writes;
+            Assert.Equal(1, writes.Count);
+        }
+
+        public void UseConfigurationReplacesOldConfiguration()
```
Eilon: Umm... missing something here??? 😄

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -132,7 +162,16 @@ public void AddProvider(string providerName, ILoggerProvider provider)
...
+        /// <summary>
+        /// Adds a filter that applies to <paramref name="providerName"/> and <paramref name="categoryName"/> with the given
+        /// <paramref name="filter"/>.
+        /// </summary>
+        /// <param name="providerName">The name of the provider.</param>
+        /// <param name="categoryName">The name of the logger category.</param>
+        /// <param name="filter">The filter that applies to logs for <paramref name="providerName"/> and <paramref name="categoryName"/>,
+        /// returning true means allow log through, false means reject log.</param>
+        /// <returns>The <see cref="LoggerFactory"/> so that additional calls can be chained.</returns>
+        public LoggerFactory AddFilter(string providerName, string categoryName, Func<LogLevel, bool> filter)
```
Eilon: Maybe add a `null` check for `filter`?

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -58,6 +58,96 @@ public void ChangingConfigReloadsDefaultFilter()
         }
```
Eilon: Do all the filter methods (whether instance methods or extension methods) have at least one or two tests each? (For whatever interesting combinations of params they have.)

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
BrennanConroy: Hahaha! Maybe thats why it passes ;)

src/Microsoft.Extensions.Logging/LoggerFactoryExtensions.cs
BrennanConroy: Performance mainly

Eilon: Wuh?? Performance of what?

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
BrennanConroy: Yes, made sure they all have at least 1 test. I wanted to do a programmatic way of testing all the filters with each other etc. but it was going to be complex and crazy so...

src/Microsoft.Extensions.Logging/LoggerFactoryExtensions.cs
Eilon: Talked in person, the perf is about optimizing the instance methods because they can do smarter stuff w/ data structures. But then there's no reason to have extension methods, they might as well all be instance.

PR #598, opened 2017-04-14
Fix AddProvider to add providers to existing loggers

BrennanConroy:


test/Microsoft.Extensions.Logging.Test/LoggerTest.cs
```diff
@@ -101,6 +101,22 @@ public void Log_AggregatesExceptionsFromMultipleLoggers()
...
+        public void LoggerCanGetProviderAfterItIsCreated()
+        {
+            // Arrange
+            var store = new List<string>();
+            var loggerFactory = new LoggerFactory();
+            var logger = loggerFactory.CreateLogger("Test");
+            loggerFactory.AddProvider(new CustomLoggerProvider("provider1", ThrowExceptionAt.None, store));
+
+            // Act
+            logger.LogInformation("Hello");
```
kichalla: Is this fix related to logs not being visible from Kestrel? Have you tested this fix with an end-to-end application?

BrennanConroy: Yes, tested with the Mvc Sample and Kestrel sample

PR #601, opened 2017-04-18
copy test logging stuff from IntegrationTesting here

anurse:
@davidfowl really wanted this available as low as possible, so I've moved it from IntegrationTesting to here, Logging.Testing. This is a noship package already.

@Eilon The Serilog dependency is moving once again, from Hosting to here. Once this is merged, I'll be converting the IntegrationTesting package to use it and removing the dependency there.

I also restructured this to allow me to write some tests for it.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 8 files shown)


src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -0,0 +1,147 @@
+﻿using System;
```
natemcmaster: Nit: copyright header

src/Microsoft.Extensions.Logging.Testing/LoggedTest.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
```
natemcmaster: Copyright

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
```diff
@@ -0,0 +1,97 @@
+﻿using System;
```
natemcmaster: Copyright

anurse:
:up: :date:

PR #608, opened 2017-04-27
Fix #607 - Implements equality members on EventId.

### LINKED ISSUE #607: Clarify the semantics of EventId and possibly override Equals, etc.
The API of EventId makes it look like the name is optional--i.e. I can create one with just an int. But if I do this, should it be considered the same as an EventId created with the same int but also a name?

If these are supposed to represent the same event, then shouldn't Equals, etc. be implemented on the type? if they are not, then can this be documented on the class to make it clear that the name, while being optional, is significant in the semantics of the type?

anpete:
cc @ajcvickers

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
@@ -5,46 +5,53 @@ namespace Microsoft.Extensions.Logging
 {
     public struct EventId
     {
-        private int _id;
-        private string _name;
-
-        public EventId(int id, string name = null)
-        {
-            _id = id;
-            _name = name;
-        }
-
-        public int Id
-        {
-            get
-            {
-                return _id;
-            }
-        }
-
-        public string Name
-        {
-            get
-            {
-                return _name;
-            }
-        }
-
         public static implicit operator EventId(int i)
         {
             return new EventId(i);
         }
 
+        public static bool operator ==(EventId left, EventId right)
+        {
+            return left.Equals(right);
+        }
+
+        public static bool operator !=(EventId left, EventId right)
+        {
+            return !left.Equals(right);
+        }
+
+        public EventId(int id, string name = null)
+        {
+            Id = id;
+            Name = name;
+        }
+
+        public int Id { get; }
+        public string Name { get; }
+
         public override string ToString()
         {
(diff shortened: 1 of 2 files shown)


anpete:
ping @BrennanConroy

anpete:
ping @ajcvickers @Eilon

src/Microsoft.Extensions.Logging.Abstractions/EventId.cs
```diff
@@ -5,46 +5,53 @@ namespace Microsoft.Extensions.Logging
 {
     public struct EventId
     {
-        private int _id;
-        private string _name;
-
-        public EventId(int id, string name = null)
+        public static implicit operator EventId(int i)
```
Eilon: `i` --> `id` ?

muratg:
@Eilon, should we close this one?

Eilon:
Hmm I think this is still a good change, no? @anpete ?

anpete:
@Eilon I think so. We were waiting on an answer to https://github.com/aspnet/Logging/issues/607#issuecomment-298964897

Eilon:
I pingo-pongo'ed that thread, thanks!

ajcvickers:
@Eilon Note that this is a breaking change for any code that is relying on the existing behavior.

Eilon:
Ah, fair point. But if we had to guess, it probably wouldn't *really* break anyone, right? I'd be surprised if many people are even testing for equality on these instances, and even if they are, they probably don't realize their code doesn't work... would you agree?

ajcvickers:
@Eilon Yes, I agree.

davidfowl:
I think the bar should be super high for braking changes with the Microsoft.Extensions.* as they have been widely adopted by the broader .net ecosystem. That said, this looks benign.

Eilon:
@ajcvickers - ok.
@davidfowl - I agree about the bar, and I think in this case it's acceptable.

anpete:
OK, will get this in.

davidfowl:
@anpete this done?

anpete:
@davidfowl Fell off my radar. We still want it?

davidfowl:
Yea, it looks harmless. Do you guys need it still?

davidfowl:
@Eilon @anpete lets close or merge.

PR #617, opened 2017-05-07
ILogger.IsEnabled(LogLevel) was returning 'True' when called with 'LogLevel.None' value

Yves57:
When the application LogLevel is dynamically calculed (for instance by configuration file) Something like `level != LogLevel.None && logger.IsEnabled(level)` is necessary.


BrennanConroy:
Could you add tests please?

muratg:
@BrennanConroy could you review please?

BrennanConroy:
:shipit: 
Thanks @Yves57 for the fix!

Issue #346, opened 2016-01-24
Consider a way to configure logging provider using the DI system


davidfowl:
This code always looks a little bizzare:

``` C#
public void ConfigureServices(IServiceCollection services)
{
    ...
}

// This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
{
    loggerFactory.AddConsole();
    ...
}
```

The fact that hosting adds the default ILoggerFactory is fine, but having to resolve it in Configure to add logging providers seems anti to how we've integrated the rest of the stack (it feels bolted on). Also, configuring logging in ConfigureServices would  mean the only thing we configure in Configure would be the http stuff (for the most part).
## Proposal 1

The default logger factory should resolve an IEnumerable<ILoggerProvider> in its ctor.

``` C#
public void ConfigureServices(IServiceCollection services)
{
    services.AddMvc();

    services.AddConsoleLogger(LogLevel.Debug);   
}
```

```
public LoggerFactory(IEnumerable<ILoggerProvider> loggerProviders)
```

The only downside if that more extension methods get added to `IServiceCollection` but meh.
## Proposal 2

Add `LoggerOptions` and hang an IList<ILoggerProvider> off it.

``` C#
public void ConfigureServices(IServiceCollection services)
{
    services.AddMvc();

    services.Configure<LoggerOptions>(options => 
    {
        options.Providers.AddConsole(LogLevel.Debug);
    });   
}
```

/cc  @DamianEdwards @lodejard @rynowak

lodejard:
All of those changes would make the logging configuration more complex, and also defer when logging "kicks in". By adding the logging provider as early as the startup class ctor in plain-old-c# you can also write messages in the configure services method, much earlier than when the options system and IOC is active.

Also with the Program.Main hosting model we might also want to add a way to pass an application-created ILoggerFactory in to the web host builder.

So yeah, end of the day I think a number of subsystems like IFileProvider, IConfiguration, ILoggerFactory are designed with IoC in mind, and may be added like services, but it's not a good idea to go the rest of the way and make their activation and building depend on DI or options.

davidfowl:
> All of those changes would make the logging configuration more complex, and also defer when logging "kicks in". By adding the logging provider as early as the startup class ctor in plain-old-c# you can also write messages in the configure services method, much earlier than when the options system and IOC is active.

[...]

Issue #620, opened 2017-05-12
Provide way to have DI services participate in the logger configuration (broken with 2.0 changes)


epignosisx:
Following the [announcement](https://github.com/aspnet/Announcements/issues/238) about the breaking changes in logging for 2.0, I noticed that we were losing the ability to add logging providers in ```Startup.Configure```. I posted the question in the [discussion](https://github.com/aspnet/Logging/issues/615#issuecomment-299353184) issue and it was later [confirmed](https://github.com/aspnet/Logging/issues/615#issuecomment-300416185) by @davidfowl.

Before I was able to configure serilog as such:

```cs
public void Configure(ILoggerFactory factory, IHttpContextAccessor httpContextAccessor)
{
    var logger = new Serilog.LoggerConfiguration()
        .WriteTo.RollingFile("log-{Date}.txt")
        .Enrich.With(new WebEnricher(httpContextAccessor))
        .CreateLogger();

    factory.AddSerilog(logger);
}
```

Note how the ```WebEnricher``` depends on the ```IHttpContextAccessor```. With the new style of configuring logging, we do not have access to DI.

pakrym:
Fixed via https://github.com/aspnet/Logging/issues/626 
You can now inject services into `ILoggerProvider`

PR #626, opened 2017-05-24
Logging in DI

pakrym:
#346


pakrym:
Idea is to move `ILoggerProvider`s to DI and use existing options infrastructure to manage logger configuration/filtering.

/cc @cwe1ss @nblumhardt @khellang

davidfowl:
We should add an AddProvider extension method off ILoggerbuilder

pakrym:
@davidfowl done

khellang:
I've mentioned it before, but I'll say it again; I'm _really_ not a fan of changing the return type (to `ILoggerBuilder`) it breaks chainability and forces me away from using expressions.

IMO, It's a much better alternative to use `Action<ILoggerBuilder>`. The pattern is already scattered around the different repos, but it's not really consistent, and I seem to remember some pushback on the pattern way back. What are the guidelines on this now?

pakrym:
@khellang I think it makes sense to take delegate if it's execution is deferred. In case of logging we would have to immediately execute it because the only thing it does is mutating service collection. So in case of logging we decided to go the same way MVC did (`IMvcBuilder`) and return a builder.

khellang:
> I think it makes sense to take delegate if it's execution is deferred.

It also makes sense for scoping, i.e.

```csharp
var provider = new ServiceCollection()
    .AddLogging(logging => { /* providers can only be added in here */ })
    .BuildServiceProvider();
```

as opposed to

```csharp
var services = new ServiceCollection();
var loggingBuilder = services.AddLogging();
// providers can be added here...
var provider = services.BuildServiceProvider();
// what happens if you try to add a provider here?
```

> So in case of logging we decided to go the same way MVC did (`IMvcBuilder`) and return a builder.

Yeah, I realize there's a mixture of `I*Something*Builder` and the `Action<I*Something*Builder>` throughout the APIs. AFAIK, Entity Framework does something similar. That's why I asked

> What are the guidelines on this now?

There's even new APIs in this PR, like `Action<ConsoleLoggerOptions>`. That's not deferred, is it?

Anyway, it's pretty easy to wrap these kinds of APIs in extension methods to achieve what I want, like

```csharp
public static class ServiceCollectionExtensions
{
    public static IServiceCollection AddLogging(this IServiceCollection services, Action<ILoggingBuilder> configure)
    {
        configure(services.AddLogging());
        return services;
    }
}
```

So it's not really a big deal :smile:

src/Microsoft.Extensions.Logging.TraceSource/TraceSourceFactoryExtensions.cs
```diff
@@ -12,39 +13,39 @@ public static class TraceSourceFactoryExtensions
         /// <summary>
         /// Adds a TraceSource logger named 'TraceSource' to the factory.
         /// </summary>
-        /// <param name="factory">The <see cref="LoggerFactory"/> to use.</param>
+        /// <param name="collection">The <see cref="LoggerFactory"/> to use.</param>
         /// <param name="switchName">The name of the <see cref="SourceSwitch"/> to use.</param>
-        public static LoggerFactory AddTraceSource(
-            this LoggerFactory factory,
+        public static ILoggerBuilder AddTraceSource(
+            this ILoggerBuilder collection,
```
nblumhardt: This argument is called `collection` in some, and `builder` in other similar methods.

src/Microsoft.Extensions.Logging/ConfigurationLoggerFilterConfigureOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+            foreach (var section in configurationSection.AsEnumerable(true))
+            {
+                if (TryGetSwitch(section.Value, out var level))
+                {
+                    rules.Add(new LoggerFilterRule(logger, section.Key, level, null));
+                }
+            }
+        }
+
+        public string ExpandLoggerAlias(string name)
```
nblumhardt: This doesn't seem like a scalable API; it'll also deter users from stepping beyond these basic providers because the natural guess will be wrong. I.e., if I'm a user looking at an existing configuration, and want to add a new one for, say, a file, I'll try `"File"` and nothing will happen... It also seems a bit questionable to favor, say, Azure App Services, but not Seq, Splunk, Elasticsearch, CloudWatch....

src/Microsoft.Extensions.Logging/ILoggerBuilder.cs
```diff
@@ -0,0 +1,22 @@
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Logging
+{
+    public interface ILoggerBuilder
```
nblumhardt: Why introduce `ILoggerBuilder` at all, here? If it's just a wrapper for `IServiceCollection`, and all the provider extension methods do is dot through to `Services`, isn't this unnecessary ceremony?

src/Microsoft.Extensions.Logging/LogMessageFilter.cs
```diff
@@ -0,0 +1,4 @@
+namespace Microsoft.Extensions.Logging
+{
+    public delegate bool LogMessageFilter(string loggerType, string categoryName, LogLevel level);
```
nblumhardt: This seems more like a `LogCategoryFilter` or something like that - it doesn't touch the log message itself.

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -9,59 +9,19 @@ namespace Microsoft.Extensions.Logging
...
         {
-            if (_loggers == null)
+            if (Loggers == null)
             {
                 return;
             }
 
             List<Exception> exceptions = null;
-            foreach (var loggerInfo in _loggers)
+            foreach (var loggerInfo in Loggers)
```
nblumhardt: Race condition; `Loggers` is public and might have been set to null between this and the check above.

nblumhardt:
I think it'd be good to state the goals of this PR a bit more clearly. Is there really much benefit building logging atop DI? (Apologies if I've overlooked an earlier discussion around this.)

In most apps I've worked on, we've gone to great pains to configure logging at start-up and stand it up independently of anything as complex as DI. Configuration errors in happen all the time, and often they're environment specific, so getting log events out even if DI configuration crashes is pretty useful.

src/Microsoft.Extensions.Logging/ConfigurationLoggerFilterConfigureOptions.cs
pakrym: Yep, it's temporary, we would way to extend the list.

src/Microsoft.Extensions.Logging/ILoggerBuilder.cs
pakrym: We want to prevent all logging related methods appearing on `IServiceCollection`. If we keep them on `IServiceCollection` they would also need to be named in more verbose way (AddConsoleLogger, AddLoggingFilter, etc. )

src/Microsoft.Extensions.Logging/LogMessageFilter.cs
pakrym: I'm thinking about removing it all together because most `AddFilter` overloads are forced to use Action<*> as filter delegates.

epignosisx:
@nblumhardt I agree with you about

>In most apps I've worked on, we've gone to great pains to configure logging at start-up and stand it up independently of anything as complex as DI. Configuration errors in happen all the time, and often they're environment specific, so getting log events out even if DI configuration crashes is pretty useful.

I've been bitten by the app failing on startup, but how would you go about resolving this issue I brought up https://github.com/aspnet/Logging/issues/620

glennc:
@nblumhardt 

- as @epignosisx raised there are people who are already injecting services and using them with the logging system, I think we already have some other providers that need services today. Which we completely broke when we went the other way in preview1. Ideally we would provide a way to use services in DI without resorting to something like a service locator.
- having providers in DI means that it is easy for various components to add providers and the composition and configuration model is well known and understood. This can be done either in program main or in `ConfigureServices` because it is understood how all the services compose and operate. Having as much of the stack as possible be configured and registered the same way is beneficial, as is the removal of special top-level APIs on `WebHostBuilder`.
- we will log using `EventSource` before the container is built, though we discussed other things as well that are possible like a single log sink that can be set on the builder or something of that sort. You are right that this is a fundamental limitation of the design. Today we log very little before the container is built, in fact I don't think we log anything. But we would use `EventSource` as we added logging. For logging that an app dev wants to use they can also presumably configure Serilog statically in program main and log to Serilog directly there, registering the provider for the rest of their app. Right? Just thinking "out loud" on that last bit.

There are also some questions about how you can use logging outside of ASP.NET Core, like with EF in a UWP application, that we need to consider.

nblumhardt:
Thanks for the follow-ups @epignosisx @glennc.

> hey can also presumably configure Serilog statically in program main and log to Serilog directly there, registering the provider for the rest of their app. Right? 

This is a good point, and I think it works if you use Serilog. From the perspective of the library as a whole it still seems like there should be a way for developers to get start-up errors through their logging pipeline using only MEL. Having to go to a separate collection point does leaves the problem on the end-user developer's plate, it would seem.

#620 is an interesting problem; there are ways around it by using static state (put the `IHttpContextAccessor` in some well-known location when it becomes available, and have the enricher look for it there). Not ideal, but might be preferable to tangling up DI and logging inextricably.

I wish I could add more useful ideas here - it's a challenging API to get right! :-)

src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
```diff
@@ -19,14 +20,31 @@ public static class LoggingServiceCollectionExtensions
...
+            return AddLogging(services, builder => { });
+        }
+
+        /// <summary>
+        /// Adds logging services to the specified <see cref="IServiceCollection" />.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
+        /// <param name="configure">The <see cref="ILoggerBuilder"/> configuration delegate.</param>
+        /// <returns>The <see cref="IServiceCollection"/> so that additional calls can be chained.</returns>
+        public static IServiceCollection AddLogging(this IServiceCollection services, Action<ILoggerBuilder> configure)
```
khellang: 😍😍😍

src/Microsoft.Extensions.Logging/LoggerRuleSelector.cs
```diff
@@ -0,0 +1,93 @@
...
+                minLevel = options.MinLevel;
+            }
+        }
+
+        public static List<LoggerFilterRule> GetMatchingRules(LoggerFilterOptions options, string logger, string category)
+        {
+// Filter rule selection:
+            // 1. Select rules for current logger type, if there is none, select ones without logger type specified
+            // 2. Select rules with longest matching categories
+            // 3. If there no category
```
BrennanConroy: Something doesn't look right here.

src/Microsoft.Extensions.Logging/FilterLoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public static ILoggerBuilder AddFilter<T>(this ILoggerBuilder builder, string category, LogLevel level) where T: ILoggerProvider
+        {
+            return AddRule(builder, new LoggerFilterRule(typeof(T).FullName, category, level, null));
+        }
+
+        public static ILoggerBuilder AddFilter(this ILoggerBuilder builder, Func<string, string, LogLevel, bool> filter)
+        {
+            return AddRule(builder, new LoggerFilterRule(null, null, null, filter));
+        }
```
BrennanConroy: nit: newline

src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
```diff
@@ -19,13 +19,9 @@ public interface ILoggerFactory : IDisposable
         ILogger CreateLogger(string categoryName);
 
         /// <summary>
-        /// <para>
-        /// This method is obsolete and will be removed in a future version. The recommended alternative is to call the Microsoft.Extensions.Logging.AddProvider() method on the Microsoft.Extensions.Logging.LoggerFactory instance.
-        /// </para>
         /// Adds an <see cref="ILoggerProvider"/> to the logging system.
         /// </summary>
         /// <param name="provider">The <see cref="ILoggerProvider"/>.</param>
-        [Obsolete("This method is obsolete and will be removed in a future version. The recommended alternative is to call the Microsoft.Extensions.Logging.AddProvider() method on the Microsoft.Extensions.Logging.LoggerFactory instance.")]
```
BrennanConroy: Did we decide to un-obsolete everything?

pakrym: Yep, EF uses AddProvider and there is no time to fix it there.

src/Microsoft.Extensions.Logging.Console/ConfigurationConsoleLoggerConfigureOptions.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.Logging.Console
+{
+    public class ConfigurationConsoleLoggerConfigureOptions : ConfigureOptions<ConsoleLoggerOptions>
```
pakrym: I need suggestions :)

src/Microsoft.Extensions.Logging/ConfigurationLoggerFilterConfigureOptions.cs
```diff
@@ -0,0 +1,121 @@
...
+                    }
+                }
+            }
+        }
+
+        private void LoadRules(LoggerFilterOptions options, IConfigurationSection configurationSection, string logger)
+        {
+            foreach (var section in configurationSection.AsEnumerable(true))
+            {
+                if (TryGetSwitch(section.Value, out var level))
```
pakrym: Whoever tries to inject `ILogger` first would get it.

src/Microsoft.Extensions.Logging/LoggerBuilder.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Logging
+{
+    internal class LoggerBuilder : ILoggerBuilder
```
anurse: Pubternal?

src/Microsoft.Extensions.Logging/LoggerRuleSelector.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+
+        private static List<LoggerFilterRule> GetMatchingRules(LoggerFilterOptions options, string logger, string category)
+        {
+            // TODO: This can be rewritten to a single loop.
+            // Filter rule selection:
+            // 1. Select rules for current logger type, if there is none, select ones without logger type specified
+            // 2. Select rules with longest matching categories
+            // 3. If there no category
```
anurse: Did you forget to

src/Microsoft.Extensions.Logging.Abstractions/ILoggerFactory.cs
BrennanConroy: Obsolete doesn't mean it wont work, calling AddProvider should still be fine they just have to ignore the warning.

pakrym: Issues is there is no "right" way for them to do it right now, so showing obsolete warning is not entirely correct.

src/Microsoft.Extensions.Logging/LoggerBuilder.cs
pakrym: How do we decide? Other .Logging stuff was internal.

pakrym:
:up::date: I'd like to get it merged tomorrow

src/Microsoft.Extensions.Logging/LoggerBuilderExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    /// <summary>
+    /// Extension methods for setting up logging services in an <see cref="ILoggerBuilder" />.
+    /// </summary>
+    public static class LoggerBuilderExtensions
+    {
+        public static ILoggerBuilder AddConfiguration(this ILoggerBuilder builder, IConfiguration configuration)
+        {
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<LoggerFilterOptions>>(new LoggerFilterConfigureOptions(configuration)));
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IOptionsChangeTokenSource<LoggerFilterOptions>>(new ConfigurationChangeTokenSource<LoggerFilterOptions>(configuration)));
```
cwe1ss: Does the "TryAddEnumerable" mean that calling AddConfiguration a second time with a different configuration source is a no-op? If so, there would be quite a mix of behaviors here (IIUC):
* AddConfiguration can only be called once and further calls are ignored
* AddProvider can be called multiple times - the same provider can even be added multiple times
* SetMinimumLevel can be called multiple times and overwrites previous values

Not sure if it's a valid scenario to call "AddConfiguration" with different configuration sources but would it be a problem if it were allowed? The IConfigureOptions system can handle this anyway, right?

src/Microsoft.Extensions.Logging/LoggingServiceCollectionExtensions.cs
cwe1ss: Nitpick: Wouldn't `ILoggingBuilder` be better since the "Logging" module is configured with it and not a single logger. (just a suggestion/question, probably really not important)

src/Microsoft.Extensions.Logging/Microsoft.Extensions.Logging.csproj
```diff
@@ -17,6 +17,8 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(AspNetCoreVersion)" />
```
cwe1ss: "M.E.O.ConfigurationExtensions" brings in some of the above as transitive dependencies. I think it's more declarative to have those here as well - but not sure how you handle these things?!?

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs
```diff
@@ -12,68 +13,89 @@ public static class ConsoleLoggerExtensions
...
-            return factory;
+            builder.Services.AddSingleton<ILoggerProvider, ConsoleLoggerProvider>();
+
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds a console logger named 'Console' to the factory.
+        /// </summary>
+        /// <param name="builder">The <see cref="LoggerFactory"/> to use.</param>
```
cwe1ss: `The <see cref="LoggerFactory"/> to use.` is no longer a good comment here, is it? (applies to all of the ILoggerBuilder based methods)

src/Microsoft.Extensions.Logging/LoggerBuilder.cs
davidfowl: If we want to preserve the ability to new this up without an explicit service collection then it should be public and have a parameters less ctor that news up a service collection internally. But meh.

src/Microsoft.Extensions.Logging/Microsoft.Extensions.Logging.csproj
cwe1ss: I think the most important thing is for `M.E.L.Abstractions` to be clean because that's what libraries will depend on. `M.E.L` will be used by hosts and they have all of these dependencies anyway, so it's probably not necessary to split this?!

pakrym: @davidfowl Why do we have `ConfigurationExtensions` anyway?

pakrym: @davidfowl it should align with our `IDefaultConfigureOptions` story

cwe1ss:
Regarding the general issue of logging crashes before/on DI: We also tried to setup the whole logging system in program.cs before the WebHost is built. However we ended up still needing the environment name, some configuration values etc - so we would have duplicated a lot of WebHostBuilder logic and gave up in the end.

One thing that should still be possible though is to create your own mini DI-provider just for logging in Program.Main and use this one to catch startup exceptions (with CaptureStartupErrors=false). This one could then be environment-independent and log everything to e.g. Console/EventLog.

I'm also not really a friend of `EventSource`/`ETW`. It sounds great in theory but it's yet another thing to deal with and gets quite complex on Azure (ARM template -> Azure Diagnostics agent -> Azure Tables -> OMS). The new EventFlow library might help here but I haven't looked into it much yet.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -53,12 +54,16 @@ public IDisposable StartTestLog(ITestOutputHelper output, string className, out
...
-            var loggerFactory = new LoggerFactory();
-            if (output != null)
+            var serviceCollection = new ServiceCollection();
+            serviceCollection.AddLogging(builder =>
             {
-                loggerFactory.AddXunit(output, LogLevel.Debug);
-            }
+                if (output != null)
+                {
+                    builder.AddXunit(output, LogLevel.Debug);
```
BrennanConroy: Should we add a provider name to the `XunitLoggerProvider`?

src/Microsoft.Extensions.Logging.Console/ConfigurationConsoleLoggerConfigureOptions.cs
HaoK: ConfigureConsole[Logger]Options?

PR #637, opened 2017-06-14
Reduce dependencies of Azure AppServices logger

pakrym:
1. Split single Settings class into two options classes file/blob
2. Use Logging filters and options to control filter reload
3. Use options monitor feature to control if loggers and enabled


samples/SampleWeb/SampleWeb.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PreserveCompilationContext>false</PreserveCompilationContext>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging\Microsoft.Extensions.Logging.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging.Abstractions\Microsoft.Extensions.Logging.Abstractions.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging.Console\Microsoft.Extensions.Logging.Console.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging.Configuration\Microsoft.Extensions.Logging.Configuration.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging.AzureAppServices\Microsoft.Extensions.Logging.AzureAppServices.csproj" />
+    <PackageReference Include="Microsoft.AspNetCore" Version="2.0.0-preview3-*" />
```
natemcmaster: You can't use the meta package from a sample. It creates a cyclical repo dependency. :(

samples/SampleWeb/SampleWeb.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
+    <PreserveCompilationContext>false</PreserveCompilationContext>
```
natemcmaster: Why opting out of the default?

samples/SampleWeb/SampleWeb.csproj
```diff
@@ -0,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <RuntimeIdentifier>win7-x64</RuntimeIdentifier>
```
natemcmaster: xplat?

src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesLoggerFactoryExtensions.cs
```diff
@@ -53,10 +82,69 @@ public static ILoggerFactory AddAzureWebAppDiagnostics(this ILoggerFactory facto
         /// <param name="settings">The setting object to configure loggers.</param>
         public static ILoggerFactory AddAzureWebAppDiagnostics(this ILoggerFactory factory, AzureAppServicesDiagnosticsSettings settings)
         {
-            if (WebAppContext.Default.IsRunningInAzureWebApp)
+            var context = WebAppContext.Default;
+            if (context.IsRunningInAzureWebApp)
```
natemcmaster: Nit/personal preference: when a block wraps the entire contents of a method, invert the condition and return early. It reduces the nesting.

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/SiteConfigurationProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
+{
+    public class SiteConfigurationProvider
+    {
+        public static IConfiguration GetAzureLoggingConfiguration(IWebAppContext context)
+        {
+            var settingsFolder = Path.Combine(context.HomeFolder, "site", "diagnostics");
+            var settingsFile = Path.Combine(settingsFolder, "settings.json");
+
+            // TODO: This is a workaround because the file provider doesn't handle missing folders/files
```
natemcmaster: Is there an issue tracking this TODO?

test/Microsoft.Extensions.Logging.AzureAppServices.Test/AzureDiagnosticsConfigurationProviderTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            contextMock.SetupGet(c => c.HomeFolder)
+                .Returns(tempFolder);
+
+            var config = SiteConfigurationProvider.GetAzureLoggingConfiguration(contextMock.Object);
+
+            Assert.NotNull(config);
+        }
+
+        [Fact]
+        public void ReadsSettingsFileAndEnvironment()
```
natemcmaster: Would be good to add a test the ensure the ordering of env vs file is correct when there are matching config keys in both. settings.json takes precedence over environment, right?

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets a strictly positive value representing the maximum retained file count.
+        /// Defaults to <c>2</c>.
+        /// </summary>
+        public int? RetainedFileCountLimit
+        {
+            get { return _retainedFileCountLimit; }
```
natemcmaster: Nit: expression syntax. `get => _retainedFileCountLimit;`

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+                }
+                _fileSizeLimit = value;
+            }
+        }
+
+        /// <summary>
+        /// Gets or sets a strictly positive value representing the maximum retained file count.
+        /// Defaults to <c>2</c>.
+        /// </summary>
+        public int? RetainedFileCountLimit
```
natemcmaster: What would `null` represent? Should this be non-nullable?

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
```diff
@@ -0,0 +1,72 @@
...
+        private int? _fileSizeLimit = 10 * 1024 * 1024;
+        private int? _retainedFileCountLimit = 2;
+        private string _fileName = "diagnostics-";
+
+        /// <summary>
+        /// Gets or sets a strictly positive value representing the maximum log size in bytes.
+        /// Once the log is full, no more messages will be appended.
+        /// Defaults to <c>10MB</c>.
+        /// </summary>
+        public int? FileSizeLimit
```
natemcmaster: Is `null` valid?

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchLoggerConfigureOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+        public void Configure(BatchingLoggerOptions options)
+        {
+            options.IsEnabled = TextToBoolean(_configuration.GetSection(_isEnabledKey)?.Value);
+        }
+
+        private static bool TextToBoolean(string text)
+        {
+            bool result;
+            if (string.IsNullOrEmpty(text) ||
+                !bool.TryParse(text, out result))
```
natemcmaster: Nit : `out var result`

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLoggerOptions.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
+{
+    public class BatchingLoggerOptions
```
natemcmaster: As mentioned before, can you clarify if `null` is valid? If it isn't make these types non-nullable, or don't throw when `!value.HasValue`

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLoggerOptions.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Gets or sets the period after which logs will be flushed to the store.
+        /// </summary>
+        public TimeSpan FlushPeriod
+        {
+            get { return _flushPeriod; }
+            set
+            {
+                if (value < TimeSpan.Zero)
```
natemcmaster: `<=`. This will allow TimeSpan.Zero

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLoggerProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+        private BlockingCollection<LogMessage> _messageQueue;
+        private Task _outputTask;
+        private CancellationTokenSource _cancellationTokenSource;
+
+        protected BatchingLoggerProvider(IOptionsMonitor<BatchingLoggerOptions> options)
+        {
+            // NOTE: Only IsEnabled is monitored
+
+            var loggerOptions = options.CurrentValue;
+            if (loggerOptions.BatchSize < 0)
```
natemcmaster: Doesn't match the exception. Should be `<=`.

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLoggerProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+            {
+                Stop();
+            }
+        }
+
+        public ILogger CreateLogger(string categoryName)
+        {
+            return new BatchingLogger(this, categoryName);
+        }
+
```
natemcmaster: nit: extra line

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLoggerProvider.cs
```diff
@@ -0,0 +1,164 @@
...
+                    _currentBatch.Clear();
+                }
+
+                await IntervalAsync(_interval, _cancellationTokenSource.Token);
+            }
+        }
+
+        protected virtual Task IntervalAsync(TimeSpan interval, CancellationToken cancellationToken)
+        {
+            return Task.Delay(interval, cancellationToken);
```
natemcmaster: Nit/personal preference: expression body syntax can be used here to reduce this to one line. We aren't consistent on this, but IMO we should use it where possible.

This is true of several dozen places through out this PR, so I'll only mention it here.

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BlobLoggerProvider.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    foreach (var logEvent in eventGroup)
+                    {
+                        writer.Write(logEvent.Message);
+                    }
+                }
+            }
+        }
+
+        private (int Year, int Month, int Day, int Hour) GetBlobKey(LogMessage e)
```
natemcmaster: 👏  tuples ftw

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/FileLoggerProvider.cs
```diff
@@ -1,98 +1,80 @@
...
-            var loggerConfiguration = new LoggerConfiguration();
-            loggerConfiguration.WriteTo.Sink(backgroundSink);
-            loggerConfiguration.MinimumLevel.ControlledBy(new WebConfigurationReaderLevelSwitch(reader,
-                configuration =>
+                foreach (var item in files)
                 {
-                    return configuration.FileLoggingEnabled ? configuration.FileLoggingLevel : LogLevel.None;
-                }));
-            return loggerConfiguration.CreateLogger();
+                    item.Delete();
```
natemcmaster: Handle errors gracefully. All sorts of mayhem can happen here (file in use, security, etc).

src/Microsoft.Extensions.Logging.AzureAppServices/Microsoft.Extensions.Logging.AzureAppServices.csproj
```diff
@@ -6,20 +6,21 @@
...
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.Extensions.Logging\Microsoft.Extensions.Logging.csproj" />
+    <ProjectReference Include="..\Microsoft.Extensions.Logging.Configuration\Microsoft.Extensions.Logging.Configuration.csproj" />
     <ProjectReference Include="..\Microsoft.Extensions.Logging.Abstractions\Microsoft.Extensions.Logging.Abstractions.csproj" />
   </ItemGroup>
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="$(AspNetCoreVersion)" />
-    <PackageReference Include="Serilog.Extensions.Logging" Version="$(SerilogExtensionsLoggingVersion)" />
```
natemcmaster: Can the Serilog\*Version properties be removed from dependencies.props?

natemcmaster:
Can you clarify why we are removing Serilog? My concern is that serilog is a time-tested, hardened logging provider. We won't have time to release this as a 'beta' (preview, w/e) to get customer feedback on this major change if we merge this now.
cc @muratg

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/SiteConfigurationProvider.cs
pakrym: That's old TODO, I guess it was https://github.com/aspnet/Configuration/issues/501 I'll remove this code and re-test scenario.

test/Microsoft.Extensions.Logging.AzureAppServices.Test/AzureDiagnosticsConfigurationProviderTests.cs
pakrym: We get different options from different sources, they never intersect so we don't care about order.

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
pakrym: Yes, `null` - no limit

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
pakrym: `null` - no limit

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
pakrym: I don't like mixing two syntaxes in one property declaration.

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/FileLoggerProvider.cs
pakrym: `WriteMessagesAsync` invocation is wrapped into try/catch that suppresses all exception.

pakrym:
This package is part of azure light up package, we are trying to bring as few external dependencies as possible in it so we don't cause version conflicts.

src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
natemcmaster: Can you add this to the xmldocs?

natemcmaster:
FYI: test fails with compilation error
```
AssemblyTestLog.cs(132,158): error CS1739: The best overload for 'File' does not have a parameter named 'flushToDiskInterval' 
```

Issue #619, opened 2017-05-08
LogError with exception and empty message causes formatting issues with ConsoleProvider


agilenut:
When calling the overload of LogError that accepts an exception and a message...

_logger.LogError(
              eventId: 0, 
              exception: exception, 
              message:"");

... the log header is missing from the resulting log. i.e. The stack trace is logged but the log level, category, and event id are not. If any non-empty / non-null string is used, then it works correctly. 

This means that separate logs appear to run together which makes them difficult to read. Also, the color coding is missing since the log level is missing. This further decreases readability. 

You could make the argument that this is by design, but I believe it should always log the log header if anything is logged.

pakrym:
What version of Logging are you using? I tried in latest and see:
```
fail: SampleApp.Program[1]
      [null]
System.Exception: Boom
   at SampleApp.Program.Execute(String[] args) in D:\github\aspnet\Logging\samples\SampleApp\Program.cs:line 60
```

PR #639, opened 2017-06-16
Log message header in console logger when message is empty

pakrym:
Fixes https://github.com/aspnet/Logging/issues/619


test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -759,37 +770,56 @@ public void ConsoleLogger_Settings_LogLevelIgnoreCase()
...
+        [Theory]
+        [MemberData(nameof(LevelsWithPrefixes))]
+        public void WriteCore_EmptyMessageWithException(LogLevel level, string prefix)
+        {
+            // Arrange
+            var t = SetUp(null);
+            var logger = t.Item1;
+            var sink = t.Item2;
+            var ex = new Exception("Exception message" + Environment.NewLine + "with a second line");
+            string message = null;
```
BrennanConroy: `string.Empty`

Issue #580, opened 2017-03-26
Add baselines to enable checks for breaking changes


dougbu:
Two shipped projects in this repo lacks baselines. Looks like baseline files (`baseline.net45.json` and `baseline.netcore.json`) should be created in `rel/1.1.0-preview1` or `rel/1.1.1` and checked into at least `dev`. The details below are reformatted from a recent Universe build on the CI.
```
[Logging\src\Microsoft.Extensions.Logging.AzureAppServices\Microsoft.Extensions.Logging.AzureAppServices.csproj]
  No baseline file for netstandard1.3 found in Microsoft.Extensions.Logging.AzureAppServices.
[Logging\src\Microsoft.Extensions.Logging.EventSource\Microsoft.Extensions.Logging.EventSource.csproj]
  No baseline file for netstandard1.3 found in Microsoft.Extensions.Logging.EventSource.
```

pakrym:
https://github.com/aspnet/Logging/pull/650
https://github.com/aspnet/Logging/pull/641

PR #640, opened 2017-06-19
Add baselines for AzureAppServicesDiagnostics and EventSource

pakrym:
Fixes https://github.com/aspnet/Logging/issues/580


pakrym:
Hm, @BrennanConroy did this in dev already: https://github.com/aspnet/Logging/pull/600

PR #651, opened 2017-06-26
React to Options

HaoK:
@pakrym


HaoK:
Merging to unblock CI, will address any feedback after

PR #657, opened 2017-06-30
Fix #656 - issue parsing format string w/ group separator

### LINKED ISSUE #656: LogValuesFormatter throws exception when format string contains custom numeric format string with group separator
LogValuesFormatter.Format throws a `FormatException` with the message: "Input string was not in a correct format." when writing a log entry with a format string like the following:

    var setupTimer = new Stopwatch();
    // ...
    _logger.LogInformation("setup took {setupTime:#,#} ms", setupTimer.ElapsedMilliseconds);

(Note that the placeholder has a [custom numeric format string](https://docs.microsoft.com/en-us/dotnet/standard/base-types/custom-numeric-format-strings) with a group separator ",".)

Even though the exception gets thrown in the `Format` method (at the `return string.Format(...)` line), it appears that the actual issue is in the LogValuesFormatter constructor, which incorrectly parses the value name as "setupTime:#" rather than "setupTime" since it gets confused by the comma "," after the colon ":". It should detect that the colon ":" is the format delimiter since it comes before the comma ",".

Using Microsoft.Extensions.Logging.Abstractions 1.1.2.0 on Windows 7 SP1 (no IIS since it's being used in non-web application). It appears that the bug exists in the 2.0 preview releases as well.

mhenry07:
Fixes #656

Fix issue parsing log message format strings with placeholders containing a format string with a comma ",", such as "{setupTime:#,#}" which uses a custom numeric format string with a group separator. This fix uses the index of the first first comma "," or colon ":" found as the delimiter instead of ignoring the colon ":" if a comma "," was found.

### DIFF
--- a/test/Microsoft.Extensions.Logging.Test/FormattedLogValuesTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/FormattedLogValuesTest.cs
@@ -61,6 +61,18 @@ namespace Microsoft.Extensions.Logging.Test
         }
 
         [Theory]
+        [InlineData("8,765", "{Number:#,#}")]
+        public void LogValues_With_CustomNumericFormat(string expected, string format)
+        {
+            var number = 8765.4321;
+            var logValues = new FormattedLogValues(format, new object[] { number, number });
+            Assert.Equal(expected, logValues.ToString());
+
+            // Original format is expected to be returned from GetValues.
+            Assert.Equal(format, logValues.First(v => v.Key == "{OriginalFormat}").Value);
+        }
+
+        [Theory]
         [InlineData("{{", "{{", null)]
         [InlineData("'{{'", "'{{'", null)]
         [InlineData("'{{}}'", "'{{}}'", null)]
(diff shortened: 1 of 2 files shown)


test/Microsoft.Extensions.Logging.Test/FormattedLogValuesTest.cs
```diff
@@ -17,6 +17,7 @@ public class FormattedLogValuesTest
         [InlineData("arg1 arg2", "{Start} {End}", new object[] { "arg1", "arg2" })]
         [InlineData("arg1     arg2", "{Start,-6} {End,6}", new object[] { "arg1", "arg2" })]
         [InlineData("0064", "{Hex:X4}", new object[] { 100 })]
+        [InlineData("8,765", "{Number:#,#}", new object[] {8765.4321})]
```
pakrym: Please add `[InlineData(" 8,765", "{Number,6:#,#}", new object[] { 8765.4321 })]` test case and I'll merge it.

pakrym:
@mhenry07  Thank you!

mhenry07:
@pakrym Glad to be able to contribute. Thanks for reviewing this and getting it merged!

PR #660, opened 2017-07-04
Removed indirection in starting the consumer in ConsoleLoggerProcessor

gfoidl:
I see no need for the indirection to the static method, which in turn only calls the instance method.

```c#
public ConsoleLoggerProcessor()
{
	// Start Console message queue processor
	_outputTask = Task.Factory.StartNew(
		ProcessLogQueue,
		this,
		TaskCreationOptions.LongRunning);
}

private void ProcessLogQueue()
{
	foreach (var message in _messageQueue.GetConsumingEnumerable())
	{
		WriteMessage(message);
	}
}

private static void ProcessLogQueue(object state)
{
	var consoleLogger = (ConsoleLoggerProcessor)state;

	consoleLogger.ProcessLogQueue();
}
```

The (little) change uses the instance method directly in the `Task.Factory.StartNew` method.


muratg:
Assigning to @pakrym for review.

pakrym:
@gfoidl thank you!

Issue #661, opened 2017-07-04
ConsoleLogger should send LogLevel.Error to Error-Stream


gfoidl:
`LogError` and `LogCritical` should send the logmessages to `Console.Error`-stream. Right now the output is just sent to `Console.Out`.

### Minimal repro steps

```c#
namespace Foo
{
    class Program
    {
        static void Main(string[] args)
        {
            var loggerFactory = new LoggerFactory()
                .AddConsole();

            var logger = loggerFactory.CreateLogger<Program>();

            logger.LogInformation("Test");
            logger.LogError("Error");

            // Make sure the consumer on the write-side can drain the queue
            Thread.Sleep(500);
        }
    }
}
```

`dotnet foo.dll > out.txt 2> err.txt`

### Expected result
_err.txt_ should contain the logged error messaged. Critical messages, although not shown in the sample above, should also get sent to stderr.

### Actual result
All logged messages are sent to the standard output stream.

davidfowl:
Yea I don't know if I agree with this. I've never seen clear guidance on this. Can you provide some evidence that this is the norm?

gfoidl:
I agree that there's no norm -- at least I'm not aware of one -- to where each LogLevel should send its output. But for error-levels this should be the error-stream, 'cause this is what error-stream is for.
Cf. [Standard streams](https://en.wikipedia.org/wiki/Standard_streams#Standard_error_.28stderr.29)

The distinction between different (output) streams is useful -- especially on Linux -- for cronjobs, where one can send standard output to null, and error output to an email.

On execution and when standard-error is not redirected it is the same as standard-output. So nothing is won, nothing is lost. But we win the possibility to redirect the streams differently and more important: indepently.
Therefore I would make this the default, without any need to configure it explicitely.

davidfowl:
I don't agree but I'm willing to be wrong here. Can you show me other platform console loggers that switch based on error vs not. It feels inconvenient that you would need to pipe both streams to capture logs to a file. On top of that, you lose the overall ordering because you split the logs between streams

gfoidl:
> It feels inconvenient that you would need to pipe both streams to capture logs to a file.

[...]

PR #662, opened 2017-07-04
ConsoleLogger sends LogLevel.Error and above to Error-Stream (stderr)

gfoidl:
Fixes Issue #661


muratg:
Assigning to @pakrym for review.

davidfowl:
I'm still not convinced this should be the default. Lets look at some other platforms before we merge this.

gfoidl:
Closed due https://github.com/aspnet/Logging/issues/661#issuecomment-353965082

When the decision is to make it configurable, a new PR is the better choice I think.

PR #684, opened 2017-08-10
Add minimum LogLevel to AssemblyTestLog

BrennanConroy:
MusicStore and SignalR would need to be updated

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -33,9 +33,9 @@ namespace Microsoft.Extensions.Logging.Testing
             _assemblyName = assemblyName;
         }
 
-        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, [CallerMemberName] string testName = null)
+        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel = LogLevel.Debug, [CallerMemberName] string testName = null)
         {
-            var factory = CreateLoggerFactory(output, className, testName);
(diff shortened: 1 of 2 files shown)


muratg:
*ping*

anurse:
*pong*

jkotalik:
@BrennanConroy

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -33,9 +33,9 @@ private AssemblyTestLog(ILoggerFactory globalLoggerFactory, ILogger globalLogger
             _assemblyName = assemblyName;
         }
 
-        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, [CallerMemberName] string testName = null)
+        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel = LogLevel.Debug, [CallerMemberName] string testName = null)
```
JunTaoLuo: Don't we usually add overloads instead of using default parameters?

BrennanConroy:
```
public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel = LogLevel.Debug, [CallerMemberName] string testName = null);

public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, [CallerMemberName] string testName = null);

// which one does it call? Ambiguous
TestOutputHelper.StartTestLog("className", out var factory);
```

JunTaoLuo: Not ambiguous.

PR #693, opened 2017-08-23
Upgrade to xunit 2.3.0-beta4

natemcmaster:
Required a few changes due to xunit.analyzers. In particular, I changed one test that had an unused theory parameter.

### DIFF
--- a/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
@@ -168,7 +168,7 @@ namespace Microsoft.Extensions.Logging.Test
             logger.Log(LogLevel.Warning, 0, _state, null, _defaultFormatter);
 
             // Assert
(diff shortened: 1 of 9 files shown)


test/Microsoft.Extensions.Logging.Test/ConsoleLoggerTest.cs
```diff
@@ -923,6 +923,19 @@ public void ConsoleLoggerOptions_IncludeScopes_IsAppliedToLoggers()
             {LogLevel.Trace, "trce"},
         };
 
+        public static TheoryData<LogLevel> Levels
+        {
+            get
+            {
+                var data = new TheoryData<LogLevel>();
+                foreach (LogLevel value in Enum.GetValues(typeof(LogLevel)))
```
BrennanConroy: nit: var

natemcmaster: In this case, implicit variable typing fails because Enum.GetValues returns `System.Array`.

Issue #694, opened 2017-08-24
How FormattedLogValues is working ?


WernerMairl:
Hi
i see a unexpected behavior, not sure if i'm doing something wrong....
(using .net Core 2.0.0 and packages in version 2.0.0)

Assuming that the following extension method is working like string.Format!

`public static void LogTrace(this ILogger logger, string message, params object[] args)
`
the following sample should return 2 idenical trace strings!

```
string s = string.Format("{1} Bytes Markdown written to '{0}'", fn, markdown.Length);
this.Logger.LogTrace(s);
this.Logger.LogTrace("{1} Bytes Markdown written to '{0}'", fn, markdown.Length);
```

returns the following output on the Console-Logger (different strings)

trce: XXX[0]
      352 Bytes Markdown written to 'c:\xxx.md'
trce: XXX[0]
      c:\xxx.md Bytes Markdown written to '352'

This is a known issue/behavior ?
assuming that the problem comes from using format placeholder {0},{1} ordered in DESC not ASC
"{1} Bytess Markdown written to '{0}'"

internally the string calculation is done in the class FormattedLogValues, so i'm assuming that the behavior comes from there....

regards
Werner

davidfowl:
See https://messagetemplates.org/. Those aren't regular format strings. They are message templates. You should be using meaningful names in there.

WernerMairl:
David, 
that's a valid explanation but then i suggest to change comments on the Arguments => Usability!

```
        /// <summary>
        /// Formats and writes a trace log message.
        /// </summary>
        /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
        /// <param name="message">Format string of the log message.</param>
        /// <param name="args">An object array that contains zero or more objects to format.</param>
        public static void LogTrace(this ILogger logger, string message, params object[] args)
```
As a "very old" C# coder, everything in this Method declaration lets me assume that i can use string.format syntax!
The parameter "message" is commented with **Format string**

so please document (xml comments, Intellisense etc) that a different syntax then string.format must be used!!

it costs me a full day to find out why my tracing is producing wrong output....


regards
Werner

davidfowl:
@WernerMairl Sorry :disappointed: we'll improve both the documentation and the doc comments here.

/cc @pakrym @muratg @Rick-Anderson

pakrym:
Updated doc comments in #702, doc changes tracked in https://github.com/aspnet/Docs/issues/4093

PR #702, opened 2017-09-08
Add logging format string usage analyzers

pakrym:
Shows compiler diagnostics for:
1. Numerics in format string like `"{0} {1} {2}"`
2. Dynamic format strings `$" {something} "`, `" something " + 2`
3. Argument count doesn't match parameter count `LogError("{name} {value}", name)`
4. Supports cases optimized by compiler: `"string" + Const`, `"string" + "other string"`, `"string" + nameof(Thing)`

To prevent things like https://github.com/aspnet/Logging/issues/694 and other format misusages.
/cc @Eilon


Eilon:
Is this intended for our internal use? Or?

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Extensions.Logging.Abstractions\Microsoft.Extensions.Logging.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.3.0-beta1" />
```
Eilon: Should this version go to some version-settings-place-thingy?

pakrym: Yeah, would do it if we decide to take it.

pakrym:
My idea is that we ship it and everyone gets it when referencing metapackage.

Logging.sln
```diff
@@ -236,6 +240,30 @@ Global
 		{6D921637-507E-4CDC-8C5F-C3D6B62D118C}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 		{6D921637-507E-4CDC-8C5F-C3D6B62D118C}.Release|x86.ActiveCfg = Release|Any CPU
 		{6D921637-507E-4CDC-8C5F-C3D6B62D118C}.Release|x86.Build.0 = Release|Any CPU
+		{99DF369F-40A4-4088-8308-1C361B59DF4E}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{99DF369F-40A4-4088-8308-1C361B59DF4E}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{99DF369F-40A4-4088-8308-1C361B59DF4E}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
```
natemcmaster: Any particular reason we need x86 and mixed platform configurations?

test/Microsoft.Extensions.Logging.Analyzer.Test/DiagnosticVerifier.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <param name="diagnostics">The list of Diagnostics to be sorted</param>
+        /// <returns>An IEnumerable containing the Diagnostics in order of Location</returns>
+        private static Diagnostic[] SortDiagnostics(IEnumerable<Diagnostic> diagnostics)
+        {
+            return diagnostics.OrderBy(d => d.Location.SourceSpan.Start).ToArray();
+        }
+
+        #endregion
+
+        #region Set up compilation and documents
```
natemcmaster: 🙈 watch out for anti-region-legion members.

test/Microsoft.Extensions.Logging.Analyzer.Test/DiagnosticVerifier.cs
```diff
@@ -0,0 +1,147 @@
...
+            {
+                projects.Add(document.Project);
+            }
+
+            var diagnostics = new List<Diagnostic>();
+            foreach (var project in projects)
+            {
+                var compilationWithAnalyzers = project.GetCompilationAsync().Result.WithAnalyzers(ImmutableArray.Create(analyzer));
+
+                var diags = compilationWithAnalyzers.GetAllDiagnosticsAsync().Result;
```
natemcmaster: Add this to make sure tests fail if there are bugs in the analyzer itself:
```c#
Assert.DoesNotContain(diags, d => d.Id == "AD0001");
```

AD0001 has plagued early releases of xunit.analyzers, and was largely missed because the tests can pass if you're not checking for this error code.

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Analyzer package for Microsoft.Extensions.Logging.</Description>
```
natemcmaster: To make this work as an analyzer and not a compilation reference, we'll need to change the default package layout. The library needs to end up in analyzers/, not lib/
```xml
<BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
```

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
natemcmaster: PrivateAssets="All". This is an analyzer for analyzer authors, not consumers.

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <NoWarn>$(NoWarn);CS1591</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Extensions.Logging.Abstractions\Microsoft.Extensions.Logging.Abstractions.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.3.0-beta1" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" />
```
natemcmaster: Also PrivateAssets='All'

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Analyzer package for Microsoft.Extensions.Logging.</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
natemcmaster: Needs to be netstandard1.3 or lower so the analyzer works in Visual Studio's MSBuild. I don't think VS supports ns2.0 yet.

test/Microsoft.Extensions.Logging.Analyzer.Test/Microsoft.Extensions.Logging.Analyzer.Test.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
```
natemcmaster: Add .NET Framework 4.6 to match the runtime used by VS when executing your analyzer.

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,259 @@
...
+            });
+        }
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics { get; }
+
+        public override void Initialize(AnalysisContext context)
+        {
+            context.EnableConcurrentExecution();
+
+            context.RegisterCompilationStartAction(analysisContext =>
```
natemcmaster: Check if logging is even referenced. If not, skip the diagnostics altogether.

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            message = null;
+            arguments = null;
+            for (int i = 0; i < methodSymbol.Parameters.Length; i++)
+            {
+                var parameter = methodSymbol.Parameters[i];
+
+                if (parameter.Type.SpecialType == SpecialType.System_String &&
+                    parameter.Name == "message" ||
+                    parameter.Name == "messageFormat")
```
natemcmaster: We typically use .Equals with StringComparison.Ordinal for string comparisons

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
```diff
@@ -0,0 +1,259 @@
...
+        {
+            return invocation.Expression is IdentifierNameSyntax identifierName &&
+                   (identifierName.Identifier.IsKind(SyntaxKind.NameOfKeyword) ||
+                   identifierName.Identifier.ToString() == SyntaxFacts.GetText(SyntaxKind.NameOfKeyword));
+        }
+
+        private static IParameterSymbol DetermineParameter(
+            ArgumentSyntax argument,
+            SemanticModel semanticModel,
+            CancellationToken cancellationToken = default(CancellationToken))
```
natemcmaster: IMO default CT is a bad pattern. It's too easy to forget to pass in the token.

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
pakrym: Trying to find extensions type few lines lower does exactly this, I don't know if it can be done without having `Compilation`

pakrym:
:up::date:

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
natemcmaster: Yeah, you need compilation. The lines you added are fine.

src/Microsoft.Extensions.Logging.Analyzers/Microsoft.Extensions.Logging.Analyzers.csproj
```diff
@@ -2,17 +2,17 @@
...
     <Description>Analyzer package for Microsoft.Extensions.Logging.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
+    <BuildOutputTargetFolder>analyzers/dotnet/cs/</BuildOutputTargetFolder>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Microsoft.Extensions.Logging.Abstractions\Microsoft.Extensions.Logging.Abstractions.csproj" />
+    <Compile Include="..\Microsoft.Extensions.Logging.Abstractions\Internal\LogValuesFormatter.cs" />
```
natemcmaster: Good solution 👍

src/Microsoft.Extensions.Logging.Analyzers/Descriptors.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal class Descriptors
+    {
+        public static DiagnosticDescriptor MEL1NumericsInFormatString = new DiagnosticDescriptor(
+            "MEL1", "Numerics should not be used in logging format string",
+            "Numerics should not be used in logging format string", "Usage", DiagnosticSeverity.Info, true);
+
+        public static DiagnosticDescriptor MEL2ConcatenationInFormatString = new DiagnosticDescriptor(
+            "MEL2", "Logging format string should not be dynamically generated",
+            "Logging format string should not be dynamically generated", "Usage", DiagnosticSeverity.Info, true);
```
natemcmaster: How does VS display Info level diagnostics? Do you get a squiggly line or a suggestion light bulb? Most analyzers I've used issue warnings for everything.

pakrym: It shows the tiny dots in the beginning:

image

pakrym: I'm not sure if we want to fail compilation for people who do `WarningAsErrors` because this suggestion is mostly useful when you log a lot.

natemcmaster: Hmm, too bad that's really short. How hard would it be to set the diagnostic location for the entire formatted string? It makes the dotted underline a little more visible

nblumhardt:
Just a thought  - what do you think about driving this from the presence of a dynamically-identified attribute, say `[MessageTemplateFormatMethod]` like we do [here](https://github.com/serilog/serilog/blob/dev/src/Serilog/ILogger.cs#L342), so that the same analyzers would work for Serilog, NLog, LibLog, etc. in addition to MEL?

If we headed in that direction (I think @suchiman is supportive - https://github.com/Suchiman/SerilogAnalyzer/issues/15) - its possible we could combine our efforts and work on one set of analyzers that would work consistently everywhere.... which would be really awesome IMHO!

nblumhardt:
Thinking about this some more; a single set of analyzers might be a bit of a coordination bottleneck; if we managed to converge on a consistent way of annotating methods that accept message templates, a variety of tools could work with them (the same way that `[StringFormatMethod]` can be used to light up ReSharper's `string.Format` assistance on any method that its applied to).

src/Microsoft.Extensions.Logging.Analyzers/Descriptors.cs
anurse: I believe the triple-dots are a fixed size, regardless of how long the location is.

I also think it's still possible to promote info diagnostics to warnings or even errors. So even if we default to Info, those who really want to do things right can promote it as needed.

anurse: I think that's controlled through https://github.com/dotnet/roslyn/blob/master/docs/compilers/Rule%20Set%20Format.md

pakrym:
@nblumhardt only thing I'm worried about is clash of analyzers, for example when referencing Extensions.Logging with serilog provider, would we get a lot of duplicated diagnostics?

I still think that having common attribute is beneficial.

nblumhardt:
@pakrym I realised the same thing 👍 - the consistent attribute name and/or shape sounds more likely to have legs. Are you keen to explore/should we try to get another thread going on it?

PR #705, opened 2017-09-14
[WIP] Sketching "RecordMetric"

anurse:
Sketching out API ideas for adding a concept of logging "Metrics" to the logger.

This is starting to look like something we may want to merge, just wanted to let people look at this sketch as I played around with it.


anurse:
We could do something even sneakier and ensure Loggers "opt-in" to receiving metrics by having `Logger` try-cast the `ILogger`s it writes to to ensure they are able to handle Metrics. I'm going to sketch that next and see how it looks.

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -136,6 +136,44 @@ public IDisposable BeginScope<TState>(TState state)
...
+            var loggers = Loggers;
+            if (loggers == null)
+            {
+                return;
+            }
+
+            List<Exception> exceptions = null;
+            foreach (var loggerInfo in loggers)
+            {
+                // REVIEW: What "level" should metrics be at?
```
pakrym: Also whould we be able to define metric filters?

src/Microsoft.Extensions.Logging.Abstractions/Metric.cs
```diff
@@ -0,0 +1,14 @@
+namespace Microsoft.Extensions.Logging
+{
+    public struct Metric
```
pakrym: Is it worth having this?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMetricsExtensions.cs
```diff
@@ -0,0 +1,16 @@
+namespace Microsoft.Extensions.Logging
+{
+    public static class LoggerMetricsExtensions
+    {
+        public static void RecordMetric(this ILogger logger, string name, double value) => logger.RecordMetric(new Metric(name, value));
+
+        public static void RecordMetric(this ILogger logger, Metric metric)
```
pakrym: Do we need some high performance alternative like `DefineMessage` is for usual logging? For example to prevent the need to look up metric by name inside loggers for aggregation.

src/Microsoft.Extensions.Logging.Abstractions/IMetricLogger.cs
```diff
@@ -0,0 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Extensions.Logging
+{
+    public interface IMetricLogger
```
pakrym: 😍

pakrym: I would add something like `DefineMetric` that returns `Func<double>` this might improve speed of some implementations.

src/Microsoft.Extensions.Logging.Abstractions/Metric.cs
anurse: Maybe not.

anurse:
Talked with @pakrym . In order to ensure metrics are as low-cost as possible we're going to refactor from a `RecordMetric(string, double)` method to a `DefineMetric(string)` method which returns an `IMetric` that has a `RecordValue(double)` method on it. That way, the recording of metrics doesn't require a lookup in a dictionary.

src/Microsoft.Extensions.Logging/Logger.cs
anurse: Well, the level comment isn't really relevant in the new design, but yeah, we probably should support metric filters. The pattern of `DefineMetric(string)` can do that by returning a single-instance "black-hole" metric recorder.

anurse:
:up: :date: Moved to a Define/Record model. I also created a sample app but it's super hacky right now. I want to build something a little more useful. There's definitely some multi-threading code in here to be reviewed heavily as well ;)

anurse:
Closing this for now as I'm working on something a little more product-worthy now that we've sketched out a pattern.

PR #706, opened 2017-09-18
Add a way to get configuration section associated with logger provider

pakrym:
And fix `IncludeScopes`. Templates would need to be updated to move `IncludeScopes` to `Console` section.


src/Microsoft.Extensions.Logging.Configuration/LoggerProviderOptionsChangeTokenSource.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <inheritdoc />
+    public class LoggerProviderOptionsChangeTokenSource<TOptions, TProvider> : ConfigurationChangeTokenSource<TOptions>
```
pakrym: This is useful for provider implementations but not sure if it's worth having here.

src/Microsoft.Extensions.Logging/ProviderAliasUtilities.cs
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+
+namespace Microsoft.Extensions.Logging
+{
+    internal class ProviderAliasUtilities
+    {
+        private const string AliasAttibuteTypeFullName = "Microsoft.Extensions.Logging.ProviderAliasAttribute";
```
davidfowl: This is by name so that 3rd party providers don't have to take an unneeded dependency ?

src/Microsoft.Extensions.Logging.Configuration/LoggingConfiguration.cs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Extensions.Logging.Configuration
+{
+    internal class LoggingConfiguration
```
davidfowl: This is so we don't assume that we can just take `IE<IConfiguration>` right?

src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Extensions.Logging.Configuration
+{
+    internal class LoggerProviderConfigurationFactory : ILoggerProviderConfigurationFactory
+    {
+        private readonly IEnumerable<LoggingConfiguration> _configurations;
```
davidfowl: This is pretty weird. Usually configuration is composed at the provider level not by using multiple IConfiguration instances. Why do we do this?

src/Microsoft.Extensions.Logging/ProviderAliasUtilities.cs
pakrym: Some of them cannot build against aspnetcore 2.0 (AI)

src/Microsoft.Extensions.Logging.Configuration/LoggingConfiguration.cs
pakrym: Yep, I don't want random IConfiguration's in DI, especially if they were passed into logging.

src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationFactory.cs
pakrym: We allow composition of logging configurations for light-up scenarios (AI is adding another logging config file), so when you call `AddConfiguration` multiple times filtering rules get flattened in one list. I just simulated the same behavior for configuration sections.

Issue #708, opened 2017-09-19
Add support for reporting application metrics via logging


anurse:
## Overview

We want to enable applications to emit arbitrary named metrics and have them captured by telemetry/logging systems (App Insights, Elasticsearch/ELK, etc.) as well as other metrics systems (Statsd, InfluxDB, etc.). An abstraction is desired in order to allow applications to emit metrics without being coupled to the destination of the metrics, particularly with features like App Insights auto-lightup where the app doesn't directly depend upon App Insights.

## Proposal

We will add support for metrics reporting to `ILogger`. It will be done via a new companion interface `IMetricsLogger`, which loggers may **optionally** implement. Applications can emit metrics through this interface (see API below), and logger providers which support metrics can receive and process those metrics. Metrics are `(name, value)` tuples, where `value` is always floating point number. Metrics are expected to be viewed in aggregate, hence it doesn't make sense to use a generic data type like `object` for the metric, so metric values are always `double`s. Supporting arbitrary types of metric values is a non-goal. Reporting metrics should be as efficient as possible, especially in the case where no logger provider is registered to listen for them, or they are filtered out.

### Why add to `ILogger`?

Initially, I considered developing a new abstraction for Metrics. However, I encountered a few concerns:

1. We would need to develop a structure for naming/organizing metrics (likely using type/namespace names as a base)
2. We would need to develop a structure for filtering metrics (by categories, names, granularities, etc.)
3. Many telemetry systems provide support for both log and metric collection, and would have to provide a new system
4. Consumers have to pull in two different diagnostics reporters from DI (`ILogger<T>`, `IMetricReporter<T>`, etc.)
5. This would create Yet Another Diagnostics Abstraction (on top of EventSource, EventCounters, DiagnosticSource, Perf Counters, yada yada yada (see what I did there? ;P))

The fact that 1 and 2 above already exist in the Logging pipeline, and many metric sinks are also logging sinks, led to the idea of integrating metrics into Logging rather than creating a completely new abstraction.

The main reason why we would **not** want to add this to `ILogger` would be a risk

### `IMetricLogger`

Optional companion interface for `ILogger` implementations to indicate their support for metrics. The `ILogger` instances returned by `ILoggerProvider` are expected to implement this interface *if and only if* they wish to receive metrics.

```csharp
namespace Microsoft.Extensions.Logging
{
    public interface IMetricLogger
    {
        /// <summary>
[...]

PR #709, opened 2017-09-19
Initial pass at Metrics support in ILogger

anurse:
This implements #708. It's got some naive elements, like the way we handle filters (which involves some locking). I want to clean that up but wanted to get this into a PR first.


src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -136,6 +139,21 @@ public IDisposable BeginScope<TState>(TState state)
             return scope;
         }
 
+        public IMetric DefineMetric(string name)
+        {
+            if (!_metrics.TryGetValue(name, out var metric))
```
pakrym: Is dictionary safe for multi thread read during the write?

src/Microsoft.Extensions.Logging/Microsoft.Extensions.Logging.csproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
     <PackageReference Include="Microsoft.Extensions.Options" />
+    <PackageReference Include="System.Collections.Immutable" />
```
pakrym: Is it still used?

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -197,5 +215,87 @@ public void Dispose()
...
+            private IMetric[] _metrics;
+            private object _sync = new object();
+
+            public Metric(Logger logger, string name)
+            {
+                _logger = logger;
+                _name = name;
+            }
+
+            public void RecordValue(double value)
```
pakrym: As we discussed this implementation has to be optimized to update metrics array only when required.

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -197,5 +215,87 @@ public void Dispose()
...
+                    // REVIEW: LogLevel for metrics?
+                    // REVIEW: Filtering by Metric Name?
+                    if (_logger.Loggers[i].IsEnabled(LogLevel.Critical))
+                    {
+                        metrics[i].RecordValue(value);
+                    }
+                }
+            }
+
+            public void RecordValue<T>(double value, T properties) where T : IEnumerable<KeyValuePair<string, object>>
```
pakrym: Test for this and `TimeSpan` extension method.

src/Microsoft.Extensions.Logging/Logger.cs
anurse: I'm honestly not sure. I was having the same concern and thinking I should just use ConcurrentDictionary

src/Microsoft.Extensions.Logging/Microsoft.Extensions.Logging.csproj
anurse: Not anymore, will remove

anurse:
Closing this, we're going to look at using [EventCounters](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.tracing.eventcounter?view=netcore-2.0) for our metrics needs.

PR #710, opened 2017-09-19
Use xunit runner config only for analyzers

pakrym:
Having it globally causes issues with `System.Net.Http`


natemcmaster:
Ah, yes this totally makes sense. Good find.

Issue #714, opened 2017-09-23
Don't set AsyncLocal for scope if not included


benaadams:
Adding ConsoleLogger disturbs the `ExecutionContext` even if scopes aren't being used and causes allocations to wildly increase.

Without ConsoleLogger (no-changes to AsyncLocal)
![](https://aoa.blob.core.windows.net/aspnet/tasks-action.png)

With ConsoleLogger and no scopes - but not logging anything to Console (changes to AsyncLocal)
![](https://aoa.blob.core.windows.net/aspnet/async-local.png)

You can see the allocations of `Action` and `MoveNextRunner` disproportionately increase

benaadams:
@pakrym resolved by https://github.com/aspnet/Logging/pull/723?

pakrym:
Nope, It's made harder to resolve by that PR. Central storage doesn't know if anyone is planning to log scope values so they have to be collected all the time.

stephentoub:
@benaadams, I'm curious how those screenshots change using 2.1?

benaadams:
274,597 requests (with asynclocals); Action has lost its top spot

![image](https://user-images.githubusercontent.com/1142958/35487896-2ae6e9e8-0479-11e8-80ec-74d0eb1e2e4e.png)

stephentoub:
Great, thanks, glad to see the async builder improvements helped.

benaadams:
374,490 requests without logging (no asynclocals)
![image](https://user-images.githubusercontent.com/1142958/35488209-be804f7a-047c-11e8-898d-9ab4bd090a61.png)

stephentoub:
Right, in 2.1 Action allocations should no longer be impacted by AsyncLocals.

benaadams:
Backtracing Action most aren't from AsyncMehodBuilder anymore; the are from [`LoggingThreadPool.Schedule(Action<object> action, object state)`](https://github.com/aspnet/KestrelHttpServer/blob/bbe8851f32984264282d80ea1a0fe4c70cbffa95/src/Kestrel.Core/Internal/Infrastructure/LoggingThreadPool.cs#L58-L62)

![image](https://user-images.githubusercontent.com/1142958/35488289-00746302-047e-11e8-9270-a2828948ff93.png)

davidfowl:
With the new design of the external scope provider, this is actually much harder to achieve.

davidfowl:
@benaadams if you want to take a look, have at it!

snakefoot:
[...]

PR #715, opened 2017-09-23
Don't set AsyncLocal for scope if not included

benaadams:
Don't disturb the `ExecutionContext` if scopes aren't being used as it wildly increases allocations

Without ConsoleLogger (no changes to AsyncLocal)

With ConsoleLogger and no scopes - but not logging anything (changes to AsyncLocal)

You can see the allocations of `Action` and `MoveNextRunner` disproportionately increase; prior to this change

Resolves https://github.com/aspnet/Logging/issues/714

/cc @davidfowl 

You are right @anurse, logging allocations make me sad 😢

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -199,7 +199,14 @@ namespace Microsoft.Extensions.Logging.Console
                 throw new ArgumentNullException(nameof(state));
             }
 
-            return ConsoleLogScope.Push(Name, state);
+            if (IncludeScopes)
+            {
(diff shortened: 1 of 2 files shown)


BrennanConroy:
Lets say you enable scopes after you're already in a scope, you now lose information for the logs in that scope until you leave and renter it. Not sure how important of a scenario that is though.

benaadams:
> Lets say you enable scopes after you're already in a scope, you now lose information for the logs in that scope until you leave and renter it. Not sure how important of a scenario that is though.

The start events won't be in a scope either

anurse:
Yeah, in that scenario, in-progress scopes are already messed up anyway. I don't think optimizations like this would cause any more problems.

benaadams:
I think this has been resolved by resolved by #723?

PR #719, opened 2017-10-04
Use TryAddEnumerable to prevernt adding logger provider multiple times

pakrym:
Fixes: https://github.com/aspnet/MetaPackages/issues/162

To prevent duplicated logs if people call `builder.Add*` multiple times


BrennanConroy:
Add test

pakrym:
Ping

pakrym:
@davidfowl what do you mean by composite?

davidfowl:
Another other loggers that use these loggers as an implementation detail?

PR #723, opened 2017-10-25
External scope storrage prototype

pakrym:
Goal: every logger should have scope support but it should be done in efficient way

Solution: provide opt in mechanism for logger providers to indicate that they can consume external scope and allow `ILoggerFactory` keep an single storage for all loggers.

When logger provider implements `ISupportsExternalScope` `LoggerFactory` would provider it's implementation of `IExternalScopeProvider` to it and never call `StartScope` in the future. In case implementation of `ILoggerFactory` does not support `IExternalScopeProvider` logger provider is able to instantiate and use `LoggerExternalScopeProvider` on it's own.

`CollectScope<T>(Action<object, T> callback, T state)` is implemented as a callback instead of returning `IEnumerable` to provider allocation free way of consuming scope values without exposing internal implementation details.

Current `IExternalScopeProvider` implementation is subject to change too.

@nblumhardt

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
@@ -1,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// Represents a storage of common scope data.
+    /// </summary>
+    public interface IExternalScopeProvider
+    {
+        /// <summary>
+        /// Executes callback for each currently active scope object from most to least recently added.
+        /// All callbacks are guaranteed to be called inline from this method.
+        /// </summary>
+        /// <param name="callback">The callback to be executed for every scope object</param>
+        /// <param name="state">The state object to be passed into the callback</param>
+        /// <typeparam name="T"></typeparam>
+        void CollectScope<T>(Action<object, T> callback, T state);
+
+        /// <summary>
+        /// Adds scope object to the list
+        /// </summary>
+        /// <param name="state">The scope object</param>
+        /// <returns>The <see cref="IDisposable"/> token that removes scope on dispose.</returns>
+        IDisposable Push(object state);
+    }
+}
--- a/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
@@ -16,8 +16,8 @@ namespace Microsoft.Extensions.Logging
         /// </summary>
         /// <param name="callback">The callback to be executed for every scope object</param>
         /// <param name="state">The state object to be passed into the callback</param>
-        /// <typeparam name="T"></typeparam>
-        void CollectScope<T>(Action<object, T> callback, T state);
+        /// <typeparam name="TState"></typeparam>
+        void ForEachScope<TState>(Action<object, TState> callback, TState state);
 
         /// <summary>
         /// Adds scope object to the list
--- a/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Extensions.Logging
     public interface IExternalScopeProvider
     {
         /// <summary>
-        /// Executes callback for each currently active scope object from most to least recently added.
+        /// Executes callback for each currently active scope objects in order of creation.
         /// All callbacks are guaranteed to be called inline from this method.
         /// </summary>
         /// <param name="callback">The callback to be executed for every scope object</param>
--- a/src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
@@ -1,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// An interface for configuring logging providers.
+    /// </summary>
+    public interface ISupportsExternalScope
+    {
+        void SetScopeProvider(IExternalScopeProvider scopeProvider);
+    }
+
+    public interface IExternalScopeProvider
+    {
+        void CollectScope<T>(Action<object, T> callback, T state);
+        IDisposable Push(object state);
+    }
+}
--- a/src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
@@ -1,21 +1,17 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
-using System;
-
 namespace Microsoft.Extensions.Logging
 {
     /// <summary>
-    /// An interface for configuring logging providers.
+    /// Represents a <see cref="ILoggerProvider"/> that is able to consume external scope information.
     /// </summary>
     public interface ISupportsExternalScope
     {
+        /// <summary>
+        /// Sets external scope information source for logger provider.
+        /// </summary>
+        /// <param name="scopeProvider"></param>
         void SetScopeProvider(IExternalScopeProvider scopeProvider);
-    }
-
-    public interface IExternalScopeProvider
-    {
-        void CollectScope<T>(Action<object, T> callback, T state);
-        IDisposable Push(object state);
     }
 }
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
@@ -10,10 +10,14 @@ using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Extensions.Logging
 {
+    /// <summary>
+    /// Default implemenation of <see cref="IExternalScopeProvider"/>
+    /// </summary>
     public class LoggerExternalScopeProvider : IExternalScopeProvider
     {
         private readonly AsyncLocal<Scope> _currentScope = new AsyncLocal<Scope>();
 
+        /// <inheritdoc />
         public void CollectScope<T>(Action<object, T> callback, T state)
         {
             var current = _currentScope.Value;
@@ -24,6 +28,7 @@ namespace Microsoft.Extensions.Logging
             }
         }
 
+        /// <inheritdoc />
         public IDisposable Push(object state)
         {
             var parent = _currentScope.Value;
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
@@ -20,12 +20,16 @@ namespace Microsoft.Extensions.Logging
         /// <inheritdoc />
         public void ForEachScope<TState>(Action<object, TState> callback, TState state)
         {
-            var current = _currentScope.Value;
-            while (current != null)
+            void Report(Scope current)
             {
+                if (current == null)
+                {
+                    return;
+                }
+                Report(current.Parent);
                 callback(current.State, state);
-                current = current.Parent;
             }
+            Report(_currentScope.Value);
         }
 
         /// <inheritdoc />
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
@@ -45,6 +45,7 @@ namespace Microsoft.Extensions.Logging
         private class Scope : IDisposable
         {
             private readonly LoggerExternalScopeProvider _provider;
+            private bool _isDisposed;
 
             internal Scope(LoggerExternalScopeProvider provider, object state, Scope parent)
             {
@@ -64,7 +65,11 @@ namespace Microsoft.Extensions.Logging
 
             public void Dispose()
             {
-                _provider._currentScope.Value = Parent;
+                if (!_isDisposed)
+                {
+                    _provider._currentScope.Value = Parent;
+                    _isDisposed = true;
+                }
             }
         }
     }
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Extensions.Logging
         private readonly AsyncLocal<Scope> _currentScope = new AsyncLocal<Scope>();
 
         /// <inheritdoc />
-        public void CollectScope<T>(Action<object, T> callback, T state)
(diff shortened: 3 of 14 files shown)


src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace Microsoft.Extensions.Logging.Console
 {
+    [Obsolete("Use " + nameof(LoggerExternalScopeProvider) + "instead" )]
```
anurse: Why was this ever public?

anurse: Ok then :)

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -86,10 +91,13 @@ public IConsole Console
             }
         }
 
-        public bool IncludeScopes { get; set; }
-
         public string Name { get; }
 
+        public IExternalScopeProvider ScopeProvider { get; set; }
+
+        [Obsolete("Changing this property would have no effect, please use " + nameof(ConsoleLoggerOptions) + "." + nameof(ConsoleLoggerOptions.IncludeScopes))]
```
anurse: "Changing this property has no effect. Use " + ... + " instead"

NinoFloris:
What about the returned IDisposable instance for each new scope? What's the perf impact of that?

pakrym:
@NinoFloris this PR would alow to reduce overall count of `IDisposable`s because we woldn't get one per provider per scope.

Future optimization is possible if all configured logger providers support external scope we can return IDisposable from `IExternalScopeProvider` without wrapping it in `Logger.Scope`

src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
```diff
@@ -248,28 +246,18 @@ private ConsoleColors GetLogLevelConsoleColors(LogLevel logLevel)
...
+            if (scopeProvider != null)
             {
-                if (length == builder.Length)
-                {
-                    scopeLog = $"=> {current}";
-                }
-                else
+                var length = builder.Length;
+
+                ScopeProvider?.CollectScope((scope, stringBuilder) => stringBuilder.Append("=> ").Append(scope), builder);
```
anurse: You capture `scopeProvider` above, then use `ScopeProvider` here. Mistake?

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -127,6 +153,12 @@ public IDisposable BeginScope<TState>(TState state)
                 }
             }
 
+            if (_loggerFactory.ScopeProvider != null)
+            {
+                scope.SetDisposable(0, _loggerFactory.ScopeProvider.Push(state));
+            }
+
+
```
anurse: nit: blank line

src/Microsoft.Extensions.Logging/LoggerExternalScopeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+namespace Microsoft.Extensions.Logging
+{
+    public class LoggerExternalScopeProvider : IExternalScopeProvider
+    {
+        private readonly AsyncLocal<Scope> _currentScope = new AsyncLocal<Scope>();
+
+        public void CollectScope<T>(Action<object, T> callback, T state)
+        {
+            var curent = _currentScope.Value;
```
anurse: nit: `current` (two `r`s)

src/Microsoft.Extensions.Logging/LoggerExternalScopeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Extensions.Logging
+{
+    public class LoggerExternalScopeProvider : IExternalScopeProvider
+    {
+        private readonly AsyncLocal<Scope> _currentScope = new AsyncLocal<Scope>();
+
+        public void CollectScope<T>(Action<object, T> callback, T state)
```
anurse: Put a doc comment on here describing how the walk is done (from leaf nodes up to the root). Doesn't have to be amazing prose, but something that makes the behavior relatively clear for us to wordsmith later (and to help provider authors)

nblumhardt:
@pakrym thanks for the heads-up!

I think Serilog is unaffected because, by integrating at the `ILoggerFactory` level, we're working at a layer above this infrastructure. (Please correct me if I'm wrong.)

Integrating this into the Seq provider seems straightforward - it's using a cut-down copy of Serilog's infrastructure, so effectively we'd call `CollectScope()` within an `ILogEventEnricher` and apply the logic that's in https://github.com/datalust/seq-extensions-logging/blob/dev/src/Seq.Extensions.Logging/Serilog/Extensions/Logging/SerilogLoggerScope.cs#L48.

One thing that doesn't quite sit right with me about the way MEL is evolving is the disconnect between scopes, events, and filters in the pipeline. In Serilog, properties from the scope go onto the event; later, filters can determine whether an event is included based on its properties, and we can even do tricky things like enrich or route events selectively based on their properties.

Events being a collection of properties is such a fundamental structured logging concept that it's strange to flesh out the implementation side of MEL without acknowledging it in the design.

I suspect that the ultimate reason for avoiding an "event" type was to eliminate mandatory allocations, and keeping it out of the MEL interface/abstraction was a great move, but with each tactical feature addition like this one, allocations appear in real-world use of the pipeline anyway. (There's a likely per-event allocation for the first delegate argument to `CollectScope()`, for example.)

In the long run, would `struct LogEvent` and some kind of low-footprint `Properties` collection on it, unify the implementation-side of MEL and help connect all of these features logically?

nblumhardt:
(In case the relevance of the `LogEvent` topic is unclear - the alternative to the design explored in the PR is to have the scope set properties on a `LogEvent` for consumption by the providers, instead of having providers pull information from the scope directly.)

pakrym:
@nblumhardt unfortunately we are not able to make large design changes at this point so what we left with is trying to improve current situation with targeted fixes.

>There's a likely per-event allocation for the first delegate argument to CollectScope(), for example.
My hope was that having `T state` would alow to prevent allocation.

nblumhardt:
@pakrym under those constraints it's fine - I am only thinking of how each API added tactically will make larger-scale changes trickier in the future (always that catch-22).

The `Action` is the object that will be allocated - at least in the case of the console provider, the closure will end up on the heap. No big deal.

The `T state` argument makes sense 👍. Naming did trip me up at first - scanning the code I read it as `TState state`. It's consistent with other callbacks like `ParameterizedThreadStart` though.

anurse:
> The Action is the object that will be allocated - at least in the case of the console provider, the closure will end up on the heap. No big deal.

Even less deal if the Action only depends on the `state` parameter, which means there'd only be one per process (hence why that pattern appears everywhere :))

> instead of having providers pull information from the scope directly

This PR is kinda trying to move us towards that model in a compatible way. I completely agree that the model for this should really just be that scope properties are aggregated with the log event properties and logger providers never really need to see scopes. This model allows us to start migrating providers to a model where they don't do scope management at all.

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// An interface for configuring logging providers.
+    /// </summary>
+    public interface ISupportsExternalScope
+    {
+        void SetScopeProvider(IExternalScopeProvider scopeProvider);
+    }
+
+    public interface IExternalScopeProvider
+    {
+        void CollectScope<T>(Action<object, T> callback, T state);
```
halter73: Nit: I prefer AggregateScopes or ReduceScopes.

Maybe `TAccumulate AggregateScopes<TAccumulate>(Function<TAccumulate, object, TAccumulate>`. It should be easy to convert any `Action<object, T> callback` into this by just returning the T/TAccumlate.

I actually don't care to much if the method is void or not, but putting Aggregate or Reduce in the name makes it more clear that the callback is called potentially multiple times inline.

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
```diff
@@ -0,0 +1,21 @@
...
+        IDisposable Push(object state);
```
halter73: Since BeginScope takes `TState state`, shouldn't this too?

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -107,14 +114,33 @@ public IDisposable BeginScope<TState>(TState state)
                 return loggers[0].Logger.BeginScope(state);
             }
 
-            var scope = new Scope(loggers.Length);
+            var scopeProvider = _loggerFactory.ScopeProvider;
+            var scopeCount = scopeProvider != null ? 1 : 0;
+
+            // TODO: do not do this every time
```
halter73: Are you going to cache the result for this?

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.Extensions.Logging
+{
+    /// <summary>
+    /// An interface for configuring logging providers.
+    /// </summary>
+    public interface ISupportsExternalScope
```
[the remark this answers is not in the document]
anurse: I was going to write a response about the `I` not being part of the name and we should keep the `s`, but then I found this: https://source.dot.net/#System.ComponentModel.Primitives/System/ComponentModel/ISupportInitialize.cs,f5af3a8671036626 ;P

So, yeah, carry on then.

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
anurse: `AggregateScopes` is fine with me if it lets you flow an accumulator, to match `Enumerable.Aggregate`. I'm not a fan of `ReduceScopes` because `Reduce` isn't a term we generally use in our APIs (even if I do like it :)).

I agree `Collect` isn't a great word since it implies collecting into a single unit (kinda like `Aggregate`). Perhaps just `ForEachScope` (precedent with `List<T>.ForEach`) or `EnumerateScopes` (other BCL examples of this return `IEnumerable` but I think we wouldn't want to do that here, however it would be pretty clear what the callback is doing when you see the `void` return).

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
anurse: `Push<TState>(TState state)`?

This is one place where the external scope provider isn't able to have the advantages that each provider has. We preserve generics all the way through in order to allow `structs` to remain unboxed, but the external provider will be forced to box them in order to store them for later.

src/Microsoft.Extensions.Logging.Abstractions/ISupportsExternalScope.cs
pakrym: I like `ForEach` more, don't think `Aggregate` pattern would be useful in this case: `TAccumulate` would often be collection or collection like object and passed in by reference.

src/Microsoft.Extensions.Logging.Abstractions/IExternalScopeProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+    public interface IExternalScopeProvider
+    {
+        /// <summary>
+        /// Executes callback for each currently active scope object from most to least recently added.
+        /// All callbacks are guaranteed to be called inline from this method.
+        /// </summary>
+        /// <param name="callback">The callback to be executed for every scope object</param>
+        /// <param name="state">The state object to be passed into the callback</param>
+        /// <typeparam name="T"></typeparam>
+        void CollectScope<T>(Action<object, T> callback, T state);
```
davidfowl: nit: T -> TState

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -34,7 +36,11 @@ public LoggerFactory(IEnumerable<ILoggerProvider> providers, LoggerFilterOptions
 
         public LoggerFactory(IEnumerable<ILoggerProvider> providers, IOptionsMonitor<LoggerFilterOptions> filterOption)
         {
-            _providerRegistrations = providers.Select(provider => new ProviderRegistration { Provider = provider }).ToList();
+            foreach (var provider in providers)
+            {
+                AddProviderRegistration(provider, false);
```
davidfowl: nit: use a named parameter for the false argument.

src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+        public IDisposable Push(object state)
+        {
+            var parent = _currentScope.Value;
+            var newScope = new Scope(this, state, parent);
+            _currentScope.Value = newScope;
+
+            return newScope;
+        }
+
+        private class Scope: IDisposable
```
davidfowl: nit: spacing

pakrym:
:up::date:

src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <inheritdoc />
+        public void ForEachScope<TState>(Action<object, TState> callback, TState state)
+        {
+            void Report(Scope current)
+            {
+                if (current == null)
+                {
+                    return;
+                }
+                Report(current.Parent);
```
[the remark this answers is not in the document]
pakrym: Not without additional data structure, I'm using recursion to reverse order of items in collection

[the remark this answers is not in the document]
anurse: Ah, I see. Didn't notice the position of the recursion was above the consumption. Ok then.

src/Microsoft.Extensions.Logging.Abstractions/LoggerExternalScopeProvider.cs
```diff
@@ -0,0 +1,76 @@
...
+                Parent = parent;
+            }
+
+            public Scope Parent { get; }
+
+            public object State { get; }
+
+            public override string ToString()
+            {
+                return State?.ToString();
```
halter73: Nit: Can State really be null? If so, is it OK to return null from ToString()?

pakrym: Copied from https://github.com/aspnet/Logging/blob/e56850141e1c538b0eccedfcb85980f9f74e47d4/src/Microsoft.Extensions.Logging.Console/ConsoleLogScope.cs#L47 but not sure how valid it is.

kierenj:
very cool, am (re-)implementing scope at the moment for a logger and of course this would be a plus. is there any kind of timeline for this appearing in a particular version?

snakefoot:
@pakrym Any reason why you removed this optimization:

```
if (loggers.Length == 1)
    return loggers[0].Logger.BeginScope(state);
```

It will hurt NLog and Log4net that doesn't replace the LoggerFactory like Serilog does.

davidfowl:
@snakefoot I'm more worried about breaking changes then the "optimization". Can you file a new bug describing the problem NLog is having with a repro project? That'll help us quickly identify the issue.

snakefoot:
@pakrym It would be nice if LoggerProviders could still be notified on BeginScope / DisposeScope. In case they had extra state to maintain without having to enter allocation frenzy.

davidfowl:
> @pakrym It would be nice if LoggerProviders could still be notified on BeginScope / DisposeScope. In case they had extra state to maintain without having to enter allocation frenzy.

Can you please file an issue for each of the problems you're having (with evidence)...

snakefoot:
@davidfowl Created #893

snakefoot:
@davidfowl Created #894

Issue #722, opened 2017-10-25
Console logger settings are not reloaded if section didn't exist on app start


pakrym:
Repro:
1.
```
{
  "Logging": {
    "LogLevel": {
      "Default": "Debug",
      "System": "Information",
      "Microsoft": "Information"
    }
  }
}
```

2. Start app
3.
```
{
  "Logging": {
    "LogLevel": {
      "Default": "Debug",
      "System": "Information",
      "Microsoft": "Information"
    },
    "Console":
    {
      "IncludeScopes": "true"
    }
  }
}
```

Scopes wouldn't appear.

PR #725, opened 2017-10-26
Fix logger provider configuration reload after adding section

pakrym:
https://github.com/aspnet/Logging/issues/722


src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationFactory.cs
```diff
@@ -30,18 +30,12 @@ public IConfiguration GetConfiguration(Type providerType)
             foreach (var configuration in _configurations)
             {
                 var sectionFromFullName = configuration.Configuration.GetSection(fullName);
-                if (sectionFromFullName.Exists())
```
muratg: Does the Exists method return false in some cases where it shouldn't?

pakrym: Even if section doesn't exist we need to add it to configuration builder so event would fire when it appears.

PR #729, opened 2017-11-10
Adding extensions to defer calculation of the log message

drungrin:
Logging can add quite a overhead if you need to put expensive information on the message, or if you ommit them may affect the purpose of the message.

Spreading conditionals based on the level can bloat the code easily.

Using a lambda, the expensive operations don't affect the production code and solve the above issues.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
@@ -89,6 +89,106 @@ namespace Microsoft.Extensions.Logging
             logger.Log(LogLevel.Debug, 0, new FormattedLogValues(message, args), null, _messageFormatter);
         }
 
+        /// <summary>
+        /// Formats and writes a debug log message.
+        /// </summary>
+        /// <param name="logger">The <see cref="ILogger"/> to write to.</param>
+        /// <param name="eventId">The event id associated with the log.</param>
+        /// <param name="exception">The exception to log.</param>
+        /// <param name="messageGenerator">A function returning message to be written. Function is not evaluated if logging level is not enabled</param>
+        /// <example>logger.LogDebug(0, exception, () => "Error while processing request from")</example>
+        public static void LogDebug(this ILogger logger, EventId eventId, Exception exception, Func<string> messageGenerator)
+        {
+            if (logger == null)
+            {
+                throw new ArgumentNullException(nameof(logger));
+            }
+
+            if (messageGenerator == null)
+            {
+                throw new ArgumentNullException(nameof(messageGenerator));
+            }
+
+            if (logger.IsEnabled(LogLevel.Debug))
+            {
+                logger.Log(LogLevel.Debug, eventId, messageGenerator(), exception, _messageFormatter);
+            }
+        }
+
+        /// <summary>
+        /// Formats and writes a debug log message.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


pakrym:
Using lambdas for this would almost certainly lead to capturing local variables and allocating a  closure per call to `Log*` even if log level is not enabled, making what seemed like optimization in fact waste more resources.

drungrin:
The lambda is an option. The evaluation of the expression is made immediately instead of passing away to the logging system, and when the funcition return the eventual capture is gone.
 
If the logging level isn’t enabled, the function just returns. The eventual capture exists, but for some ticks.

And also the devloper should know the impact of a capture and how to minimize it. The ide warns about those.

Think of logging why a purchase was denied. Increasing the logging level to trace, for example, should do additional queries to the database, what is way expensive than the capture.

Those aren’t intended to be used to log simple messages. Is the developer that makes the decision if the capture worth the benefit of those new functions.

pakrym:
People are more often then not not aware of closures and their cost and even short lived objects are causing performance penalty when there's a lot of them. 

Also these extension methods are easy to implement on top of existing APIs if someone want's to avoid `IsEnabled` check and understands the cost.

drungrin:
I don't believe this is a show-stopper. If the concern that people doesn't have the skills or aren't aware of the drawbacks was really a issue, LINQ and others Technologies that uses closures wouldn't exists. I just don't agree with those arguments. 

I was using exactly as extension method, but thought of being an improvement to the logging api, since the other extension methods are also easy to implemente on top of existing APIs if someone wants to avoid passing `LogLevel.Debug` on every function.

95% of my logging is made with the regular methods. Some can benefit of having additional expensive information.

benaadams:
What benefit to these provide over [`LoggerMessage.Define`](https://github.com/aspnet/Logging/blob/7f4bcd95ba1d0c982ad42cc78426a23fd5065f39/src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs#L80-L232)?

```csharp
static readonly Action<ILogger, string, string, Exception> _locatedDevelopmentCertificate =
    LoggerMessage.Define<string, string>(
        LogLevel.Debug, 
        new EventId(0, nameof(LocatedDevelopmentCertificate)), 
        "Using development certificate: {certificateSubjectName} (Thumbprint: {certificateThumbprint})");

static void LocatedDevelopmentCertificate(this ILogger logger, X509Certificate2 certificate)
 => _locatedDevelopmentCertificate(logger, certificate.Subject, certificate.Thumbprint, null);
```

drungrin:
@benaadams I believe LoggerMessage.Define cann't do this (assuming that X509Certificate2 .Verify() is a very expensive operation):

```
logger.LogDebug(new EventId(0, nameof(LocatedDevelopmentCertificate)), () =>
            {
                return certificate.Verify() 
                ? $"Using development certificate: {certificate.Subject} (Thumbprint: {certificate.Thumbprint}, Certificate Verified)" 
                : $"Using development certificate: {certificate.Subject} (Thumbprint: {certificate.Thumbprint}, Certificate not verified)";
            });
```
If LogLevel isn't debug, Verify isn't called.

davidfowl:
@drungrin in that case, then use an if statement. It's cheaper than allocating a closure.

davidfowl:
We're not going to add these extension methods.

drungrin:
@davidfowl thanks for the feedback.

PR #733, opened 2017-11-23
Fix config loading to be case insensitive

AndreasBieber:
Hi,

this PR files an issue with loading filter configuration. The configuration section `LogLevel` and the category name as well were read case-sensitive while the Microsoft.Extensions.Configuration is case-insensitive, which leads to an unexpected behavior.

see:
https://github.com/aspnet/Logging/blob/71e619a5599af7bdf903061ac011ed2c6f62a744/src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs#L33

https://github.com/aspnet/Logging/blob/71e619a5599af7bdf903061ac011ed2c6f62a744/src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs#L58

Cheers,
 Andy

### DIFF
--- a/src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs
+++ b/src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Extensions.Logging
 {
     internal class LoggerFilterConfigureOptions : IConfigureOptions<LoggerFilterOptions>
     {
+        private const string LOGLEVEL_KEY = "LogLevel";
         private readonly IConfiguration _configuration;
 
         public LoggerFilterConfigureOptions(IConfiguration configuration)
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs
```diff
@@ -27,17 +27,18 @@ private void LoadDefaultConfigValues(LoggerFilterOptions options)
             {
                 return;
             }
-
+            
+            const string logLevelKey = "LogLevel";
```
pakrym: Class level const

src/Microsoft.Extensions.Logging.Configuration/LoggerFilterConfigureOptions.cs
```diff
@@ -55,7 +56,7 @@ private void LoadRules(LoggerFilterOptions options, IConfigurationSection config
                 if (TryGetSwitch(section.Value, out var level))
                 {
                     var category = section.Key;
-                    if (category == "Default")
+                    if (category.Equals("Default", StringComparison.OrdinalIgnoreCase))
```
pakrym: Class level const

test/Microsoft.Extensions.Logging.Test/LoggerFilterTest.cs
```diff
@@ -398,6 +400,42 @@ public void AddFilterWithProviderNameAndCategoryFilterFuncFiltersCorrectly()
...
+        {
+            var serviceProvider = new ServiceCollection()
+                .AddLogging(
+                    builder => builder.AddConfiguration(new ConfigurationBuilder()
+                                                            .AddInMemoryCollection(new[] {new KeyValuePair<string, string>("logLevel:Default", "Error")})
+                                                            .Build())
+                )
+                .BuildServiceProvider();
+            var filterOptions = new LoggerFilterOptions();
+            var options = serviceProvider.GetRequiredService<IConfigureOptions<LoggerFilterOptions>>();
```
pakrym: Just resolve `IOptions<LoggerFilterOptions>` here and below.

pakrym:
We prefer CamelCase for `consts`:
https://github.com/aspnet/Logging/blob/cc350d7ef616ef292c1b4ae7130b8c2b45fc1164/src/Microsoft.Extensions.Logging/ProviderAliasUtilities.cs#L11

Otherwise LGTM

pakrym (approved):

pakrym:
@glennc are we fine to take this change in 2.1?

davidfowl:
@glennc ping

pakrym:
Thank you.

PR #739, opened 2017-12-05
Don't use Task.Factory.StartNew, it's dangerous

bojanrajkovic:
Using `Task.Factory.StartNew`, without specifying a task scheduler or
denying child attach or any of a half dozen non-obvious flags/parameters
can cause undesired behavior from the task. This hidden behavior causes
any uses of the console logger in Xamarin Workbooks to hijack the main thread,
preventing further execution, regardless of the workbook platform (.NET Core,
iOS, Desktop .NET all exhibit this issue at slightly different points).

See this article for some of the dangers of `Task.Factory.StartNew`,
particularly around not using `DenyChildAttach`, and why `LongRunning`
is unnecessary: https://blog.stephencleary.com/2013/08/startnew-is-dangerous.html.
This PR replaces `Task.Factory.StartNew` with `Task.Run`, which is semantically more
correct, functionally equivalent, and more importantly, does not cause any main
thread hijacking.

Would be nice if this could make it into 2.1.0. :)

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
+++ b/src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Threading.Tasks;
+using System.Threading;
 
 namespace Microsoft.Extensions.Logging.Console.Internal
 {
@@ -12,16 +12,19 @@ namespace Microsoft.Extensions.Logging.Console.Internal
         private const int _maxQueuedMessages = 1024;
 
         private readonly BlockingCollection<LogMessageEntry> _messageQueue = new BlockingCollection<LogMessageEntry>(_maxQueuedMessages);
-        private readonly Task _outputTask;
+        private readonly Thread _outputThread;
 
         public IConsole Console;
 
         public ConsoleLoggerProcessor()
         {
             // Start Console message queue processor
-            _outputTask = Task.Factory.StartNew(
-                ProcessLogQueue,
-                TaskCreationOptions.LongRunning);
+            _outputThread = new Thread(ProcessLogQueue)
+            {
+                IsBackground = true,
+                Name = "Console logger queue processing thread"
+            };
+            _outputThread.Start();
         }
 
         public virtual void EnqueueMessage(LogMessageEntry message)
@@ -54,9 +57,20 @@ namespace Microsoft.Extensions.Logging.Console.Internal
 
         private void ProcessLogQueue()
         {
-            foreach (var message in _messageQueue.GetConsumingEnumerable())
+            try
             {
-                WriteMessage(message);
+                foreach (var message in _messageQueue.GetConsumingEnumerable())
+                {
+                    WriteMessage(message);
+                }
+            }
+            catch
+            {
+                try
+                {
+                    _messageQueue.CompleteAdding();
+                }
+                catch { }
             }
         }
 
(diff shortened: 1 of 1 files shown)


bojanrajkovic:
Build failure seems unrelated--it failed when trying to report test results to AppVeyor, seemingly due to an issue in AppVeyor itself.

pakrym:
I'm wondering if a thread is better then starting a task that is blocked most of the time.

bojanrajkovic:
Might be, though I wouldn't be surprised if the thread pool backing this didn't adjust, as Stephen's blog post points out.

benaadams:
Using `Task.Run` will cause it to permanently eat a threadpool thread.

Would be better just to specify the scheduler? (as there are no child tasks created)

```csharp
Task.Factory.StartNew(
    ProcessLogQueue,		
    CancellationToken.None, 
    TaskCreationOptions.LongRunning, 
    TaskScheduler.Default);
```

bojanrajkovic:
I considered that approach as well, but I felt the functional result was roughly equivalent and the Task.Run variant is cleaner&mdash;my understanding is that `TaskCreationOptions.LongRunning` creating a new thread is merely an implementation detail and is not _guaranteed_ to create a new thread in perpetuity.

 If we really want to avoid possibly consuming a threadpool thread, the code should probably create a thread explicitly.

benaadams:
Really shouldn't consume a threadpool thread forever; which is what changing it to Task.Run will do (for this method)

bojanrajkovic:
In that case, I'd prefer to go with @pakrym's suggestion&mdash;it's at least explicit about the desired behavior.

benaadams:
Set `IsBackground = true` :)

pakrym:
And a thread name

bojanrajkovic:
Thanks for the suggestions/discussions y'all, updated the PR. Shame the CLA bot is broken!

pakrym:
@bojanrajkovic I think this is a good change but I'll have to wait until bot comes alive again.

BrennanConroy:
@Eilon Could you take a look?

BrennanConroy:
@bojanrajkovic Did you actually sign the license? Even if you're an employee you're still required to sign it.

bojanrajkovic:
Pretty sure I have in the past--I'm almost certain the check came up green originally, though it's a bit hard to spot now, as I squashed in changes.

bojanrajkovic:
OK, turns out I thought I'd signed it, but hadn't. All set now.

pakrym:
Add `try/catch` suppressing all exception in `ProcessLogQueue` please. Throwing from thread entry point would crash the process.

bojanrajkovic:
Updated!

src/Microsoft.Extensions.Logging.Console/Internal/ConsoleLoggerProcessor.cs
```diff
@@ -54,10 +57,14 @@ internal virtual void WriteMessage(LogMessageEntry message)
...
-            foreach (var message in _messageQueue.GetConsumingEnumerable())
+            try
             {
-                WriteMessage(message);
+                foreach (var message in _messageQueue.GetConsumingEnumerable())
+                {
+                    WriteMessage(message);
+                }
             }
+            catch { }
```
davidfowl: If this ever happens it'll be a pain to figure out what went wrong with your logging.

davidfowl: Actually on second thought, I think we need to close the _messageQueue here as well. Otherwise it can just back up until it reaches the max and then the application will just block trying to log.

pakrym: Just don't forget to wrap that code into try/catch too.

bojanrajkovic:
Updated to close the queue in the catch block&mdash;I thought about adding a CWL or something similar to make it clear what happened, but wasn't sure on how useful/necessary it would be.

abock:
I would write the exception via `Debug.WriteLine` so any trace listeners will see it. I would avoid writing anything to stdout/err explicitly.

davidfowl:
Thanks @bojanrajkovic

PR #743, opened 2017-12-07
[Logging.Testing] Drop test logging level down to Trace

anurse:
Also, the tests were leaving directories and files around in Temp. That ended up requiring some changes to how objects are disposed because they were keeping the test files locked.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -25,13 +25,15 @@ namespace Microsoft.Extensions.Logging.Testing
         private readonly ILogger _globalLogger;
         private readonly string _baseDirectory;
         private readonly string _assemblyName;
+        private readonly IServiceProvider _serviceProvider;
 
-        private AssemblyTestLog(ILoggerFactory globalLoggerFactory, ILogger globalLogger, string baseDirectory, string assemblyName)
+        private AssemblyTestLog(ILoggerFactory globalLoggerFactory, ILogger globalLogger, string baseDirectory, string assemblyName, IServiceProvider serviceProvider)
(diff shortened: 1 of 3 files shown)


anurse:
Changing the approach very slightly to use `minLogLevel` parameter instead of changing the default. But there is still a bug in over-locking the files here. Stand-by for an update

anurse:
Note this also removed the Console logger from all the logging contexts. It was just too noisy and unnecessary when you set the environment variable (which we do on the CI)

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -51,60 +58,85 @@ public IDisposable StartTestLog(ITestOutputHelper output, string className, out
...
-
-        public static AssemblyTestLog Create(string assemblyName, string baseDirectory)
-        {
             var serviceCollection = new ServiceCollection();
+            serviceCollection.AddLogging(builder =>
+            {
+                builder.SetMinimumLevel(minLogLevel);
 
-            // Let the global logger log to the console, it's just "Starting X..." "Finished X..."
-            serviceCollection.AddLogging(builder => builder.AddConsole());
```
halter73: What do we need to do to enable xunit's long running test reporting now that this is being removed?

anurse: Globally: https://github.com/aspnet/Coherence-Signed/issues/573

Locally: Just ensure the build drops an xunit.runner.json in the output directory with the [necessary config](https://xunit.github.io/docs/configuring-with-json.html). See https://github.com/aspnet/SignalR/blob/ce39b47e7de0f3baf219732801481497d2f1591e/test/xunit.runner.json for more detail

PR #761, opened 2018-02-08
Alter the dependency of Console to Abstractions rather than Microsoft.Extensions.Logging

LordMike:
Currently, `Microsoft.Extensions.Logging.Console` depends on `Microsoft.Extensions.Logging` - but this is unecessary, as only interfaces from `Microsoft.Extensions.Logging.Abstractions` are actually used. This will reduce the referenced assemblies by approximately 2 MB (`System.Linq.Expressions` alone is 1.4 MB alone, and isn't needed if you don't need `Microsoft.Extensions.Logging`)

I've tried altering the other sink implementations, but they depend on stuff from Logging directly, such as `ILoggingBuilder`.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/Microsoft.Extensions.Logging.Console.csproj
+++ b/src/Microsoft.Extensions.Logging.Console/Microsoft.Extensions.Logging.Console.csproj
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


davidfowl:
Can't change dependencies like this, it's a breaking change.

LordMike:
Well shoot. I'd push more for this if I could change all the sinks to only depend on Abstractions, because then it would go from "consistent" to "consistent" - having only Console logging depend on Abstractions and the rest not, is inconsistent.

PR #766, opened 2018-02-27
Moce exteions methods of LoggerFactory to the class itself

siyamandayubi:
The main point is, we can not use these useful extension methods, because they can not be mocked for automated tests. By moving them to the class itself, they become mockable


Eilon:
Hi, this is a breaking change, so unfortunately we cannot accept it.

Eilon:
For your own unit testing, consider using a wrapper that can be mocked, and at runtime use an implementation that calls the existing methods.

siyamandayubi:
Thanks for the quick feedback. I will go with using a wrapper, but generally speaking, should we use extension methods in such a way to override existing methods of an interface?

Eilon:
@siyamandayubi the extension methods don't override anything; they are just convenience methods that make the logging system easier to use.

Issue #751, opened 2018-01-17
Windows EventLog package for .NET Core


henkmollema:
The Windows compatibility pack provides a preview [Windows EventLog](https://www.nuget.org/packages/System.Diagnostics.EventLog/) package for .NET Core/.NET Standard 2.0. Are there any plans to integrate this package in the existing EventLog implementation? I successfully did so locally and it compiles cleanly. 

However, the compatibility package is still in preview. Are there any timeframes for this?

henkmollema:
@Eilon do you think a .NET Core compatible package could ship in 2.1? We're hosting a NuGet package of our own no to support this. I could create a PR to accommodate this if you like.

DamianEdwards:
We should look at this and consider making the defaults log fatal errors there.

davidfowl:
@DamianEdwards if you're saying we should log all existing fatal errors here that's fine but going through and adding more fatal logs would be a new feature at this point. Somebody would need to go through and identify what types of errors we want to be diagnosable via the event log (AFAIK we don't have that many fatal log messages).

DamianEdwards:
I meant existing, and maybe just a new one at some very root level. Or we convince dotnet to log app crashes to Event Log and avoid us having to do anything 😀

henkmollema:
@davidfowl @DamianEdwards are there any timelines regarding a final `System.Diagnostics.EventLog` package? Will it ship in time for ASP.NET Core 2.1?

davidfowl:
@terrajobst whats the timeline for the windows compatibility pack?

terrajobst:
It will ship RTM with .NET Core 2.1. I just checked and it looks like we didn't publish an update although the preview for .NET Core 2.1 just went out. Checking with my devs; an update should come any day now.

muratg:
Let's use this bug to track bringing .NETStandard support to the EventLog implementation. We'll triage based on @terrajobst's answer.

@DamianEdwards Could you file another bug with with logging fatal errors?

terrajobst:
Update. We couldn't push our package because someone is squatting on one of our package IDs. However, the [event log support just shipped](https://www.nuget.org/packages/System.Diagnostics.EventLog). So give it a spin and see whether it works!

DamianEdwards:
Issue logged on dotnet: https://github.com/dotnet/cli/issues/8719

henkmollema:
Thanks guys!

PR #770, opened 2018-03-02
Add EventLog support for netstandard20

pakrym:
Cross compile library and tests.

Fixes https://github.com/aspnet/Logging/issues/751


pakrym:
Yep, fixing

Issue #763, opened 2018-02-13
Unnecessary code execution in LogValuesFormatter


mikevanoo:
### Title
Unnecessary code execution in LogValuesFormatter. The ctor contains the following code:

https://github.com/aspnet/Logging/blob/5a906a4e8492f709d25a2f7cc737da2915771d44/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs#L38-L53

However, as `formatDelimiterIndex` is only ever used inside the `else` block, it should be moved there, thus removing some unnecessary calls to `FindIndexOfAny`.

### Functional impact
No functional impact, just more optimal code execution.

davidfowl:
Sounds like it would make a great pull request.

mikevanoo-fsg:
Fair point, I'll see what I can do....

d-saravanan:
I have submitted a [pull-request](https://github.com/aspnet/Logging/pull/771). Please review the same and let me know the review comments.

PR #771, opened 2018-03-03
Constructor logic is refactored for the delayed execution of the placeholder refactoring algorithm

d-saravanan:
**Summary**
This pull-request has the logic that will lazy initialize the `_format`  field so that when the `Format`  method is invoked with the arguments, the string is processed and then used for transformation. 
Hence, there will be faster object construction and similar to that of the existing implementation..

Kindly review this and share your review comments. 

Addresses #763

### DIFF
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Extensions.Logging.Internal
             char openBrace = '{', closingBrace = '}';
 
             //{} => Nothing to process, min length should be greater than 2. Ex: "{0}"
-            if ((scanIndex == endIndex) || format.Length < 3) 
+            if (format.Length < 3) 
             {
                 _format = format;
                 return format;
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Extensions.Logging.Internal
         private static readonly char[] FormatDelimiters = { ',', ':' };
         private string _format;
         private List<string> _valueNames;
+        private List<string> _emptyValueNames = new List<string>(0);
 
         public LogValuesFormatter(string format)
         {
@@ -31,14 +32,14 @@ namespace Microsoft.Extensions.Logging.Internal
         {
             get
             {
-                if (string.IsNullOrEmpty(_format)) FormatInput(OriginalFormat); //construct the valuenames from the input
-                return _valueNames ?? new List<string>(0);
+                if (_format == null) FormatInput(OriginalFormat); //construct the valuenames from the input
+                return _valueNames ?? _emptyValueNames;
             }
         }
 
         private string FormatInput(string format)
         {
-            if (!string.IsNullOrEmpty(_format))
+            if (_format != null)
             {
                 return _format;
             }
@@ -168,6 +169,7 @@ namespace Microsoft.Extensions.Logging.Internal
 
         public KeyValuePair<string, object> GetValue(object[] values, int index)
         {
+            if (_format == null) FormatInput(OriginalFormat);
             if (index < 0 || index > _valueNames?.Count)
             {
                 throw new IndexOutOfRangeException(nameof(index));
@@ -185,6 +187,7 @@ namespace Microsoft.Extensions.Logging.Internal
         {
             var valueArray = new KeyValuePair<string, object>[values.Length + 1];
 
+            if (_format == null) FormatInput(OriginalFormat);
             if (_valueNames != null)
             {
                 for (var index = 0; index != _valueNames.Count; ++index)
--- a/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
+++ b/src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
@@ -17,9 +17,9 @@ namespace Microsoft.Extensions.Logging.Internal
     {
         private const string NullValue = "(null)";
         private static readonly object[] EmptyArray = new object[0];
-        private static readonly char[] FormatDelimiters = {',', ':'};
+        private static readonly char[] FormatDelimiters = { ',', ':' };
         private string _format;
-        private List<string> _valueNames = new List<string>();
+        private List<string> _valueNames;
 
         public LogValuesFormatter(string format)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,36 @@ public class LogValuesFormatter
...
         private static readonly object[] EmptyArray = new object[0];
         private static readonly char[] FormatDelimiters = {',', ':'};
-        private readonly string _format;
+        private readonly Lazy<string> _format;
         private readonly List<string> _valueNames = new List<string>();
 
         public LogValuesFormatter(string format)
         {
             OriginalFormat = format;
+            _format = new Lazy<string>(() => FormatInput(format), true);
```
pakrym: 2 additional allocations per format string.

Just make FormatInput check the _format field and calculate when it's null.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,36 @@ public class LogValuesFormatter
...
+        }
+
+        public string OriginalFormat { get; private set; }
```
pakrym: Why?

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,36 @@ public class LogValuesFormatter
...
+        public List<string> ValueNames => _valueNames;
```
pakrym: Why?

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: Used this lazy loading logic as the variable was readonly. Refactored this code as per the comments and checked-in now.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: Since _format was readonly, used this to lazily load the value

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: refactored and checked-in. Please review the latest check-in and share review comments.

pakrym:
@d-saravanan tests fail.

d-saravanan:
@pakrym I have updated the code and fixed the test cases. The Travis-CI build has passed in Linux  and have failed in Mac version due to dotnet build. However, in my repository, the Travis-CI builds are passing. Can you please trigger a build and let me know review comments if any.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,46 @@ public class LogValuesFormatter
...
+                return _format;
+            }
             var sb = new StringBuilder();
             var scanIndex = 0;
             var endIndex = format.Length;
+            char openBrace = '{', closingBrace = '}';
 
-            while (scanIndex < endIndex)
+            //There should be an open and a closed brace, absence of which, results in unwanted processing.
+            if((!format.Contains(openBrace) && !format.Contains(closingBrace))
```
pakrym: Previous code was optimized to traverse string only ones in search of `{` or `}`, new code would do it multiple times.

In the case where the string doesn't contain braces `_format` would never be set and string would get re-scanned every time `FormatInput` is called.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,46 @@ public class LogValuesFormatter
...
+            get
+            {
+                if (string.IsNullOrEmpty(_format)) FormatInput(OriginalFormat); //construct the valuenames from the input
+                return _valueNames;
+            }
+        }
+
+        private string FormatInput(string format)
+        {
+            if(!string.IsNullOrEmpty(_format))
```
pakrym: Space after if

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -18,24 +18,46 @@ public class LogValuesFormatter
         private const string NullValue = "(null)";
         private static readonly object[] EmptyArray = new object[0];
         private static readonly char[] FormatDelimiters = {',', ':'};
-        private readonly string _format;
-        private readonly List<string> _valueNames = new List<string>();
+        private string _format;
+        private List<string> _valueNames = new List<string>();
```
pakrym: Do we need to eagerly allocate _valueNames?

d-saravanan: The allocation is modified to be initialized when there are braces in the input format and suitable null checks are added.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: code formatting done for the full file.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: Sorry, the check made only considering the negative case. Since the contains will cause additional search operation, the conditions are removed. 

Also, the _format is set even if the input string does not contain any braces.

Please review the check-in and share your comments.

d-saravanan:
@pakrym I have checked-in the code addressing your review comments. Also, I wrongly started a review thinking that it will raise to your name. Please review the changes and let me know your review comments.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -17,25 +17,47 @@ public class LogValuesFormatter
...
+            if (!string.IsNullOrEmpty(_format))
+            {
+                return _format;
+            }
             var scanIndex = 0;
             var endIndex = format.Length;
+            char openBrace = '{', closingBrace = '}';
 
+            //{} => Nothing to process, min length should be greater than 2. Ex: "{0}"
+            if ((scanIndex == endIndex) || format.Length < 3)
```
pakrym: `scanIndex == endIndex` Length==0? It's covered by second condition already.

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
```diff
@@ -17,25 +17,47 @@ public class LogValuesFormatter
...
+            get
+            {
+                if (string.IsNullOrEmpty(_format)) FormatInput(OriginalFormat); //construct the valuenames from the input
+                return _valueNames ?? new List<string>(0);
+            }
+        }
+
+        private string FormatInput(string format)
+        {
+            if (!string.IsNullOrEmpty(_format))
```
pakrym: Why isn't null check enough?

d-saravanan: This is initialized with an empty list because we have directly used the variable without null checks in the LoggerMessage.cs, FormattedLogValues.cs & LogFormatAnalyzer.cs. If I set null value here, those files will have to check for null and then add to the KeyValuePair. The references are listed below for your reference. Kindly confirm

image

src/Microsoft.Extensions.Logging.Abstractions/Internal/LogValuesFormatter.cs
d-saravanan: removed the condition and checked-in.

d-saravanan:
@pakrym can you please let me know if the changes are fine.

d-saravanan:
@pakrym I have made the changes and checked-in. Please review and let me know if it is fine.

pakrym:
Closing this PR. It hasn't been updated in a long time and tests are failing.

Issue #764, opened 2018-02-19
Add a monochrome output option to ConsoleLogger


joergjo:
Microsoft.Extensions.Logging.Console.ConsoleLogger should allow to suppress colored output.

Rationale: The colored category output from ConsoleLogger breaks several log viewing features in container orchestration platforms like Service Fabric (Service Fabric Explorer trips over the ANSI escape sequences) or Kubernetes (`kubectl logs` dumps garbage characters on a Windows console). Therefore it would be useful to turn off colored console output to play along nicely with these platforms.

```
[40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Mvc.RedirectToActionResult[1]
      Executing RedirectResult, redirecting to /Cart/Index?returnUrl=%2FChess%2FPage1.
[40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Mvc.Internal.ControllerActionInvoker[2]
      Executed action SportsStore.Controllers.CartController.AddToCart (SportsStore) in 28.8502ms
[40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
      Request finished in 40.7131ms 302
```

(Applies to Microsoft.AspNetCore.All 2.0.5 and previous versions.)

davidfowl:
> (Applies to Microsoft.AspNetCore.All 2.0.5 and previous versions.)

We won't backport changes like this to all prior versions of ASP.NET Core but we can add a new option to disable this for 2.1.

/cc @muratg @Eilon @glennc

joergjo:
That's OK for me. Just wanted to adhere to the issue contribution guidelines :-)

muratg:
We would accept a PR for this.  If anyone wants to contribute, it may be useful to talk about the approach here first.

bruce-lindsay:
@muratg I would like to fix this. What are you looking for in terms of approach?

davidfowl:
We should expose an option on ConsoleLoggerOptions  to disable the colors.

/cc @pakrym @glennc

pakrym:
https://github.com/aspnet/Logging/pull/772

PR #772, opened 2018-03-03
Add DisableColors option (#764)

bruce-lindsay:


davidfowl:
@bruce-lindsay the code doesn't seem to compile...

bruce-lindsay:
@davidfowl resolved. Sorry I made an error

bruce-lindsay:
I have done everything from docker image microsoft/dotnet:2.1-sdk-stretch because the dotnet 2.1.4 doesn't tolerate the build target. as a result some blocks of code were windows only and could not be built or tested. I'm hoping I can lead on the review + CI to catch anything

```The current .NET SDK does not support targeting .NET Core 2.1.  Either target .NET Core 2.0 or lower, or use a version of the .NET SDK that supports .NET Core 2.1.```

File: src/Microsoft.Extensions.Logging.Console/ConsoleLoggerOptions.cs [hunk 1]
```diff
@@ -8,5 +8,6 @@ namespace Microsoft.Extensions.Logging.Console
     public class ConsoleLoggerOptions
     {
         public bool IncludeScopes { get; set; } = false;
+        public bool DisableColors { get; set; } = false;
```
Eilon: No need to set default "default values". I'd recommend removing the ` = false;` from this line *and* the one above it.

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -22,9 +22,10 @@ public class ConsoleLoggerProvider : ILoggerProvider, ISupportExternalScope
         private static readonly Func<string, LogLevel, bool> falseFilter = (cat, level) => false;
         private IDisposable _optionsReloadToken;
         private bool _includeScopes;
+        private bool _disableColors;
         private IExternalScopeProvider _scopeProvider;
 
-        public ConsoleLoggerProvider(Func<string, LogLevel, bool> filter, bool includeScopes)
+        public ConsoleLoggerProvider(Func<string, LogLevel, bool> filter, bool includeScopes, bool disableColors = false)
```
Eilon: This looks like a breaking change. I recommend adding a 2nd constructor that has the extra parameter, and then chain the two constructors.

bruce-lindsay: Addressed in 2bfbdfb2d5356bc7145a47dff05f953b99b76279

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerOptions.cs
bruce-lindsay: Addressed in 2bfbdfb2d5356bc7145a47dff05f953b99b76279

src/Microsoft.Extensions.Logging.Console/ConsoleLoggerProvider.cs
```diff
@@ -77,11 +86,14 @@ private void OnConfigurationReload(object state)
                 _settings = _settings.Reload();
 
                 _includeScopes = _settings?.IncludeScopes ?? false;
+                bool disableColors = _settings?.DisableColors ?? false;
```
Eilon: Can you use `var` here?

bruce-lindsay: addressed in 67c4ee557b7597425a7f722551cf4e5514a81df6

Eilon:
Overall looks good to me, but I'm not a Logging expert, so I'll let someone from @muratg 's team review.

pakrym:
Can you please add ConsoleLoggerOptions_IncludeScopes_IsReadFromLoggingConfiguration and ConsoleLoggerOptions_IncludeScopes_IsAppliedToLoggers test variations for console collors.

bruce-lindsay:
@pakrym added requested test variations

pakrym:
@muratg, this change involves a breaking change of rarely used `IConsoleLoggerSettings` interface. Are we fine with it?

Eilon:
I would add a 2nd interface that adds the new API. The bar for breaking changes is very high.

pakrym:
@bruce-lindsay I had to do couple changes to avoid breaking changes in legacy `IConsoleLoggerSettings` code.

PR #774, opened 2018-03-06
Update for when log message is an empty string

danielabbatt:
If logging an empty string, e.g. "", this fails to get lines.First().
This fix worked for my copy of this code.


pakrym:
Thank you.

PR #777, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


src/Microsoft.Extensions.Logging.Configuration/baseline.netcore.json
```diff
@@ -31,111 +31,6 @@
         }
       ],
       "GenericParameters": []
-    },
-    {
-      "Name": "Microsoft.Extensions.Logging.Configuration.ILoggerProviderConfiguration<T0>",
```
pranavkm: Isn't this a breaking change? Maybe @Eilon has more context?

src/Microsoft.Extensions.Logging.Configuration/baseline.netcore.json
```diff
@@ -1,5 +1,5 @@
 {
-  "AssemblyIdentity": "Microsoft.Extensions.Logging.Configuration, Version=2.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
ryanbrandenburg: The fact that this is listed as having been generated off of 2.1.0 seems concerning.

src/Microsoft.Extensions.Logging.Configuration/baseline.netcore.json
Eilon: If the baseline appears to have a breaking change, does it even matter anymore? That would imply a break from 1.x to 2.0, so it's too late?

ryanbrandenburg: SO, `ILoggerProviderConfiguration` exists in the release/2.1 branch but does not exist in either the release/2.0 or release/1.1 branch. I'm pretty sure that as mentioned above, this baselines was accidentally generated against 2.1.0, and therefore most changes to it listed here don't mean much.

Eilon: Ah yes that presumably explains it.

src/Microsoft.Extensions.Logging.Testing/baseline.netcore.json
```diff
@@ -1,5 +1,9 @@
 {
+<<<<<<< HEAD
```
Eilon: Bad merge?

ryanbrandenburg:
:up::date:

src/Microsoft.Extensions.Logging.Abstractions/breakingchanges.netcore.json
```diff
@@ -0,0 +1,14 @@
+[
+    {
+        "TypeId": "public class Microsoft.Extensions.Logging.Abstractions.NullLogger<T0> : Microsoft.Extensions.Logging.ILogger<T0>",
```
pranavkm: This type exists - https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/NullLoggerOfT.cs. Why is it marked as removal?

pranavkm: Is this because the baseline was incorrect to start with?

Eilon: We think it's a bug in the tool, which is logged, and we have another issue to revert these "fake" breaking changes.

Issue #500, opened 2016-10-14
Microsoft.Extensions.Logging.Abstractions not marked CLSCompliant


PavelSafronov:
The Microsoft.Extensions.Logging.Abstractions assembly is missing the CLSCompliant(true) attribute. Is there a reason for this, or was this merely an accidental omission?

muratg:
@PavelSafronov do you have a use case to have this attribute in the Logging repo?

PavelSafronov:
The use case is writing a CLS-compliant library which uses or extends Microsoft.Extensions.Logging.Abstractions.

Is there a reason _not to_ have this attribute?

gulbanana:
ran into this today. we may drop clscompliant from our own library but it would be nice to not be forced to make the choice

MichaCo:
@muratg I think it is about consistency most of all. I got asked to add that back to my libraries although it has very little value I think?! But some Microsoft.* or aspnet core packages actually do have that set.

https://github.com/search?p=7&q=org%3Aaspnet+CLSCompliant&type=Code
https://github.com/search?l=C%23&q=org%3Adotnet+CLSCompliant&type=Code

Personal opinion: Wouldn't be too bad if everyone would just drop it, kinda annoying to care about it (in static code analysis) ;)

muratg:
We are closing this issue because no further action is planned for this issue. If you still have any issues or questions, please log a new issue with any additional details that you have.

PR #778, opened 2018-03-15
Mark Logging.Abstractions as CLSCompliant

tibel:
Only change is to mark Logging.Abstractions assembly as CLSCompliant.

Background:
We have some .NET Framework projects that we want to port to .NET Standard. All these projects are marked as CLSCompliant. So this change would help a lot to make a step-by-step migration much easier.

This resolve issue #500


Eilon:
For the reasons mentioned in the issue, plus https://github.com/aspnet/Home/issues/2689#issuecomment-354693946, we are not planning to mark these types as CLSCompliant.

PR #785, opened 2018-03-24
add timestamp to test logs

anurse:
Debugging some tests and timestamps would be very useful.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/XunitLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/XunitLoggerProvider.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Extensions.Logging.Testing/XunitLoggerProvider.cs
```diff
@@ -59,7 +59,8 @@ public XunitLogger(ITestOutputHelper output, string category, LogLevel minLogLev
             // Buffer the message into a single string in order to avoid shearing the message when running across multiple threads.
             var messageBuilder = new StringBuilder();
 
-            var firstLinePrefix = $"| {_category} {logLevel}: ";
+            var timestamp = DateTime.Now.ToString("s");
```
benaadams: Fine choice 😄

src/Microsoft.Extensions.Logging.Testing/XunitLoggerProvider.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Copyright (c) .NET Foundation. All rights reserved.
```
pakrym: Are you adding or removing BOMs?

anurse: Not intentionally?

PR #788, opened 2018-03-24
MOAR analyzers

anurse:
I got tired of working on stuff that was [being stupid](https://twitter.com/anurse/status/977546425037815809) so I wrote these two new diagnostics, which are OFF by default (and will also help tidy up some of the chaotic logging I added while debugging the aforementioned stupidity, so it wasn't all a waste ;))

* `MEL0004: For improved performance, use pre-compiled log messages instead of calling '[MethodName]' with a string message.`
* `MEL0005: For consistency with logs emitted from other components, use 'PascalCase' for log message tokens.`

I also re-numbered them from `MEL1`-`MEL3` to `MEL0001`-`MEL0003` for consistency (most diagnostic IDs are at least 4 digits long)

I promoted them to Error in the below test, they work pretty well:

image

cc @davidfowl 

As a reminder, this package does not ship today, it's purely an "alpha"


src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
```diff
@@ -113,7 +120,10 @@ private void AnalyzeFormatArgument(SyntaxNodeAnalysisContext syntaxContext, Expr
                 if (int.TryParse(valueName, out _))
                 {
                     syntaxContext.ReportDiagnostic(Diagnostic.Create(Descriptors.MEL1NumericsInFormatString, formatExpression.GetLocation()));
-                    break;
```
anurse: I couldn't see a reason for this `break`, it would mean you only get one diagnostic per format string, but I'd rather emit all of the diagnostics to avoid new errors appearing when you fix one.

anurse:
I spent like, 10 minutes looking in to writing a code fix for the "use compiled messages" one, and noped out of there ;).

src/Microsoft.Extensions.Logging.Analyzers/Descriptors.cs
```diff
@@ -8,15 +8,23 @@ namespace Microsoft.Extensions.Logging.Analyzers
...
-            "MEL2", "Logging format string should not be dynamically generated",
+            "MEL0002", "Logging format string should not be dynamically generated",
             "Logging format string should not be dynamically generated", "Usage", DiagnosticSeverity.Info, true);
 
         public static DiagnosticDescriptor MEL3FormatParameterCountMismatch = new DiagnosticDescriptor(
-            "MEL3", "Logging format string parameter count mismatch",
+            "MEL0003", "Logging format string parameter count mismatch",
             "Logging format string parameter count mismatch", "Usage", DiagnosticSeverity.Warning, true);
+
+        public static DiagnosticDescriptor MEL4UseCompiledLogMessages = new DiagnosticDescriptor(
```
pakrym: Should we rename properties to be 0003 too?

src/Microsoft.Extensions.Logging.Analyzers/UseCompiledLogMessagesAnalyzer.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.Extensions.Logging.Internal;
+
+namespace Microsoft.Extensions.Logging.Analyzers
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class UseCompiledLogMessagesAnalyzer : DiagnosticAnalyzer
```
pakrym: Can we stick this logic into existing analyzer?

anurse: I guess we could

src/Microsoft.Extensions.Logging.Analyzers/Descriptors.cs
anurse: I'll do that. It's most important for the ID since that's a string and is sorted as a string.

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
```diff
@@ -113,7 +120,10 @@ private void AnalyzeFormatArgument(SyntaxNodeAnalysisContext syntaxContext, Expr
...
+                }
+                else if (char.IsLower(valueName[0]))
```
pakrym: Would it fire on numbers? It's a little strange to show MEL1 and MEL5 at the same place.

anurse: I'll add a test. I don't believe so though.

anurse:
I'm adding support for `LoggerMessage.Define` and `LoggerMessage.DefineScope` as well.

src/Microsoft.Extensions.Logging.Analyzers/LogFormatAnalyzer.cs
anurse: Actually, the existing tests verify that. They assert that only a single diagnostic is returned for each case.

anurse: I'll enable MEL0005 on the numeric ones just to ensure it doesn't accidentally happen

pakrym:
Nice work, we should start enabling these in our repositories.

anurse:
I'll probably be enabling this in some SignalR projects assuming I can merge this before my big PR I'm making there ;). Already using them from my local build.

anurse:
:up: :date: it now supports `LoggerMessage.Define` (for checking format tokens and parameter counts match up with Generic parameters)

test/Microsoft.Extensions.Logging.Analyzer.Test/FormatStringAnalyzerTests.cs
```diff
@@ -12,60 +13,90 @@ namespace Microsoft.Extensions.Logging.Analyzer.Test
...
+            Assert.Equal("MEL0003", diagnostic.Id);
         }
 
         [Theory]
-        [MemberData(nameof(GenerateTemplateUsages), @"""{string}"", 1, 2")]
-        [MemberData(nameof(GenerateTemplateUsages), @"""{str"" + ""ing}"", 1, 2")]
-        [MemberData(nameof(GenerateTemplateUsages), @"""{"" + nameof(ILogger) + ""}""")]
-        [MemberData(nameof(GenerateTemplateUsages), @"""{"" + Const + ""}""")]
-        public void DiagnosticIsProducedForFormatArgumentCountMismatch(string format)
+        [InlineData(@"LoggerMessage.Define<int>(LogLevel.Information, 42, ""{One} {Two} {Three}"");")]
```
pakrym: Do we have no argument Define?

test/Microsoft.Extensions.Logging.Analyzer.Test/FormatStringAnalyzerTests.cs
```diff
@@ -12,60 +13,90 @@ namespace Microsoft.Extensions.Logging.Analyzer.Test
...
         }
 
         [Theory]
-        [MemberData(nameof(GenerateTemplateUsages), @"$""{string.Empty}""")]
-        [MemberData(nameof(GenerateTemplateUsages), @"""string"" + 2")]
-        public void DiagnosticIsProducedForDynamicFormatArgument(string format)
+        [MemberData(nameof(GenerateTemplateUsages), @"""{string}""", "1, 2", false)]
+        [MemberData(nameof(GenerateTemplateUsages), @"""{str"" + ""ing}""", "1, 2", false)]
+        [MemberData(nameof(GenerateTemplateUsages), @"""{"" + nameof(ILogger) + ""}""", "", false)]
+        [MemberData(nameof(GenerateTemplateUsages), @"""{"" + Const + ""}""", "", false)]
```
pakrym: Why false here?

test/Microsoft.Extensions.Logging.Analyzer.Test/FormatStringAnalyzerTests.cs
```diff
@@ -12,60 +13,90 @@ namespace Microsoft.Extensions.Logging.Analyzer.Test
...
-        public void DiagnosticNotIsProduced(string format)
+        // MEL0005 is not enabled by default.
+        [MemberData(nameof(GenerateTemplateUsages), @"""{camelCase}""", "1", true)]
+        public void TemplateDiagnosticsAreNotProduced(string format)
         {
             Assert.Empty(GetDiagnostics(format));
         }
 
-        public static IEnumerable<object[]> GenerateTemplateUsages(string templateAndArguments)
+        public static IEnumerable<object[]> GenerateTemplateUsages(string template, string arguments, bool includeDefineMessage)
```
pakrym: Feels cleaner to add another property to generate `.Define` usages.

test/Microsoft.Extensions.Logging.Analyzer.Test/FormatStringAnalyzerTests.cs
anurse: The next test is verifying Define cases, since they involve a different pattern (generating generic args instead of specifying expressions). It seemed cleaner.

anurse:
:up: :date: feedback addressed

Issue #629, opened 2017-06-01
Provide more generic Log() method


ddobric:
Right now we have several extension methods for logging defined in LoggerExtensions (LogWarning, LogError,..). 

Additionally we have a publc more generic method in ILogger

void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter);

That one is public, but very ugly to use. Moreover, it requires as to provide a formatter instead of using default one, which is not accessible.

Suggestion:
It would be good to have a following (additional) method in LoggerExtensions,cs

 public static void Log(this ILogger logger, string message, params object[] args)

davidfowl:
Just call `ILogger.LogInformation`?

ddobric:
Sorry David, I missed an important argument in the post above. Looking for generic helper method, which can be invoked  from multiple places in code to simplify complex formatting. Right now in every application we are implementing over and over again such or similar methods.
Here is an example:
```
  public static void Log(this ILogger logger, LogLevel level,  EventId eventId, Exception exception, string message, params object[] args)
        {
            if (logger == null)
            {
                throw new ArgumentNullException(nameof(logger));
            }

            logger.Log(level, eventId, new FormattedLogValues(message, args), exception, _messageFormatter);
        }

```

davidfowl:
@ddobric Why are you doing that? Why aren't you just calling the overloads specific to the thing you're trying to do (LogError, LogWarning, LogInformation). Can you show me specific examples of usages not covered by our existing extension methods? 

https://github.com/aspnet/Logging/blob/c821494678a30c323174bea8056f43b93a3ca6f4/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs

ddobric:
Because in larger applications, sometimes is better or even required that some components provide more generalized logging method.

For example. Take a look on following code. Arguments p1-pN are on component level same or similar or can be computed. Following simple statements would look to complex if they would have to provide such arguments.

```
LogWarning(b1,b2,b3, p1,..,pN)
..
LogInf(a1,a2,a3,p1,..pN)
```

So, they will always have to be refactored as:

```
myLog(LogLevel.Warning, b1,b2,b3);
..
myLog(LogLevel.Inf, a1,a2,a3);
```

By implementation of refactoring method we would need 'missing' log method:

```
private myLog(LogLevel level, ...)
{
   // add additional/common params here. at whatever logic to provide parameters.
    Log(level,...);
[...]

PR #790, opened 2018-03-26
Expand log level extension

ChrisSimmons:
This PR expands on #629.  Changes:

* Adds a new extension method that also accepts an `EventId` and an `Exception`
* Update [the previous method](https://github.com/ChrisSimmons/Logging/blob/a4b901159255abbdd47cf979171838799f971f66/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L350) to use this new method
* Update all the level-specific methods (e.g. [`LogDebug()`](https://github.com/ChrisSimmons/Logging/blob/a4b901159255abbdd47cf979171838799f971f66/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L42) and [`LogWarning()`](https://github.com/ChrisSimmons/Logging/blob/a4b901159255abbdd47cf979171838799f971f66/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L191)) to use this new method
* Use the new method to [handle the null `logger` argument exception](https://github.com/ChrisSimmons/Logging/blob/a4b901159255abbdd47cf979171838799f971f66/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L366)
    * If this is not desirable, e.g. for stacktrace reasons, I can revert and resubmit the PR

### DIFF
--- a/test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
+++ b/test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
@@ -623,338 +623,158 @@ namespace Microsoft.Extensions.Logging.Test
                 debug.Formatter(debug.State, debug.Exception));
         }
 
-        [Fact]
-        public void LogLevel_MessageOnly_LogsCorrectValues()
+        [Theory]
+        [InlineData(LogLevel.Trace)]
+        [InlineData(LogLevel.Information)]
+        [InlineData(LogLevel.Warning)]
+        [InlineData(LogLevel.Error)]
+        [InlineData(LogLevel.Critical)]
+        [InlineData(LogLevel.Debug)]
+        public void LogLevel_MessageOnly_LogsCorrectValues(LogLevel logLevel)
         {
             // Arrange
             var sink = new TestSink();
             var logger = SetUp(sink);
 
             // Act
-            logger.Log(LogLevel.Trace, _state);
-            logger.Log(LogLevel.Information, _state);
-            logger.Log(LogLevel.Warning, _state);
-            logger.Log(LogLevel.Error, _state);
-            logger.Log(LogLevel.Critical, _state);
-            logger.Log(LogLevel.Debug, _state);
+            logger.Log(logLevel, _state);
 
             // Assert
-            Assert.Equal(6, sink.Writes.Count);
-
-            Assert.True(sink.Writes.TryTake(out var trace));
-            Assert.Equal(LogLevel.Trace, trace.LogLevel);
-            Assert.Equal(_state, trace.State.ToString());
-            Assert.Equal(0, trace.EventId);
-            Assert.Null(trace.Exception);
-
-            Assert.True(sink.Writes.TryTake(out var information));
-            Assert.Equal(LogLevel.Information, information.LogLevel);
-            Assert.Equal(_state, information.State.ToString());
-            Assert.Equal(0, information.EventId);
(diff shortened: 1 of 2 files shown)


Eilon:
@muratg - do we want to have someone take a look at this?

muratg:
@ChrisSimmons thanks for the PR!

@pakrym could you review when you have some time?

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
```diff
@@ -38,10 +38,15 @@ public void MessageOnly_LogsCorrectValues()
             logger.LogError(_state);
             logger.LogCritical(_state);
             logger.LogDebug(_state);
+            logger.Log(LogLevel.Trace, _state);
```
pakrym: Add a separate test method that calls new overloads (with exception and non zero event id)

pakrym:
Please add tests for new overloads, otherwise LGTM.

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
ChrisSimmons: Sorry for the delay.  I'll get to this now.

ChrisSimmons:
I rebased and squashed commits, which blew away your previous comments, @pakrym:

> Add a separate test method that calls new overloads (with exception and non zero event id)

I've added separate tests for the new overloads, one for each of the 10 "flavors" of the equivalent `logger.LogLEVELNAME()` extensions.  Each uses a `LogLevel_` prefix to the test method name ([Example](https://github.com/aspnet/Logging/pull/790/files#diff-099caadefe774232f5d421983c3ee853R627)).  Let me know if that naming works.

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
ChrisSimmons: I've added separate tests for the new overloads, one for each of the 10 "flavors" of the equivalent `logger.LogLEVELNAME()` extensions.  Each uses a `LogLevel_` prefix to the test method name ([Example](https://github.com/aspnet/Logging/pull/790/files#diff-099caadefe774232f5d421983c3ee853R627)).  Let me know if that naming works.

(FYI, I rebased and squashed commits, @pakrym.)

ChrisSimmons:
Also, is there any value to using a `[Theory]` to simplify these tests?  See [this](https://github.com/ChrisSimmons/Logging/commit/a113909bda2c685938e0c80a4469add8d364afcb?diff=split), for example.

pakrym:
Totally, bring the Theories in.

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
```diff
@@ -628,6 +623,267 @@ public void LogValuesEventIdAndError_LogsCorrectValues()
...
+            var testLogValues = new TestLogValues()
+            {
+                Value = 1
+            };
+
+            // Act
+            logger.Log(logLevel, eventId, _exception, testLogValues.ToString());
+
+            // Assert
+            Assert.True(sink.Writes.TryTake(out var write));
```
pakrym: Assert.Single(sink.Writes);

PR #792, opened 2018-03-26
Fix doc examples

### LINKED ISSUE #791: Some documentation examples do not match in LoggerExtensions
See [this](https://github.com/aspnet/Logging/blob/fb0103e6b7cc7987dbfc1cc91f7a20b14eaa4937/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs#L81), for example, where the example is for one of the other methods.

I will have a PR to follow shortly.

ChrisSimmons:
Fixes #791


ChrisSimmons:
D'Oh!  Already being done on #786

PR #795, opened 2018-03-28
 Add long path and collision handling to test logger

JunTaoLuo:
Some ideas for improvements while working on https://github.com/aspnet/KestrelHttpServer/pull/2433

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -18,6 +18,8 @@ namespace Microsoft.Extensions.Logging.Testing
     public class AssemblyTestLog : IDisposable
     {
         public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
+        private static readonly string LogFileExtension = ".log";
+        private static readonly int MaxPathLength = 245;
 
         private static readonly object _lock = new object();
         private static readonly Dictionary<Assembly, AssemblyTestLog> _logs = new Dictionary<Assembly, AssemblyTestLog>();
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -84,7 +86,45 @@ public IServiceProvider CreateLoggerServices(ITestOutputHelper output, string cl
...
+                    testName = testName.Substring(0, testNameLength / 2) + testName.Substring(testName.Length - testNameLength / 2, testNameLength / 2);
+                }
+
+                var testOutputFile = Path.Combine(testOutputDirectory, $"{testName}{LogFileExtension}");
+
+                if (File.Exists(testOutputFile))
+                {
+                    output.WriteLine($"Warning: Output log file {testOutputFile} already exists. Please try to keep log file names unique.");
+
+                    for (var i = 0; i < 1000; i++)
```
anurse: This should be rare, and exceptional (i.e. the test isn't properly named) right?

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
```diff
@@ -51,70 +53,124 @@ public void TestLogWritesToITestOutputHelper()
         }
 
         [Fact]
-        public async Task TestLogWritesToGlobalLogFile()
+        public Task TestLogWritesToGlobalLogFile() =>
+            RunTestLogFunctionalTest((tempDir, loggerFactory) =>
```
anurse: Not sure I like this pattern of using an expression body and then putting the real thing in a lambda.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
JunTaoLuo: I think it will happen quite often as theory based tests might not end up with unique names. For example if one of the parameter is a string and the test uses string.Empty and null as parameters. It is very easy to overlook but we still probably want to capture the logs. The big downside to this is that if you run the same test repeatedly, the old log files will no longer be overwritten and you'll end up with a long list of test files.

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
JunTaoLuo: Testing out @pakrym's suggestion from https://github.com/aspnet/KestrelHttpServer/pull/2433#issuecomment-376751160

pakrym: @anurse but you can put everything on one line this way :)

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
```diff
@@ -51,70 +52,135 @@ public void TestLogWritesToITestOutputHelper()
...
-                    catch
+                }
+            }
+            logger.LogInformation("Finished test log in {baseDirectory}", tempDir);
+
+            // The first log file exists
+            Assert.True(File.Exists(Path.Combine(tempDir, "FakeTestAssembly", RuntimeInformation.FrameworkDescription.TrimStart('.'), "FakeTestClass", $"FakeTestName.log")));
+
+            // Subsequent files exist
+            for (var i = 0; i < 9; i++)
```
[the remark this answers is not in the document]
JunTaoLuo: This is just a test that the mechanism works, the actual limit is 999 https://github.com/aspnet/Logging/pull/795/files#diff-908d8b665df5cf7406c4b895ee0fc59dR114

PR #798, opened 2018-04-09
Implement file logging with default xunit tests

JunTaoLuo:
This is needed for https://github.com/aspnet/KestrelHttpServer/issues/2390:

The usage is:

1. Extend base class of `LoggedTest`
```c#
    public class MyTests : LoggedTest
    {
    }
```
2. Use the properties of the base class `LoggedTest`
```c#
        [Fact]
        public void MyTest()
        {
            // using LoggerFactory
            using (var webhost = WebHost.CreateDefaultBuilder().ConfigureServices(AddTestLogging).Build())
            {
                //...
            }
            // using Logger
            Logger.LogInformation("Hello world");
            // using TestOutputHelper
           new XunitLogger(TestOutpuHelper, nameof(MyClass), LogLevel.Information);
            // using TestSink
            TestSink.Writes.ForEach(c => Assert.True(c.LogLevel >= LogLeve.Debug));
        }
```
3. The verbosity can be controlled via the LogLevelAttribute:
```c#
        [Fact]
        [LogLevel(LogLevel.Information)]
        public void MyTest()
        {
            // This will noop
            Logger.Debug("Hello world");
        }
```

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -5,9 +5,11 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Text;
 using Microsoft.Extensions.DependencyInjection;
 using Serilog;
 using Serilog.Extensions.Logging;
@@ -20,6 +22,14 @@ namespace Microsoft.Extensions.Logging.Testing
         public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
         private static readonly string LogFileExtension = ".log";
         private static readonly int MaxPathLength = 245;
+        private static char[] InvalidFileChars = new char[]
+        {
+            '\"', '<', '>', '|', '\0',
+            (char)1, (char)2, (char)3, (char)4, (char)5, (char)6, (char)7, (char)8, (char)9, (char)10,
+            (char)11, (char)12, (char)13, (char)14, (char)15, (char)16, (char)17, (char)18, (char)19, (char)20,
+            (char)21, (char)22, (char)23, (char)24, (char)25, (char)26, (char)27, (char)28, (char)29, (char)30,
+            (char)31, ':', '*', '?', '\\', '/', ' ', (char)127
+        };
 
         private static readonly object _lock = new object();
         private static readonly Dictionary<Assembly, AssemblyTestLog> _logs = new Dictionary<Assembly, AssemblyTestLog>();
@@ -42,9 +52,12 @@ namespace Microsoft.Extensions.Logging.Testing
         public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, [CallerMemberName] string testName = null) =>
             StartTestLog(output, className, out loggerFactory, LogLevel.Debug, testName);
 
-        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel, [CallerMemberName] string testName = null)
+        public IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel, [CallerMemberName] string testName = null) =>
+            StartTestLog(output, className, out loggerFactory, minLogLevel, out var _, testName);
+
+        internal IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel, out string resolvedTestName, [CallerMemberName] string testName = null)
         {
-            var serviceProvider = CreateLoggerServices(output, className, minLogLevel, testName);
+            var serviceProvider = CreateLoggerServices(output, className, minLogLevel, out resolvedTestName, testName);
             var factory = serviceProvider.GetRequiredService<ILoggerFactory>();
             loggerFactory = factory;
             var logger = loggerFactory.CreateLogger("TestLifetime");
@@ -72,11 +85,13 @@ namespace Microsoft.Extensions.Logging.Testing
 
         public ILoggerFactory CreateLoggerFactory(ITestOutputHelper output, string className, LogLevel minLogLevel, [CallerMemberName] string testName = null)
         {
-            return CreateLoggerServices(output, className, minLogLevel, testName).GetRequiredService<ILoggerFactory>();
+            return CreateLoggerServices(output, className, minLogLevel, out var _, testName).GetRequiredService<ILoggerFactory>();
         }
 
-        public IServiceProvider CreateLoggerServices(ITestOutputHelper output, string className, LogLevel minLogLevel, [CallerMemberName] string testName = null)
+        public IServiceProvider CreateLoggerServices(ITestOutputHelper output, string className, LogLevel minLogLevel, out string normalizedTestName, [CallerMemberName] string testName = null)
         {
+            normalizedTestName = string.Empty;
+
             // Try to shorten the class name using the assembly name
             if (className.StartsWith(_assemblyName + "."))
             {
@@ -87,6 +102,7 @@ namespace Microsoft.Extensions.Logging.Testing
             if (!string.IsNullOrEmpty(_baseDirectory))
             {
                 var testOutputDirectory = Path.Combine(GetAssemblyBaseDirectory(_assemblyName, _baseDirectory), className);
+                testName = RemoveIllegalFileChars(testName);
 
                 if (testOutputDirectory.Length + testName.Length + LogFileExtension.Length >= MaxPathLength)
                 {
@@ -118,11 +134,13 @@ namespace Microsoft.Extensions.Logging.Testing
                         if (!File.Exists(testOutputFile))
                         {
                             _globalLogger.LogWarning($"To resolve log file collision, the enumerated file {testOutputFile} will be used.");
+                            testName = $"{testName}.{i}";
                             break;
                         }
                     }
                 }
 
+                normalizedTestName = testName;
                 serilogLoggerProvider = ConfigureFileLogging(testOutputFile);
             }
 
@@ -232,6 +250,17 @@ namespace Microsoft.Extensions.Logging.Testing
                 .WriteTo.File(fileName, outputTemplate: "{Timestamp:yyyy-MM-dd HH:mm:ss.fff zzz} [{SourceContext}] [{Level}] {Message}{NewLine}{Exception}", flushToDiskInterval: TimeSpan.FromSeconds(1), shared: true)
                 .CreateLogger();
             return new SerilogLoggerProvider(serilogger, dispose: true);
+        }
+
+        private static string RemoveIllegalFileChars(string s)
+        {
+            var sb = new StringBuilder();
+
+            foreach (var c in s)
+            {
+                sb.Append(InvalidFileChars.Contains(c) ? '_' : c);
+            }
+            return sb.ToString();
         }
 
         public void Dispose()
--- a/src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
+++ b/src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Helpers for writing tests that use Microsoft.Extensions.Logging. Contains null implementations of the abstractions that do nothing, as well as test implementations that are observable.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <PackageTags>$(PackageTags);testing</PackageTags>
     <EnableApiCheck>false</EnableApiCheck>
@@ -14,11 +14,16 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionPackageVersion)" />
     <PackageReference Include="xunit.abstractions" Version="$(XunitAbstractionsPackageVersion)" />
+    <PackageReference Include="xunit.extensibility.execution" Version="$(XunitExtensibilityExecutionPackageVersion)" />
     <PackageReference Include="xunit.assert" Version="$(XunitAssertPackageVersion)" />
     <PackageReference Include="Serilog.Extensions.Logging" Version="$(SerilogExtensionsLoggingPackageVersion)" />
     <PackageReference Include="Serilog.Sinks.File" Version="$(SerilogSinksFilePackageVersion)" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Content Include="build\**\*.props" PackagePath="%(Identity)" />
+  </ItemGroup>
 </Project>
--- a/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
@@ -1,0 +1,41 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedConditionalTheoryDiscoverer : LoggedTheoryDiscoverer
+    {
+        public LoggedConditionalTheoryDiscoverer(IMessageSink diagnosticMessageSink)
+            : base(diagnosticMessageSink)
+        {
+        }
+
+        protected override IEnumerable<IXunitTestCase> CreateTestCasesForTheory(
+            ITestFrameworkDiscoveryOptions discoveryOptions,
+            ITestMethod testMethod,
+            IAttributeInfo theoryAttribute)
+        {
+            var skipReason = testMethod.EvaluateSkipConditions();
+            return skipReason != null
+               ? new[] { new SkippedTestCase(skipReason, DiagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), testMethod) }
+               : base.CreateTestCasesForTheory(discoveryOptions, testMethod, theoryAttribute);
+        }
+
+        protected override IEnumerable<IXunitTestCase> CreateTestCasesForDataRow(
+            ITestFrameworkDiscoveryOptions discoveryOptions,
+            ITestMethod testMethod, IAttributeInfo theoryAttribute,
+            object[] dataRow)
+        {
+            var skipReason = testMethod.EvaluateSkipConditions();
+            return skipReason != null
+                ? base.CreateTestCasesForSkippedDataRow(discoveryOptions, testMethod, theoryAttribute, dataRow, skipReason)
+                : base.CreateTestCasesForDataRow(discoveryOptions, testMethod, theoryAttribute, dataRow);
+        }
+
+    }
+}
--- a/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestCaseRunner.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestCaseRunner.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Threading;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedTestCaseRunner : XunitTestCaseRunner
+    {
+        public LoggedTestCaseRunner(
+            IXunitTestCase testCase,
+            string displayName,
+            string skipReason,
+            object[] constructorArguments,
+            object[] testMethodArguments,
+            IMessageBus messageBus,
+            ExceptionAggregator aggregator,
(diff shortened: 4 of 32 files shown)


src/Microsoft.Extensions.Logging.Testing/Xunit/Shared/SkippedTestCase.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.AspNetCore.Testing.xunit
```
JunTaoLuo: Going to make this type public from testing instead of duplicating here.

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        public LoggedTestXunitTests(ITestOutputHelper output) : base(output)
+        {
+        }
+
+        [Fact]
+        public void LoggedFactGetsInitializedLoggerFactory()
+        {
+            Assert.NotNull(LoggerFactory);
+            LoggerFactory.CreateLogger(nameof(AssemblyTestLogTests)).LogInformation("Hello world");
+            LoggerFactory.CreateLogger(nameof(AssemblyTestLogTests)).LogDebug("Debug Hello world");
```
JunTaoLuo: TODO: add more tests

src/Microsoft.Extensions.Logging.Testing/Microsoft.Extensions.Logging.Testing.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Helpers for writing tests that use Microsoft.Extensions.Logging. Contains null implementations of the abstractions that do nothing, as well as test implementations that are observable.</Description>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
```
JunTaoLuo: I needed to cross-target to enable test discovery on full framework.

src/Microsoft.Extensions.Logging.Testing/Xunit/Shared/SkippedTestCase.cs
JunTaoLuo: https://github.com/aspnet/Testing/pull/301

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -20,6 +22,15 @@ public class AssemblyTestLog : IDisposable
...
         private static readonly int MaxPathLength = 245;
+        private static IEnumerable<char> InvalidFileChars =
+            Path.GetInvalidPathChars()
+            .Union(Path.GetInvalidFileNameChars())
+            .Union(new char[] {
+                ' ', // space
+                '\\', // back slash
+                '/', // forward slash
+                '\x7F', // delete
+            });
```
pakrym: ToArray

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -234,6 +253,26 @@ private static SerilogLoggerProvider ConfigureFileLogging(string fileName)
...
 
+        private static string RemoveIllegalFileChars(string s)
+        {
+            var sb = new StringBuilder();
+
+            foreach (var c in s)
+            {
+                if (InvalidFileChars.Contains(c))
+                {
+                    var escapedChar = Convert.ToByte(c).ToString("x");
```
pakrym: This expands string size a lot, maybe just do '_' ?

JunTaoLuo: I did do that at one point but it made test names highly unreadable. Also converting to the same character for all invalid chars creates lots of collisions. With the test name shortening logic in place, I don't think this should be an issue.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestFrameworkDiscoverer.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedTestFrameworkDiscoverer : XunitTestFrameworkDiscoverer
+    {
+        private List<(Type Type, IXunitTestCaseDiscoverer Discoverer)> Discoverers { get; }
```
anurse: `Dictionary`?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestCaseRunner.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Reflection;
+using System.Threading;
+using Xunit.Abstractions;
+using Xunit.Sdk;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedTestCaseRunner : XunitTestCaseRunner
+    {
+        public LoggedTestCaseRunner(IXunitTestCase testCase, string displayName, string skipReason, object[] constructorArguments, object[] testMethodArguments, IMessageBus messageBus, ExceptionAggregator aggregator, CancellationTokenSource cancellationTokenSource) : base(testCase, displayName, skipReason, constructorArguments, testMethodArguments, messageBus, aggregator, cancellationTokenSource)
```
pakrym: Linebreaks? This line spans 2 monitors )

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        public LoggedTestInvoker(ITest test, IMessageBus messageBus, Type testClass, object[] constructorArguments, MethodInfo testMethod, object[] testMethodArguments, IReadOnlyList<BeforeAfterTestAttribute> beforeAfterAttributes, ExceptionAggregator aggregator, CancellationTokenSource cancellationTokenSource) : base(test, messageBus, testClass, constructorArguments, testMethod, testMethodArguments, beforeAfterAttributes, aggregator, cancellationTokenSource)
+        {
+        }
+
+        protected override object CreateTestClass()
+        {
+            var testClass = base.CreateTestClass();
+
+            if (testClass is LoggedTest loggedTestClass)
```
pakrym: Can't we do it for any class, so you're not forced to inherit and can just inject the constructor argument?

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging.Testing.Tests
+{
+    public class LoggedTestXunitTests : LoggedTest
+    {
+        public LoggedTestXunitTests(ITestOutputHelper output) : base(output)
```
pakrym: Do we still have to pass `ITestOutputHelper` around? Can ILoggerFactory be injected directly?

JunTaoLuo:
Note: WIP adding and fixing tests.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
JunTaoLuo: Hmm right now I expand illegal characters like `/` => `-x2f`, maybe I can reduce the verbosity and just leave it as `2f` which is a bit harder to parse out but still retains most of the information.

pakrym: > I did do that at one point but it made test names highly unreadable

Hex is not very readable too.

> With the test name shortening logic in place, I don't think this should be an issue. 

Replacing a bunch of '_' with one would be a nice addition to shortening logic.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                var classType = loggedTestClass.GetType();
+                var testOutputHelper = ConstructorArguments.Single(a => typeof(ITestOutputHelper).IsAssignableFrom(a.GetType())) as ITestOutputHelper;
```
anurse: In theory we could get the `ITestOutputHelper` from the source (whatever that is) rather than extracting it from constructor arguments.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -0,0 +1,39 @@
...
+                var logLevelAttribute = TestMethod.GetCustomAttribute(typeof(LogLevelAttribute)) as LogLevelAttribute;
+                var testName = TestMethodArguments.Aggregate(TestMethod.Name, (a, b) => $"{a}_{(b ?? "null")}");
```
anurse: `string.Join('_', ...)`?

pakrym:
Supporting https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Testing/ITestSink.cs would be interesting too, would make some kestrel tests even simpler

anurse:
Ooh, yeah, automatically register a TestLogger and TestSink. I like that. If it's simple, do it, if it's more than a couple minutes work, file a bug and move on :)

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestFrameworkDiscoverer.cs
JunTaoLuo: Ah yes, at one point, I was using a best match on the type which needed a list of discoverers for types from the most specific to the most general. Now that I'm using an exact match on the type and falling back to non-logged discoverers, this can totally be just a dictionary.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
JunTaoLuo: Poor choice of words, I meant file name collisions instead of readability. Even though we have collision resolution, I would prefer retaining as much information as possible regarding the arguments used by default.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
JunTaoLuo: We are not injecting it as a constructor argument. We are setting the property of the base class LoggedTest.

JunTaoLuo: Unless you are suggesting we change the design?

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
pakrym: Do you really prefer `Testing_https:-x2f-x2flocalhost` over `Testing_https:_localhost` ? To me, the second one is much easier to figure out.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
pakrym: I always thought we are going argument injection, that's why we started with argument injection DI prototype.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
JunTaoLuo: For that case maybe, but this will create many collisions for kestrel tests, for example: https://github.com/aspnet/KestrelHttpServer/blob/dev/test/shared/HttpParsingData.cs#L192. Also I don't like coalescing multiple `_` since that makes tests like https://github.com/aspnet/KestrelHttpServer/blob/dev/test/shared/HttpParsingData.cs#L254-L255 indistiguishable.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
JunTaoLuo: Closest I can come up with is `string.Join("_", TestMethod.Name, TestMethodArguments)` but that won't be able to mark `null` object with `"null"` string

JunTaoLuo: From adding tests previously, I remember it being useful to distinguish between null strings and empty strings for example. That's why I special cased null here.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -20,6 +22,15 @@ public class AssemblyTestLog : IDisposable
         public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
         private static readonly string LogFileExtension = ".log";
         private static readonly int MaxPathLength = 245;
+        private static char[] InvalidFileChars =
+            Path.GetInvalidPathChars()
+            .Union(Path.GetInvalidFileNameChars())
+            .Union(new char[] {
+                ' ', // space
+                '\\', // back slash
+                '/', // forward slash
```
halter73: Isn't this already an InvalidFileNameChar on every platform? Maybe a whitelist would be better.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
```diff
@@ -0,0 +1,35 @@
...
+            : base(diagnosticMessageSink)
+        {
+        }
+
+        protected override IEnumerable<IXunitTestCase> CreateTestCasesForTheory(ITestFrameworkDiscoveryOptions discoveryOptions, ITestMethod testMethod, IAttributeInfo theoryAttribute)
+        {
+            var skipReason = testMethod.EvaluateSkipConditions();
+            return skipReason != null
+               ? new[] { new SkippedTestCase(skipReason, DiagnosticMessageSink, discoveryOptions.MethodDisplayOrDefault(), testMethod) }
+               : base.CreateTestCasesForTheory(discoveryOptions, testMethod, theoryAttribute);
```
halter73: Does the base class not call EvaluateSkipConditions?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
halter73: 👍 for foldl

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
JunTaoLuo: I think so. I prefer not using a whitelist since that probably takes more work.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
JunTaoLuo: What is foldl?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
JunTaoLuo: Not LoggedTheoryDiscoverer.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
pakrym: @halter73 what would you prefer?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
halter73: But isn't that what you're replacing? If this didn't need to call EvaluateSkipConditions before, why now?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
halter73: The hipster name for Aggregate.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
JunTaoLuo: No I'm replacing ConditionalTheoryDiscoverer. But this inherits from LoggedTheoryDiscoverer which doesn't call EvaluateSkipConditions.

halter73: Wouldn't ConditionalTheoryDiscoverer have it's own LoggedTheoryDiscoverer that does exactly this then? Why not use that?

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitTests.cs
JunTaoLuo: Actually I don't think we can remove ITestOutputHelper. What if the user wants to initialize a different LoggerFactory by calling StartLog (to override test name for example)? That call needs the ITestOutputHelper.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
halter73: I think I misunderstood. Could you derive from ConditionalTheoryDiscoverer?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
JunTaoLuo: Found a way of removing the need to flow the ITestOutputHelper with constructors.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedConditionalTheoryDiscoverer.cs
JunTaoLuo: Then I'll still need to override these two methods to create `LoggedTestCase` or `LoggedTheoryTestCase`. I don't see the benefit here.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -20,6 +22,14 @@ public class AssemblyTestLog : IDisposable
         public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
         private static readonly string LogFileExtension = ".log";
         private static readonly int MaxPathLength = 245;
+        private static char[] InvalidFileChars = new char[]
+        {
+            '\"', '<', '>', '|', '\0',
```
JunTaoLuo: Created a static blacklist by concatenating https://github.com/dotnet/coreclr/blob/f5ae6d14d9d49dbb9f4e97f605d2a5cceab04879/src/mscorlib/shared/System/IO/Path.Unix.cs#L13-L15 and https://github.com/dotnet/coreclr/blob/9ac41819ab05282798c498b373a5017b1c2c4e64/src/mscorlib/shared/System/IO/Path.Windows.cs#L12-L28 
as well as space and delete.

pakrym:
Can we bring `[assembly: Xunit.TestFramework("Microsoft.Extensions.Logging.Testing.LoggedTestFramework", "Microsoft.Extensions.Logging.Testing")]` via MSBuild target?

pakrym:
We can also add a method to `LoggetTest` to change
```
 [Fact]
        public void MyTest()
        {
            // using LoggerFactory
            using (var webhost = WebHost.CreateDefaultBuilder().ConfigureServices(collection => collection.AddSingleton(LoggerFactory)).Build())
            {
                //...
            }

```
to
```
[Fact]
        public void MyTest()
        {
            // using LoggerFactory
            using (var webhost = WebHost.CreateDefaultBuilder().ConfigureServices(AddTestLogging).Build())
            {
                //...
            }
```
Makes it a bit shorter and simpler.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -234,6 +252,17 @@ private static SerilogLoggerProvider ConfigureFileLogging(string fileName)
             return new SerilogLoggerProvider(serilogger, dispose: true);
         }
 
+        private static string RemoveIllegalFileChars(string s)
+        {
+            var sb = new StringBuilder();
+
+            foreach (var c in s)
+            {
+                sb.Append(InvalidFileChars.Contains(c) ? Convert.ToByte(c).ToString("x") : $"{c}");
```
pakrym: Not when we have space in the invalid character list, please, let's not use hex, a bunch of random lower case characters would make file names completely unreadable.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                // Try resolving ITestOutputHelper from constructor arguments
+                loggedTestClass.TestOutputHelper = ConstructorArguments?.SingleOrDefault(a => typeof(ITestOutputHelper).IsAssignableFrom(a.GetType())) as ITestOutputHelper;
+
+                // None resolved so create a new one and retain a reference to it for initialization/uninitialization
+                if (loggedTestClass.TestOutputHelper == null)
+                {
+                    loggedTestClass.TestOutputHelper = _output = new TestOutputHelper();
```
pakrym: Don't you have to call
 https://github.com/xunit/xunit/blob/bccfcccf26b2c63c90573fe1a17e6572882ef39c/src/xunit.execution/Sdk/Frameworks/Runners/XunitTestRunner.cs#L65
 for `TestOutputHelper` to work?

pakrym:
Couple of comments then LGTM, let's try to use it.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
pakrym: Only append underscore if the previous character is not underscore. `A_B` is better than `A_____B` and might help long paths.

JunTaoLuo:
Addressed the feedback items.

anurse:
🎉

PR #799, opened 2018-04-09
Harden logging tests

pakrym:
Recent logging failures were caused by tests logging to TraceSource so I fixed that.

While investigating failures I thought it was caused by an error being logged to console so redirected console output too.


natemcmaster:
If only we had an `IConsole` for mocking :trollface:

PR #803, opened 2018-04-11
Addtional setup step for LoggedTest

JunTaoLuo:
Is this what you had in mind?


src/Microsoft.Extensions.Logging.Testing/LoggedTest.cs
```diff
@@ -39,5 +39,7 @@ public IDisposable StartLog(out ILoggerFactory loggerFactory, LogLevel minLogLev
         {
             return AssemblyTestLog.ForAssembly(GetType().GetTypeInfo().Assembly).StartTestLog(TestOutputHelper, GetType().FullName, out loggerFactory, minLogLevel, testName);
         }
+
+        public virtual void AdditionalSetup() { }
```
pakrym: Name it somehow logger related.

JunTaoLuo: Well it doesn't have to be logger related setup but I'll name it `AdditionalLoggerSetup`

Issue #797, opened 2018-04-03
TestSink should be thread safe


halter73:
aspnet/KestrelHttpServer#2427

Eilon:
Is that even possible? The API looks a bit weak to enable that:

https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Testing/TestSink.cs#L26-L28

We could add thread-safety guards to the *methods* that operate on the properties, but guarding the properties doesn't seem possible without a breaking change. So, even if we guard *one* way of doing writes, we can't guard all writes, and we can guard none of the reads.

Eilon:
If what I said is right, an alternative would be to create a new set of thread-safe types (ThreadSafeTestSink, etc.), and design those from the ground up to be safe...

halter73:
I was thinking we could expose the Scopes and Writes properties as ConcurrentQueues instead of Lists. Enumerating a ConcurrentQueue takes a snapshot so it's possible to see if something was logged previously while new logs are still being added. I agree that if we can't change the API, thread safety won't be possible.

halter73:
@muratg @JunTaoLuo mentioned he's already looking into this, so I assigned it.

JunTaoLuo:
I need to fix this after adding a TestSink to every LoggedTest by default. This blocks https://github.com/aspnet/KestrelHttpServer/issues/2390

muratg:
@JunTaoLuo to be closed today? :)

JunTaoLuo:
Yup. The PR is approved and preemptive reaction PRs have been merged. I just need to make sure I didn't break anything else in Universe.

PR #804, opened 2018-04-12
Fix concurrency issue

JunTaoLuo:
Found some concurrency issues while adding logging to Kestrel

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/ITestSink.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/ITestSink.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 
 namespace Microsoft.Extensions.Logging.Testing
 {
@@ -9,9 +9,9 @@ namespace Microsoft.Extensions.Logging.Testing
 
         Func<BeginScopeContext, bool> BeginEnabled { get; set; }
 
-        List<BeginScopeContext> Scopes { get; set; }
+        IProducerConsumerCollection<BeginScopeContext> Scopes { get; set; }
 
-        List<WriteContext> Writes { get; set; }
+        IProducerConsumerCollection<WriteContext> Writes { get; set; }
 
         void Write(WriteContext context);
 
--- a/src/Microsoft.Extensions.Logging.Testing/TestSink.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/TestSink.cs
@@ -2,12 +2,15 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 
 namespace Microsoft.Extensions.Logging.Testing
 {
     public class TestSink : ITestSink
     {
+        private ConcurrentQueue<BeginScopeContext> _scopes;
+        private ConcurrentQueue<WriteContext> _writes;
+
         public TestSink(
             Func<WriteContext, bool> writeEnabled = null,
             Func<BeginScopeContext, bool> beginEnabled = null)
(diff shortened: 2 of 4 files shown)


src/Microsoft.Extensions.Logging.Testing/TestSink.cs
```diff
@@ -31,15 +33,21 @@ public void Write(WriteContext context)
         {
             if (WriteEnabled == null || WriteEnabled(context))
             {
-                Writes.Add(context);
+                lock(_writeLock)
```
pakrym: ` `

JunTaoLuo:
Apparently this fixes #797

JunTaoLuo:
updated the implementation

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
```diff
@@ -501,47 +501,47 @@ public void LogValuesAndError_LogsCorrectValues()
...
+            sink.Writes.TryDequeue(out var critical);
             Assert.Equal(LogLevel.Critical, critical.LogLevel);
             Assert.Equal(0, critical.EventId);
             Assert.Equal(_exception, critical.Exception);
             Assert.Equal(
                 "Test 1",
                 critical.Formatter(critical.State, critical.Exception));
 
-            var debug = sink.Writes[5];
+            sink.Writes.TryDequeue(out var debug);
```
pakrym: Assert.True(

src/Microsoft.Extensions.Logging.Testing/ITestSink.cs
```diff
@@ -9,9 +10,9 @@ public interface ITestSink
 
         Func<BeginScopeContext, bool> BeginEnabled { get; set; }
 
-        List<BeginScopeContext> Scopes { get; set; }
+        ConcurrentQueue<BeginScopeContext> Scopes { get; set; }
```
pranavkm: ConcurrentQueue

JamesNK:
Will this change eliminate the need for this provider? https://github.com/aspnet/SignalR/blob/dev/test/Microsoft.AspNetCore.SignalR.Tests.Utils/ServerFixture.cs#L148-L206

JamesNK:
// @anurse

JunTaoLuo:
Though mostly the same, LogRecord also contains the date time which isn't included on the WriteContext. I think there's ways to work around it though with formatters and states.

pakrym:
> Can it not be added to WriteContext? I don't remember the "ship/noship" state of this package :)

It can, what do you use time for?

test/Microsoft.Extensions.Logging.Test/LoggerExtensionsTest.cs
```diff
@@ -40,46 +41,48 @@ public void MessageOnly_LogsCorrectValues()
             logger.LogDebug(_state);
             logger.Log(LogLevel.Information, _state);
 
+            var writeList = sink.Writes.ToList();
```
Tratcher: Do you expect tests elsewhere to be written like this? Or should they be popping items off the front of the list?

JunTaoLuo: Only if you need to check the order. I see most of the tests use something like Assert.Empty(sink.Writes), Assert.Single(sink.Writes, predicate) or Assert.Contains(sink.Writes, w => predicate) which works with just the enumerable interface. Enumerating as a list is the simplest for test back compat for now but using the TryTake method could also work.

Tratcher: This test should reflect our recommendations for other repos.

pakrym: I liked the TryTake version of this, much easier to maintain.

src/Microsoft.Extensions.Logging.Testing/ITestSink.cs
```diff
@@ -9,9 +9,9 @@ public interface ITestSink
 
         Func<BeginScopeContext, bool> BeginEnabled { get; set; }
 
-        List<BeginScopeContext> Scopes { get; set; }
+        IProducerConsumerCollection<BeginScopeContext> Scopes { get; set; }
```
pranavkm: 👍

PR #806, opened 2018-04-12
Add azure logging change token source unconditionally

pakrym:
https://github.com/aspnet/AzureIntegration/issues/177

It was regressed by https://github.com/aspnet/Logging/pull/719, `loggerBuilder.AddConfiguration` adds some services that we call `TryAddEnumerable` on afterwards. It causes us not to add required configuration change token source.

### DIFF
--- a/src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesLoggerFactoryExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesLoggerFactoryExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging.AzureAppServices;
 using Microsoft.Extensions.Logging.AzureAppServices.Internal;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesLoggerFactoryExtensions.cs
```diff
@@ -30,30 +38,42 @@ public static ILoggingBuilder AddAzureWebAppDiagnostics(this ILoggingBuilder bui
...
+                services.AddSingleton(context);
+                services.AddSingleton<IOptionsChangeTokenSource<LoggerFilterOptions>>(
+                    new ConfigurationChangeTokenSource<LoggerFilterOptions>(config));
+            }
 
-            services.TryAddEnumerable(Singleton<IWebAppContext>(context));
+            if (addedFileLogger)
+            {
+                services.AddSingleton<IConfigureOptions<LoggerFilterOptions>>(CreateFileFilterConfigureOptions(config));
+                services.AddSingleton<IConfigureOptions<AzureBlobLoggerOptions>>(new BlobLoggerConfigureOptions(config, context));
```
BrennanConroy: I think you swapped this with the other one

src/Microsoft.Extensions.Logging.AzureAppServices/AzureAppServicesLoggerFactoryExtensions.cs
```diff
@@ -30,30 +38,42 @@ public static ILoggingBuilder AddAzureWebAppDiagnostics(this ILoggingBuilder bui
...
+                services.AddSingleton<IOptionsChangeTokenSource<AzureBlobLoggerOptions>>(
```
BrennanConroy: Same with this one

Issue #819, opened 2018-04-27
Xunit logger throws when class is not taking ITestOutputHelper 


pakrym:
`StartTestLog` would always throw when we create `TestOutputHelper` and don't call `Initialize` on it.

https://github.com/aspnet/Logging/blob/a8115c9660357e7032e9c680e5459bf5299aadb6/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs#L72-L85

JunTaoLuo:
Initialization and uninitialization logic lives here: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs#L33-L51. I'm not sure why it's not working.

JunTaoLuo:
I'll see if there's a more robust way of doing it.

pakrym:
`StartTestLog` tries to log inline so initialization logic doesn't have a chance to run.

PR #822, opened 2018-05-01
Initialize ITestOutputHelper earlier

JunTaoLuo:
Addresses #819. Though this is a good thing to fix, I don't think this is causing the exception you are seeing. 

The exceptions that might occur are caught here: https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Testing/XunitLoggerProvider.cs#L90-L100. This explains why our tests were working and were writing to file. However, I'm no longer sure why debugging is broken on your machine.


pakrym:
I had first chance exceptions enabled.

PR #829, opened 2018-05-08
Add readonly to EventId and ConsoleColors structures

meziantou:
Add readonly to `EventId` and `ConsoleColors` structures.


meziantou:
The build fails, but it does not seem to be related to this change.

````
The specified framework 'Microsoft.NETCore.App', version '2.1.0-rc1-26425-06' was not found.
````

pakrym:
Thank you.

Issue #809, opened 2018-04-17
Shorten file logging date/time format


pakrym:
We can start by removing UTC offset.
Or go further and just log time span since the test start
/cc @JunTaoLuo

pakrym:
What about XUnit logs?

PR #831, opened 2018-05-14
Logging improvements

JunTaoLuo:
- Create separate LoggedTest class without TestSink
- Format test logs with timestamp for the first message and offset thereafter

Addresses feedback from usage of logging framework and #809.

The global logs now look like:
```
[2018-05-14T09:07:24] [GlobalTestLog] [Information] Global Test Logging initialized. Set the 'ASPNETCORE_TEST_LOG_DIR' Environment Variable in order to create log files on disk.
[0.034s] [GlobalTestLog] [Information] Starting test "ITestOutputHelperInitializedByDefault"
```
Test logs now look like:
```
[2018-05-14T09:07:24] [TestLifetime] [Information] Starting test "LoggedFactFilteredByLogLevel"
[0.001s] [Microsoft.Extensions.Logging.Testing.Tests.LoggedTestXunitTests] [Information] Information
```


src/Microsoft.Extensions.Logging.Testing/LoggedTest/LoggedTest.cs
```diff
@@ -0,0 +1,23 @@
...
+    public class LoggedTest : SinklessLoggedTest
+    {
+        // Obsolete but keeping for back compat
+        public LoggedTest(ITestOutputHelper output = null) : base (output) { }
+
+        public ITestSink TestSink { get; set; }
+
+        public override void AdditionalSetup()
+        {
+            base.AdditionalSetup();
```
JunTaoLuo: This means we will need to update the implementation in Kestrel at https://github.com/aspnet/KestrelHttpServer/blob/2d51d236979ee731b8675e5879a56b708ec2fa4b/test/Kestrel.FunctionalTests/TestHelpers/TestApplicationErrorLoggerLoggedTest.cs#L12

src/Microsoft.Extensions.Logging.Testing/LoggedTest/SinklessLoggedTest.cs
```diff
@@ -9,10 +9,10 @@
 
 namespace Microsoft.Extensions.Logging.Testing
 {
-    public abstract class LoggedTest
+    public class SinklessLoggedTest : ILoggedTest
```
JunTaoLuo: BaseLoggedTest? Other names? I kept the LoggedTest class as is to reduce churn in our code base.

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
```diff
@@ -46,10 +46,13 @@ public void TestLogWritesToITestOutputHelper()
...
             }
 
-            Assert.Equal(@"[TIMESTAMP] TestLifetime Information: Starting test TestLogWritesToITestOutputHelper
-[TIMESTAMP] TestLogger Information: Information!
-[TIMESTAMP] TestLifetime Information: Finished test TestLogWritesToITestOutputHelper in DURATION
-", MakeConsistent(output.Output), ignoreLineEndingDifferences: true);
+            var testLogContent = MakeConsistent(output.Output);
+
+            Assert.Equal(
+@"[XUNITTIMESTAMP] TestLifetime Information: Starting test TestLogWritesToITestOutputHelper at TIMESTAMP
```
JunTaoLuo: I did not update the xunit timestamp format that's displayed to the console upon failure. Do we want this too?

src/Microsoft.Extensions.Logging.Testing/LoggedTest/SinklessLoggedTest.cs
JunTaoLuo: The TestSink is removed on this type so we can reduce the memory used by the tests cc @Tratcher .

pakrym: It's unfortunate that a type with the nice name would be the wrong one to use most of the time.

JunTaoLuo: I don't think it's wrong to use LoggedTest. In fact I would imagine it should be the one you would use most of the time. You would only switch to SinklessLoggedTest if you are using a large amount of memory.

pakrym: You should use it only when asserting logs, there is no point in wasting resources otherwise.

src/Microsoft.Extensions.Logging.Testing/LoggedTest/ILoggedTest.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    internal interface ILoggedTest
+    {
+        ILogger Logger { get; set; }
+
+        ILoggerFactory LoggerFactory { get; set; }
+
+        ITestOutputHelper TestOutputHelper { get; set; }
+
+        IDisposable StartLog(out ILoggerFactory loggerFactory, string testName);
```
pakrym: Keep only one overload?

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -44,45 +44,46 @@ protected override object CreateTestClass()
         {
             var testClass = base.CreateTestClass();
 
-            if (testClass is LoggedTest loggedTestClass)
+            if (testClass is ILoggedTest loggedTest)
             {
-                var classType = loggedTestClass.GetType();
+                var classType = loggedTest.GetType();
```
pakrym: Can you move all non-xunit related logic into ILoggerTest implementation?

src/Microsoft.Extensions.Logging.Testing/LoggedTest/SinklessLoggedTest.cs
JunTaoLuo: Right, but that would mean I need to update all the repos that currently rely on LoggedTest for asserting logs. Not sure if it's really worth the effort to update them all.

src/Microsoft.Extensions.Logging.Testing/LoggedTest/LoggedTest.cs
```diff
@@ -0,0 +1,24 @@
...
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedTest : SinklessLoggedTest
+    {
+        // Obsolete but keeping for back compat
+        public LoggedTest(ITestOutputHelper output = null) : base (output) { }
+
+        public ITestSink TestSink { get; set; }
+
+        public override void Initialize(MethodInfo methodInfo, object[] testMethodArguments)
```
JunTaoLuo: Definitely a breaking change. Will need a rename in Kestrel.

src/Microsoft.Extensions.Logging.Testing/LoggedTest/ILoggedTest.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Reflection;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    internal interface ILoggedTest
```
pakrym: All properties `{ get; }` only.

pakrym: You can also drop the interface now when you have a base class again.

JunTaoLuo: TestOutputHelper is set via the Invoker so we need to at least maintain the setter on that one. I could remove the other ones but I guess we can just drop the interface completely.

JunTaoLuo: The interface offers an escape hatch if we want to change the logging framework behaviour without extending from the base class. If we had this, we would have been able to implement a SinklessLoggedClass in Kestrel instead of having to make a cross repo change here.

test/Microsoft.Extensions.Logging.Testing.Tests/AssemblyTestLogTests.cs
pakrym: Yes

src/Microsoft.Extensions.Logging.Testing/LoggedTest/LoggedTest.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Reflection;
+using Xunit.Abstractions;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    public class LoggedTest : SinklessLoggedTest
```
JunTaoLuo: LoggedTestBase? BaseLoggedTest? LoggedTestSlim?

pakrym: LoggedTestBase

PR #832, opened 2018-05-15
Skip member data by property

Tratcher:
Mirror of https://github.com/aspnet/Common/pull/354


pakrym:
I hate it a little bit. Can we do a source package or something else to carry common logic?

Tratcher:
Logged https://github.com/aspnet/Logging/issues/833 for all the duplication.

PR #834, opened 2018-05-16
Change the file logger directory to use tfm

JunTaoLuo:
Old Path: Logging\artifacts\logs\Microsoft.Extensions.Logging.Testing.Tests\NET Framework 4.7.2650.0\AssemblyTestLogTests\ForAssembly_ReturnsSameInstanceForSameAssembly.log
New Path: Logging\artifacts\logs\Microsoft.Extensions.Logging.Testing.Tests\net461\AssemblyTestLogTests\ForAssembly_ReturnsSameInstanceForSameAssembly.log

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -8,7 +8,6 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.CompilerServices;
-using System.Runtime.InteropServices;
 using System.Text;
 using Microsoft.Extensions.DependencyInjection;
 using Serilog;
@@ -39,15 +38,15 @@ namespace Microsoft.Extensions.Logging.Testing
         private readonly ILoggerFactory _globalLoggerFactory;
         private readonly ILogger _globalLogger;
         private readonly string _baseDirectory;
-        private readonly string _assemblyName;
+        private readonly Assembly _assembly;
         private readonly IServiceProvider _serviceProvider;
 
-        private AssemblyTestLog(ILoggerFactory globalLoggerFactory, ILogger globalLogger, string baseDirectory, string assemblyName, IServiceProvider serviceProvider)
+        private AssemblyTestLog(ILoggerFactory globalLoggerFactory, ILogger globalLogger, string baseDirectory, Assembly assembly, IServiceProvider serviceProvider)
         {
             _globalLoggerFactory = globalLoggerFactory;
             _globalLogger = globalLogger;
             _baseDirectory = baseDirectory;
-            _assemblyName = assemblyName;
+            _assembly = assembly;
             _serviceProvider = serviceProvider;
         }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -228,7 +228,7 @@ private static string GetAssemblyBaseDirectory(string assemblyName, string baseD
         {
             if (!string.IsNullOrEmpty(baseDirectory))
             {
-                return Path.Combine(baseDirectory, assemblyName, RuntimeInformation.FrameworkDescription.TrimStart('.'));
+                return Path.Combine(baseDirectory, assemblyName, new DirectoryInfo(AppContext.BaseDirectory).Name);
```
JunTaoLuo: A bit hacky because this relies on the AppContext.BaseDirectory to be in bin/{Config}/{tfm} but I don't know if there's a better way to find the TFM that's compatible with 4.6.1. Since we launch tests via xunit this should work.

pakrym: Or we can hardcode a dictionary :) How many frameworks do we run on 3-4?

JunTaoLuo: Hardcoding a dictionary is not ideal since when we change the TFMs, like we did for netcoreapp2.2, we'd also need to update the dictionary. I'll look into the other suggestion though.

pakrym: It happens once a year. And fallback would be there.

JunTaoLuo:
Originally I wanted to make a separate PR for setting the default file logging location but since I opted for the assembly attribute method of setting the TFM, I might as well include this change as well.

The default location is $(RepositoryRoot)/artifacts/logs. This can be overridden by specifying a different location via ASPNETCORE_TEST_LOG_DIR.

JunTaoLuo:
Also added fix to close file handles

src/Microsoft.Extensions.Logging.Testing/build/Microsoft.Extensions.Logging.Testing.props
```diff
@@ -1,8 +1,17 @@
...
+    Condition="'$(GenerateLoggingTestingAssemblyAttributes)' != 'false'">
+    <ItemGroup>
+      <AssemblyAttribute Include="Xunit.TestFramework">
+        <_Parameter1>Microsoft.Extensions.Logging.Testing.LoggedTestFramework</_Parameter1>
+        <_Parameter2>Microsoft.Extensions.Logging.Testing</_Parameter2>
+      </AssemblyAttribute>
+
+      <AssemblyAttribute Include="Microsoft.Extensions.Logging.Testing.TestFrameworkFileLoggerAttribute">
+        <_Parameter1>$(TargetFramework)</_Parameter1>
+        <_Parameter2 Condition="'$(RepositoryRoot)' != ''">$(RepositoryRoot)artifacts\logs\</_Parameter2>
```
pakrym: Turn `$(RepositoryRoot)artifacts\logs\` into conditional property.

PR #836, opened 2018-05-17
Add opt-out for file logging and add error when FileLoggerAttribute is missing

JunTaoLuo:
Couldn't find a way to ensure references to Logging.Testing are marked with PrivateAssets="None" so I'm opting to add a more explicit error instead. Also adding an environment variable to turn off file logging which is now on by default and there's no way of turning it off.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Extensions.Logging.Testing
 {
     public class AssemblyTestLog : IDisposable
     {
-        public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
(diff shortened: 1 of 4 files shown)


pakrym:
Add opt out via msbuild property.

JunTaoLuo:
So far the pattern is environment based, for example setting ASPNETCORE_TEST_LOG_DIR for the directory. Is there a reason why we want to pivot to msbuild based configuration? We'd need to flow msbuild properties via the assembly attribute but that's unorthodox and I prefer not resorting to that for everything.

JunTaoLuo:
I suppose we can do all configuration via msbuild properties and keep the ASPNETCORE_TEST_LOG_DIR for back compat only. Technically msbuild picks up environment variables anyways. Curious how we make this more discoverable since I doubt folks would know to look at the msbuild properties in this package to figure out how to turn on or configure the log directory.

JunTaoLuo:
Update

pakrym:
LGTM if works

PR #837, opened 2018-05-21
Timestamp and ITestOutputHelper fixes

JunTaoLuo:
- Standardize timestamp to use offsets in xunit output.
- Return the TestOutputHelper's output back to xunit so it doesn't go to /dev/null.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
@@ -57,7 +57,8 @@ namespace Microsoft.Extensions.Logging.Testing
 
         internal IDisposable StartTestLog(ITestOutputHelper output, string className, out ILoggerFactory loggerFactory, LogLevel minLogLevel, out string resolvedTestName, [CallerMemberName] string testName = null)
         {
-            var serviceProvider = CreateLoggerServices(output, className, minLogLevel, out resolvedTestName, testName);
+            var logStart = DateTimeOffset.UtcNow;
+            var serviceProvider = CreateLoggerServices(output, className, minLogLevel, out resolvedTestName, testName, logStart);
             var factory = serviceProvider.GetRequiredService<ILoggerFactory>();
             loggerFactory = factory;
             var logger = loggerFactory.CreateLogger("TestLifetime");
@@ -67,7 +68,7 @@ namespace Microsoft.Extensions.Logging.Testing
             var scope = logger.BeginScope("Test: {testName}", testName);
 
             _globalLogger.LogInformation("Starting test {testName}", testName);
-            logger.LogInformation("Starting test {testName}", testName);
+            logger.LogInformation("Starting test {testName} at {logStart}", testName, logStart.ToString("s"));
 
             return new Disposable(() =>
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Extensions.Logging.Testing/XunitLoggerFactoryExtensions.cs
```diff
@@ -22,6 +22,12 @@ public static ILoggingBuilder AddXunit(this ILoggingBuilder builder, ITestOutput
             return builder;
         }
 
+        public static ILoggingBuilder AddXunit(this ILoggingBuilder builder, ITestOutputHelper output, LogLevel minLevel, DateTimeOffset? utcNow)
```
kichalla: That's the advantage of using DateTimeOffset that you need not specify the contract with parameter name like `utcNow` here. Let the user give whatever value(either with timezone or utc) they want, you could always convert it to universal time when actually using it.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -291,24 +294,18 @@ private class AssemblyLogTimestampOffsetEnricher : ILogEventEnricher
...
-                }
+                _logStart = logStart;
             }
+
+            public void Enrich(LogEvent logEvent, ILogEventPropertyFactory propertyFactory)
+                => logEvent.AddPropertyIfAbsent(
+                    propertyFactory.CreateProperty(
+                        "TimestampOffset",
+                        _logStart.HasValue
+                            ? $"{(DateTimeOffset.UtcNow - _logStart.Value).TotalSeconds.ToString("N3")}s"
```
kichalla: Related to my earlier comment. You can always use `_logStart.Value.ToUniversalTime()` here.

JunTaoLuo: https://msdn.microsoft.com/en-us/library/bb354180(v=vs.110).aspx

> This method converts both DateTimeOffset objects to Coordinated Universal Time (UTC) before calculating the time interval that separates them. This removes any effect that different offsets from UTC may have on the comparison.

I don't think the `ToUniversalTime()` is needed.

src/Microsoft.Extensions.Logging.Testing/XunitLoggerFactoryExtensions.cs
JunTaoLuo: I guess this is poor naming since it doesn't actually need to be utcNow. I should just call it log start.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
kichalla: Interesting. Didn't knew that Subtract did that, but you still need to call `ToUniversalTime()` in other places. 
For example, `logger.LogInformation("Starting test {testName} at {logStart}", testName, logStart.ToString("s"));` (and one more place)...I would say just call `ToUniversalTime()` in all places. This way the code is readable and correct.

kichalla: Ignore my last comment. I see that you are using UtcNow few lines above the code.

PR #867, opened 2018-08-19
Remove redundant if condition from ConsoleLogger

martincostello:
Remove [redundant `if` condition](https://github.com/aspnet/Logging/blob/bdd33cdcfd6345c01f820ae0726c49b2c3bc4939/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L171) that checks for the `StringBuilder` being empty, when it always contains [at least 6 characters](https://github.com/aspnet/Logging/blob/bdd33cdcfd6345c01f820ae0726c49b2c3bc4939/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs#L143-L147) (`':'`, `' '`, `'['`, `'{digit}'`, `']'`, `'\n'`).

### DIFF
--- a/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
+++ b/src/Microsoft.Extensions.Logging.Console/ConsoleLogger.cs
@@ -168,19 +168,16 @@ namespace Microsoft.Extensions.Logging.Console
                 logBuilder.AppendLine(exception.ToString());
             }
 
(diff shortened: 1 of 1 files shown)


Eilon:
@muratg - can you assign? This change looks legit to me.

muratg:
@pakrym could you take a look and merge if it's all good?

Issue #877, opened 2018-09-06
EventSource filter spec should be applied to LoggerFilterOptions


pakrym:
So it can override default rules.

PR #883, opened 2018-09-25
Apply event source FilterSpec to filter options

pakrym:
Fixes: https://github.com/aspnet/Logging/issues/877

There is a breaking change in this PR: 
We stopped creating scopes for loggers that are completely disabled (LogLevel.Non). It's consistent with what event source logger did before and I think makes sense in general.

This PR also drops some net452 workarounds we had.

### DIFF
--- a/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
@@ -92,9 +92,8 @@ namespace Microsoft.Extensions.Logging.EventSource
                 return LogLevel.Trace;
             }
 
-            var rules = _rules;
             var level = LogLevel.None;
-            foreach (var rule in rules)
+            foreach (var rule in _rules)
             {
                 Debug.Assert(rule.LogLevel.HasValue);
                 Debug.Assert(rule.ProviderName == GetType().FullName);
--- a/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
@@ -25,6 +25,11 @@ namespace Microsoft.Extensions.Logging.EventSource
 
         private IDisposable _filterChangeToken;
 
+        public EventSourceLoggerProvider(LoggingEventSource eventSource) : this(eventSource, handleFilters: false)
+        {
+
+        }
+
         public EventSourceLoggerProvider(LoggingEventSource eventSource, bool handleFilters)
         {
             if (eventSource == null)
@@ -59,8 +64,13 @@ namespace Microsoft.Extensions.Logging.EventSource
 
         public void Dispose()
         {
-            _filterChangeToken.Dispose();
-            SetFilterSpec(null); // Turn off any logging
+            _filterChangeToken?.Dispose();
+
+            // Turn off any logging
+            for (var logger = _loggers; logger != null; logger = logger.Next)
+            {
+                logger.Level = LogLevel.None;
+            }
         }
 
         // Sets the filtering for a particular logger provider
--- a/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
@@ -2,7 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
-using System.Diagnostics.Tracing;
+using System.Diagnostics;
+using System.Threading;
 
 namespace Microsoft.Extensions.Logging.EventSource
 {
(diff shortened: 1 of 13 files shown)


src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -12,226 +13,94 @@ namespace Microsoft.Extensions.Logging.EventSource
...
-            _eventSource.ApplyFilterSpec();
             var newLogger = _loggers = new EventSourceLogger(categoryName, _factoryID, _eventSource, _loggers);
-            newLogger.Level = ParseLevelSpecs(_filterSpec, _defaultLevel, newLogger.CategoryName);
+            newLogger.Level = GetLoggerLevel(newLogger.CategoryName);
             return newLogger;
         }
 
         public void Dispose()
         {
+            _filterChangeToken.Dispose();
```
BrennanConroy: `?.` as this is null in some cases

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -12,226 +13,94 @@ namespace Microsoft.Extensions.Logging.EventSource
...
-        /// The first speciciation that 'loggers' Name matches is used.
-        /// </summary>
-        private LogLevel ParseLevelSpecs(string filterSpec, LogLevel defaultLevel, string loggerName)
-        {
-            if (filterSpec == null)
-            {
-                return LoggingEventSource.LoggingDisabled;      // Null means disable.
-            }
-            if (filterSpec == string.Empty)
+            var rules = _rules;
```
BrennanConroy: Null check, sometimes `SetFilterSpec` is given null

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -12,226 +13,94 @@ namespace Microsoft.Extensions.Logging.EventSource
...
-                    {
-                        continue;
-                    }
-
-                    // We allow wildcards at the end.
-                    if (specChar == '*' && ParseLevel(defaultLevel, filterSpec, specPos, ref level))
-                    {
-                        return level;
-                    }
+                    level = rule.LogLevel.Value;
```
BrennanConroy: Should this be checking if the `rule.LogLevel` is more restrictive than the current value? Otherwise it will just apply the last rule only.

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
pakrym: it's handled in ParseFilterSpec

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
pakrym: It would take first matching rule the same way EventLogger worked before the change.

pakrym: Only rules coming from `ParseFilterSpec` are passed in this method and only one of them wouldn't have category.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -208,29 +191,168 @@ protected override void OnEventCommand(EventCommandEventArgs command)
...
 
-            // In .NET 4.5.2 the internal EventSource level hasn't been correctly set
-            // when this callback is invoked. To still have the logger behave correctly
-            // in .NET 4.5.2 we delay checking the level until the logger is used the first
-            // time after this callback.
-            _checkLevel = true;
+        [NonEvent]
+        private void FireChangeToken()
+        {
+            var tcs = Interlocked.Exchange(ref _cancellationTokenSource, null);
```
anurse: nit: `cts`, it's not a TCS :)

src/Microsoft.Extensions.Logging.EventSource/EventSourceLoggerProvider.cs
```diff
@@ -12,226 +13,104 @@ namespace Microsoft.Extensions.Logging.EventSource
...
+                return LogLevel.Trace;
             }
 
-            var level = LoggingEventSource.LoggingDisabled;   // If the logger does not match something, it is off.
-
-            // See if logger.Name  matches a _filterSpec pattern.
-            var namePos = 0;
-            var specPos = 0;
-            for (;;)
+            var rules = _rules;
```
anurse: nit: If copying the reference here is intentional, comment on why.

src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
```diff
@@ -208,29 +191,168 @@ protected override void OnEventCommand(EventCommandEventArgs command)
...
+        }
+
+        [NonEvent]
+        private LogLevel GetDefaultLevel()
+        {
+            var allMessageKeywords = Keywords.Message | Keywords.FormattedMessage | Keywords.JsonMessage;
+
+            if (IsEnabled(EventLevel.Informational, allMessageKeywords))
+            {
+                if (IsEnabled(EventLevel.Verbose, allMessageKeywords))
```
anurse: Is this funky branching just to avoid checking Verbose unless we know we're at least enabled for Informational?

pakrym: Copied from old implementation. I'll flattened it all the way and add mapping from LogAlways to Trace.

PR #884, opened 2018-09-28
Don't double quote AssemblyTestLog parameters

halter73:
See https://github.com/serilog/serilog/issues/956

### DIFF
--- a/build/dependencies.props
+++ b/build/dependencies.props
@@ -26,7 +26,7 @@
     <NETStandardLibrary20PackageVersion>2.0.3</NETStandardLibrary20PackageVersion>
     <NewtonsoftJsonPackageVersion>11.0.2</NewtonsoftJsonPackageVersion>
     <SerilogExtensionsLoggingPackageVersion>1.4.0</SerilogExtensionsLoggingPackageVersion>
-    <SerilogSinksFilePackageVersion>3.2.0</SerilogSinksFilePackageVersion>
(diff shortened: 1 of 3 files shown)


JunTaoLuo:
Is this what broke some of our tests?

halter73:
I doubt this is what broke our tests. The double quoting was just annoying me.

BrennanConroy:
Besides notifying Eilon, I think you need to update universe too. https://github.com/aspnet/Universe/blob/release/2.2/build/dependencies.props#L119

Eilon:
Sorry I'm not sure I understood - what's the ultimate motivation of updating the dependency? Is it due to test failures? Or just to help users get the latest Serilog? To fix a bug?

JunTaoLuo:
I think this is just for aesthetics and the double quotes were annoying @halter73.

Eilon:
That's a pretty weak-sauce argument 😄 Are there any other reasons to update? We do like to keep things fresh, but this is also a major version update of a dependency, so are there any Serilog breaking changes that we would be forcing users to deal with?

JunTaoLuo:
I didn't know we needed a version update when I originally reviewed the PR but as mentioned above, this package is only used in noship packages and test projects so I think updating the version is okay as long as all of our tests continue pass.

Eilon:
Oh it's not in anything we ship? I'll gladly approve that then, if someone can confirm that.

PR #886, opened 2018-10-01
Add retry capability to tests

JunTaoLuo:
I'm piggy-backing off the infrastructure we have in place for LoggedTest since that is much simpler than creating its own set of discoverers, runners and invokers. It would also make it hard to handle a test class that's both logged and retried. The trade-off is that all retried tests must also be logged and I think that's reasonable since most of the tests would be functional tests that should be logged anyway.

Backgound:

We have some functional tests in Kestrel that's flaky on macOS due to underlying issues in the kernel. We don't want to disable all affected tests since that would dramatically reduce our coverage. In these cases, we would want to retry tests. This should only be allowed on tests with well known diagnosis and when all other possible fixes has been implemented.

This will help us resolve flaky test issues such as https://github.com/aspnet/KestrelHttpServer/issues/2907.

### DIFF
--- a/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Extensions.Logging.Testing
     public class LoggedTestInvoker : XunitTestInvoker
     {
         private readonly ITestOutputHelper _output;
+        private readonly RetryContext _retryContext;
 
         public LoggedTestInvoker(
             ITest test,
@@ -25,22 +26,39 @@ namespace Microsoft.Extensions.Logging.Testing
             IReadOnlyList<BeforeAfterTestAttribute> beforeAfterAttributes,
             ExceptionAggregator aggregator,
             CancellationTokenSource cancellationTokenSource,
-            ITestOutputHelper output)
+            ITestOutputHelper output,
+            RetryContext retryContext)
             : base(test, messageBus, testClass, constructorArguments, testMethod, testMethodArguments, beforeAfterAttributes, aggregator, cancellationTokenSource)
         {
             _output = output;
+            _retryContext = retryContext;
         }
 
         protected override object CreateTestClass()
         {
             var testClass = base.CreateTestClass();
 
-            if (testClass is ILoggedTest loggedTest)
+            (testClass as ILoggedTest).Initialize(
+                TestMethod,
+                TestMethodArguments,
+                _output ?? ConstructorArguments.SingleOrDefault(a => typeof(ITestOutputHelper).IsAssignableFrom(a.GetType())) as ITestOutputHelper);
+
+            if (testClass is LoggedTestBase loggedTestBase)
             {
-                loggedTest.Initialize(
-                    TestMethod,
-                    TestMethodArguments,
-                    _output ?? ConstructorArguments.SingleOrDefault(a => typeof(ITestOutputHelper).IsAssignableFrom(a.GetType())) as ITestOutputHelper);
+                // Used for testing
+                loggedTestBase.RetryContext = _retryContext;
+
+                if (_retryContext != null)
+                {
+                    // Log retry attempt as warning
+                    if (_retryContext.CurrentIteration > 0)
+                    {
+                        loggedTestBase.Logger.LogWarning($"{TestMethod.Name} failed and retry conditions are met, re-executing. The reason for failure is {_retryContext.Reason}.");
+                    }
+
+                    // Save the test class instance for non-static predicates
+                    _retryContext.TestClassInstance = testClass;
+                }
             }
 
             return testClass;
--- a/src/Microsoft.Extensions.Logging.Testing/Xunit/RetryTestAttribute.cs
+++ b/src/Microsoft.Extensions.Logging.Testing/Xunit/RetryTestAttribute.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel;
+using Microsoft.AspNetCore.Testing.xunit;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    /// <summary>
+    /// WARNING: This attribute should only be used on well understood flaky test failures caused by external issues and should be removed once the underlying issues have been resolved.
+    /// This is not intended to be a long term solution to ensure passing of flaky tests but instead a method to improve test reliability without reducing coverage.
+    /// Issues should be filed to remove these attributes from affected tests as soon as the underlying issue is fixed.
+    /// </summary>
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false)]
+    public class RetryTestAttribute : Attribute
+    {
+        /// <summary>
+        /// WARNING: This attribute should only be used on well understood flaky test failures caused by external issues and should be removed once the underlying issues have been resolved.
+        /// This is not intended to be a long term solution to ensure passing of flaky tests but instead a method to improve test reliability without reducing coverage.
+        /// Issues should be filed to remove these attributes from affected tests as soon as the underlying issue is fixed.
+        /// </summary>
+        /// <param name="retryPredicateName">The predicate of the format Func&lt;Exception,bool&gt; that is used to determine if the test should be retried</param>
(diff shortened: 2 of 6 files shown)


test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitRetryTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        public LoggedTestXunitRetryTests()
+        {
+            RetryCounter.Reset();
+        }
+
+        [Fact]
+        public void CompletesWithoutRetryOnSuccess()
+        {
+            Assert.Equal(2, TestRetries);
```
Tratcher: RetryLimit

Tratcher:
The offending piece of code is often a socket that's been closed and a fair amount of the test needs to be restarted. @JunTaoLuo any examples on what it would take to re-write a test for partial retries?

Eilon:
Would the type of failure we're seeing in the tests also be a type of failure someone could see at runtime in an app?

JunTaoLuo:
While we can see these types of failures at runtime in apps, they are restricted to macOS which is mostly a dev scenario. Nothing in Logging need retries in particular, but as I mentioned, to duplicate the infrastructure to separate out LoggedTest and RetriedTest would be significant and I think it would provide very little benefit since in almost all scenarios, we will only retry functional tests which should already be logged. Retrying parts of tests could be different on each individual basis but I can see what that would look like. Also, these types of failures tend to manifest in many different types of errors so we'll end up retrying many parts of the tests.

Eilon:
If we log retries, who would ever look at those logs? Presumably it won't fail the build, so it would just get ignored indefinitely anyway, no?

But, back to Mac being dev-only: That's true, but could this issue be irritating to someone using a Mac for dev?

JunTaoLuo:
I wonder if I can add a mechanism for only retrying on specific types of exceptions. For the mac scenario, it would be interesting to see how the current inconsistencies are being dealt with. I'm not sure if adding retries for every socket functionality we use is a feasible alternative.

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitRetryTests.cs
```diff
@@ -51,6 +52,22 @@ public void RetryCountOverridenWhenOSMatches()
         {
             Assert.Equal(3, TestRetries);
         }
+
+        [ConditionalFact]
+        [RetryTest(2, typeof(TestRetryPredicate))]
```
Tratcher: We shouldn't have to declare a full type just for a callback. MemberData uses a string pointing at a property that we set with nameof. We can do the same here with the string name of a local method.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -51,6 +51,11 @@ protected override async Task<decimal> InvokeTestMethodAsync(object testClassIns
             {
                 if (i > 0)
                 {
+                    if (!loggedTestBase.RetryPredicate(Aggregator.ToException()))
```
Tratcher: loggedTestBase does not guarantee RetryPredicate is set

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitRetryTests.cs
JunTaoLuo: Hmm not sure if that would work as an assembly attribute. I'll check.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
JunTaoLuo: Actually it does. The RetryTestAttribute guarantees that the predicated is set. Then if a RetryTestAttribute can be resolved, the loggedTestBase sets both TestRetries and RetryPredicate. Since this code path can only be triggered if TestRetries is greater than 1, RetryPredicate is guaranteed to be set.

src/Microsoft.Extensions.Logging.Testing/Xunit/LoggedTestInvoker.cs
```diff
@@ -35,15 +36,40 @@ protected override object CreateTestClass()
...
+                    {
+                        return Timer.Total;
+                    }
+
+                    // This can only occur if the retry count has been overridden on the LoggedTestBase class
+                    loggedTestBase.Logger.LogWarning($"{TestMethod.Name} failed and retries are enabled, re-executing.");
+                }
+
+                Aggregator.Clear();
+                await base.InvokeTestMethodAsync(testClassInstance);
```
Tratcher: At what level does this retry happen? It run-runs the method right? That might not actually work for some of the Kestrel tests that depend on state set up in the constructors. Have you been able to test it in Kestrel?

JunTaoLuo: Good catch I'll rerun the setup logic as well.

src/Microsoft.Extensions.Logging.Testing/Xunit/RetryTestAttribute.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.AspNetCore.Testing.xunit;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    [AttributeUsage(AttributeTargets.Method | AttributeTargets.Class, AllowMultiple = false)]
+    public class RetryTestAttribute : Attribute
+    {
+        /// <summary>
+        /// Warning this attribute should only be used on well understood flaky test failures caused by external issues and should be removed once the underlying issues have been resolved.
```
Eilon: `WARNING: ...`

JunTaoLuo:
FYI the usage in Kestrel would look something like https://github.com/aspnet/KestrelHttpServer/commit/162273d90668dcd8419dc394c9eafe296523a13f albeit with more detailed message and predicate probably.

src/Microsoft.Extensions.Logging.Testing/Xunit/RetryTestAttribute.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.ComponentModel;
+using Microsoft.AspNetCore.Testing.xunit;
+
+namespace Microsoft.Extensions.Logging.Testing
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
```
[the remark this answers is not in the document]
Eilon: Perhaps not... we're trying to have it be somewhat hidden...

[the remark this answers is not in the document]
Tratcher: At that point we're just trolling ourselves.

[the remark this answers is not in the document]
Eilon: I do think that's the point, though. But I don't feel strongly on this.

halter73:
Could we allow predicates to take a LoggedTest argument? I'm thinking we might want a predicate that could inspect the TestApplicationErrorLogger of a test using the TestApplicationErrorLoggerLoggedTest base class to determine whether to retry. This would require casting in the predicate of course.

test/Microsoft.Extensions.Logging.Testing.Tests/LoggedTestXunitRetryTests.cs
```diff
@@ -0,0 +1,101 @@
...
+            Assert.Equal($"{nameof(RetryIfPredicateIsTrue)} failed and retry conditions are met, re-executing. The reason for failure is sample reason.", loggedMessage.Message);
+        }
+
+        // Static predicates are valid
+        private static bool RetryAllPredicate(Exception e)
+            => true;
+
+        // Instance predicates are valid
+        private bool RetryInvalidOperationExceptionPredicate(Exception e)
+            => TestSink.Writes.Any(m => m.Message.Contains("Throw on first iteration"));
```
halter73: Never mind about the logged test parameter. I didn't see that you already supported instance methods as predicates. This is exactly what I want.

src/Microsoft.Extensions.Logging.Testing/Xunit/RetryTestAttribute.cs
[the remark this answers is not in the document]
JunTaoLuo: I figured we don't really want to advertise this functionality and hopefully the strangeness will motivate the reading of the xml docs.

PR #888, opened 2018-10-02
Fixes the description for the property FileName

### LINKED ISSUE #863: Wrong property comment/description in AzureFileLoggerOptions
Property `FileName` has the same description/comment as property `RetainedFileCountLimit`:

https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs

Gimly:
The property FileName was the same as the one for `RetainedFileCountLimit`, this commits changes to something that describes this property more correctly.

Fixes #863

### DIFF
--- a/src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
+++ b/src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
@@ -51,8 +51,8 @@ namespace Microsoft.Extensions.Logging.AzureAppServices
         }
 
         /// <summary>
-        /// Gets or sets a string representing the name of the file used to store the logging information.
-        /// The current date, in the format YYYYMMDD will be added after the given filename.
+        /// Gets or sets a string representing the prefix of the file name used to store the logging information.
+        /// The current date, in the format YYYYMMDD will be added after the given value.
         /// Defaults to <c>diagnostics-</c>.
         /// </summary>
         public string FileName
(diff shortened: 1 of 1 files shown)


src/Microsoft.Extensions.Logging.AzureAppServices/AzureFileLoggerOptions.cs
```diff
@@ -51,8 +51,8 @@ public class AzureFileLoggerOptions: BatchingLoggerOptions
         }
 
         /// <summary>
-        /// Gets or sets a strictly positive value representing the maximum retained file count or null for no limit.
-        /// Defaults to <c>2</c>.
+        /// Gets or sets a string representing the name of the file used to store the logging information.
```
Eilon: Isn't this actually a filename prefix or something like that? I took a quick look at the code and in at least some circumstances that seems to be the case. But I've never used this provider so I'm not too familiar.

But, your comment is certainly better than what was there before!

Gimly: Well, it's the filename, but it's automatically suffixed by the date, as can be seen in this code snippet (from FileLoggerProvider.cs):

```C#
private string GetFullName((int Year, int Month, int Day) group)
{
    return Path.Combine(_path, $"{_fileName}{group.Year:0000}{group.Month:00}{group.Day:00}.txt");
}
```

Maybe I can add this information in the description.

pakrym: `Gets or sets a string representing the prefix of the file the file name used to store the logging...` ?

Gimly:
Updated the comment to make it clear that it's a prefix

pakrym:
Thank you for your contribution!

PR #895, opened 2018-10-09
Added override of ToString for Logger.Scope

### LINKED ISSUE #893: Logger.BeginScope ToString fails when LoggerProvider without ISupportExternalScope
Registering a single LoggerProvider that doesn't implement `ISupportExternalScope` will now fail in the following code:

```
using (var scopeState = _logger.BeginScope("{World} People", "Earth"))
{
    _logger.LogInformation(scopeState.ToString());  // No longer prints Earth People
}
```

Bug introduced by #723, because the following optimization was removed:

```
if (loggers.Length == 1)
    return loggers[0].Logger.BeginScope(state);
```

### LINKED ISSUE #894: Logger.BeginScope with IncludeScopes = false gets performance hit
Normally one can have very fast logging when registering a single LoggerProvider with IncludeScopes = false, where BeginScope returns a NullScope.

Now because of #723 then extra allocation occurs because the following optimization was removed:

```
if (loggers.Length == 1)
    return loggers[0].Logger.BeginScope(state);
```

If the LoggerProvider attempts to implement ISupportExternalScope (And has IncludeScopes = false), then the LoggerFactory will generate even more garbage-allocation because it forces creation of AsyncLocal-state that must be transferred when using async-threading-tasks.

Again if the above optimization was used, then one could skip the external-scope-provider-logic when only a single LoggerProvider (IncludeScopes = false would continue to work).

See also #714

snakefoot:
Resolves #893 and resolves #894 

Updated to react to CreateScopes introduced with #883

### DIFF
--- a/src/Microsoft.Extensions.Logging/Logger.cs
+++ b/src/Microsoft.Extensions.Logging/Logger.cs
@@ -145,7 +145,10 @@ namespace Microsoft.Extensions.Logging
                 {
                     scopeCount++;
                 }
-
+            }
+            else if (loggers.Length == 1 && loggers[0].CreateScopes)
+            {
+                return loggers[0].Logger.BeginScope(state);
             }
 
             var scope = new Scope(scopeCount);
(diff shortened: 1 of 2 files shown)


snakefoot:
Will become obsolete when #929 is merged.

pakrym:
Fixed by https://github.com/aspnet/Logging/pull/929 in a bit more generic way.

Issue #899, opened 2018-10-11
Inability to filter EventSource categories defined by LoggingEventSource.Keywords


rymeskar:
LoggingEventSource should be public because one needs to use the 'inside-class' defined Keywords for EventListeners. This pattern was even visible in tests before. The visibility was bent, by marking the test as a friend assembly.

LoggingEventSource should be also sealed as per this [documentation](https://github.com/Microsoft/dotnet-samples/blob/master/Microsoft.Diagnostics.Tracing/EventSource/docs/EventSource.md).  The reasoning behind is to keep the class hierarchy as trivial as possible.

davidfowl:
> LoggingEventSource should be public because one needs to use the 'inside-class' defined Keywords for EventListeners. This pattern was even visible in tests before. 

Can you show an example. This isn't your typical event source, it's an implementation detail.

rymeskar:
Yes, a good example would be if one wants to only allow the emission of JsonMessage.
This would be achieved through calling `EnableEvents(eventSource, EventLevel.LogAlways, LoggingEventSource.Keywords.JsonMessage);` in the overriden OnEventSourceCreated method of a derived EventListener.

What other way is there to filter the 'categories' so that it does not spam your system unnecessarily? I could not find any without the access to the Keywords.

rymeskar:
@davidfowl 
I have thought about the issue a bit more. And you are right, in this case, the LoggingEventSource should be an implementation detail.

I should have rather named the issue: "Inability to filter EventSource categories defined by LoggingEventSource.Keywords"

Then the initially described issue here is just one of the possible solutions to filtering the categories.

The other (and quite frankly cleaner solution without leaking implementaiton details) is in high-level terms:
to provide this 'category' filter on initialization. Then we do not have to leak the information about how the ETW events are emitted. But the fact that the category definitions (Keywords) have to be accessible to the public still holds.

davidfowl:
I fixed it for you, lets expose those keywords.

cc @pakrym

rymeskar:
Do you mean to:

1. Expose only the Keywords, or
2. Also have the possibility to filter the categories right inside the Extensions.Logging.EventSource package? Without needing to leak the info about how the actual logging events are emitted.

pakrym:
I feel like the entire goal of EventSource logger is to be as detailed as possible by default and then collector decides which events to filter.

So I would expose only the keywords.

rymeskar:
I also like to see detailed detailed events.

[...]

PR #900, opened 2018-10-11
Making LoggingEventSource public sealed

rymeskar:
Fixing [#899](https://github.com/aspnet/Logging/issues/899)

### DIFF
--- a/src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
+++ b/src/Microsoft.Extensions.Logging.EventSource/LoggingEventSource.cs
@@ -222,7 +222,7 @@ namespace Microsoft.Extensions.Logging.EventSource
         /// The first specification that 'loggers' Name matches is used.
         /// </summary>
         [NonEvent]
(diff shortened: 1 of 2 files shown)


rymeskar:
The check failures do not seem to be related to my changes. Unfortunately, I am not able to rerun them without a new commit :(

pakrym:
Not sure we want to expose Keywords in this manner. I was thinking about moving the enum out of the class and exposing only it.

rymeskar:
When I read [this blog post ](https://blogs.msdn.microsoft.com/vancem/2012/08/14/etw-in-c-controlling-which-events-get-logged-in-an-system-diagnostics-tracing-eventsource/) from 2012.
It mentioned that the Keywords must be nested.

But TBH, I am not sure what is the enforcement mechanism or whether that requirement still holds.

pakrym:
Good point.

Make sure every other member is private/internal then.

pakrym:
Thank you for your contribution.

PR #903, opened 2018-10-12
Fix EventSource logger for cases where listener exists before app starts

pakrym:
Found some issues with the logger when doing E2E testing with perfview.

### DIFF
--- a/test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
+++ b/test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
@@ -114,6 +114,37 @@ namespace Microsoft.Extensions.Logging.Test
         }
 
         [Fact]
+        public void Logs_AsExpected_WithDefaults_EnabledEarly()
+        {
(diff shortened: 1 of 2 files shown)


test/Microsoft.Extensions.Logging.EventSource.Test/EventSourceLoggerTest.cs
```diff
@@ -511,6 +556,12 @@ protected override void OnEventSourceCreated(System.Diagnostics.Tracing.EventSou
                 {
                     _loggingEventSource = eventSource;
                 }
+
+                if (_enableWhenCreated != null)
```
anurse: Is this guaranteed to be called sequentially? What if the event source is being created while `EnableEvents` is being called?

pakrym: It's testing infrastructure, we only call them sequentially.

anurse: Ah, I see, this is the test listener.

PR #904, opened 2018-10-19
Lower test logger max path limit to allow wiggle room for build syste…

BrennanConroy:
…ms copying artifacts

@muratg We want this for 2.1+


pakrym:
Let's try it.

src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs
```diff
@@ -21,7 +21,7 @@ public class AssemblyTestLog : IDisposable
     {
         public static readonly string OutputDirectoryEnvironmentVariableName = "ASPNETCORE_TEST_LOG_DIR";
         private static readonly string LogFileExtension = ".log";
-        private static readonly int MaxPathLength = 245;
```
JunTaoLuo: Hmmm maybe make this configurable via an env var.

BrennanConroy: That could be good

BrennanConroy:
Added an environment variable

Issue #860, opened 2018-07-16
EventLogLogger Doesn't Dispose of System.Diagnostics.EventLog


jeff-bacon:
Microsoft.Extensions.Logging.EventLog.WindowsEventLog.cs creates a System.Diagnostics.EventLog which is IDisposable, but WindowsEventLog itself is not IDisposable.  This could be a potential memory leak.

PR #910, opened 2018-10-24
Implementing IDisposable on WindowsEventLog #860

arnaudleclerc:
Implements the IDisposable interface on WindowsEventLog to dispose the instance of EventLog created on the constructor. Cf. #860.


pakrym:
This doesn't fix the issue. Noone is disposing the WindowsEventLog instance.

arnaudleclerc:
Fair enough. Another possibility could be to create an instance of EventLog on demand when WriteEntry is called and dispose it. There's still an issue with the DiagnosticEventLog property though, which, even if it is also created when the property is requested and not already disposed, will never be disposed for the same reasons I guess.

pakrym:
This repository is going to get merged into https://github.com/aspnet/Extensions/.
This requires closing all incomplete PRs.

PR #913, opened 2018-10-25
Support stderr in console logger

pakrym:
Fixes: https://github.com/aspnet/Logging/issues/661

### DIFF
--- a/samples/SampleApp/logging.json
+++ b/samples/SampleApp/logging.json
@@ -8,7 +8,8 @@
     "Console":
     {
       "IncludeScopes": "true",
-      "TimestampFormat": "[HH:mm:ss] "
+      "TimestampFormat": "[HH:mm:ss] ",
+      "LogAsErrorLevel": "Warning"
     }
   }
 }
--- a/samples/SampleApp/logging.json
+++ b/samples/SampleApp/logging.json
@@ -9,7 +9,7 @@
     {
       "IncludeScopes": "true",
       "TimestampFormat": "[HH:mm:ss] ",
-      "LogAsErrorLevel": "Warning"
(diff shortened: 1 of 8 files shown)


samples/SampleApp/logging.json
```diff
@@ -8,7 +8,8 @@
     "Console":
     {
       "IncludeScopes": "true",
-      "TimestampFormat": "[HH:mm:ss] "
+      "TimestampFormat": "[HH:mm:ss] ",
+      "LogAsErrorLevel": "Warning"
```
BrennanConroy: Not a fan of the name.

"LogLevelToStdErr"?
"LogToStdErr" - and just do error and critical, don't let user choose a level?
"LogAsStdErrLevel"?

pakrym: Don't like the name either but like the flexibility.

muratg: @shirhatti @glennc for PM opinions on naming.

anurse: I'm not sure a Level configuration is really needed. The only thing I could think users might want to do is control what happens to `Warning`. Most logging systems I've used either a) write everything to stdout or b) write `Warning` and above to stderr and everything else to stdout.

I'm OK with the level though, if we want the flexibility. The name probably can't really be super self-documenting without getting super long, so we should add this to docs. At a minimum, let's get an issue filed to track this. I think I like `LogToStandardError` or something similar as a name (we generally avoid abbreviations like `StdErr` in APIs and I think config options are similar here). As long as it's documented, I'm not super invested though :D

gfoidl:
@pakrym thanks for implementing this 👍

PR #917, opened 2018-10-26
Decrease logging overhead

pakrym:
1. Make a lot of things struct to avoid boxing
2. Add specialized overloads for 1-3 parameter messages that match `string.Format` and avoid allocation
3. Fix some special casing for no argument logs

### Before
```

                                     Method |      Mean |     Error |    StdDev |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
------------------------------------------- |----------:|----------:|----------:|-------------:|-------:|---------:|-------:|----------:|
                NoArguments_FilteredByLevel |  54.30 ns | 0.4572 ns | 0.4053 ns | 18,415,469.9 |   2.72 |     0.02 | 0.0001 |      40 B |
  NoArguments_DefineMessage_FilteredByLevel |  19.94 ns | 0.1230 ns | 0.1027 ns | 50,138,751.2 |   1.00 |     0.00 |      - |       0 B |
                                NoArguments |  74.18 ns | 1.2185 ns | 1.1398 ns | 13,480,512.4 |   3.72 |     0.06 | 0.0001 |      40 B |
                  NoArguments_DefineMessage |  78.33 ns | 1.0545 ns | 0.9348 ns | 12,766,084.6 |   3.93 |     0.05 | 0.0001 |      24 B |
                               TwoArguments | 146.06 ns | 1.7506 ns | 1.6375 ns |  6,846,641.8 |   7.32 |     0.09 | 0.0005 |     104 B |
               TwoArguments_FilteredByLevel | 128.65 ns | 2.5191 ns | 2.8000 ns |  7,772,791.9 |   6.45 |     0.14 | 0.0005 |     104 B |
                 TwoArguments_DefineMessage |  89.52 ns | 0.9914 ns | 0.9274 ns | 11,170,548.1 |   4.49 |     0.05 | 0.0001 |      40 B |
 TwoArguments_DefineMessage_FilteredByLevel |  20.63 ns | 0.2005 ns | 0.1875 ns | 48,463,314.9 |   1.03 |     0.01 |      - |       0 B |


 
       Method |     Mean |    Error |   StdDev |        Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
------------- |---------:|---------:|---------:|------------:|-------:|---------:|-------:|----------:|
 TwoArguments | 355.1 ns | 6.855 ns | 8.161 ns | 2,815,805.2 |   1.65 |     0.04 | 0.0005 |     168 B |
  NoArguments | 215.3 ns | 3.141 ns | 2.785 ns | 4,645,103.7 |   1.00 |     0.00 |      - |      64 B |

```

### After
```
                                     Method |      Mean |     Error |    StdDev |         Op/s | Scaled | ScaledSD |  Gen 0 | Allocated |
------------------------------------------- |----------:|----------:|----------:|-------------:|-------:|---------:|-------:|----------:|
                NoArguments_FilteredByLevel |  49.29 ns | 0.5002 ns | 0.4679 ns | 20,286,488.9 |   2.60 |     0.04 |      - |       0 B |
  NoArguments_DefineMessage_FilteredByLevel |  18.96 ns | 0.2912 ns | 0.2724 ns | 52,732,789.4 |   1.00 |     0.00 |      - |       0 B |
                                NoArguments |  64.03 ns | 0.2904 ns | 0.2717 ns | 15,616,744.5 |   3.38 |     0.05 |      - |       0 B |
                  NoArguments_DefineMessage |  53.39 ns | 0.3256 ns | 0.3046 ns | 18,731,554.7 |   2.82 |     0.04 |      - |       0 B |
                               TwoArguments | 134.16 ns | 0.6308 ns | 0.5268 ns |  7,453,628.1 |   7.08 |     0.10 | 0.0002 |      64 B |
               TwoArguments_FilteredByLevel | 124.44 ns | 1.5384 ns | 1.3637 ns |  8,035,928.9 |   6.56 |     0.11 | 0.0002 |      64 B |
                 TwoArguments_DefineMessage |  96.02 ns | 0.6235 ns | 0.5206 ns | 10,414,278.9 |   5.06 |     0.07 |      - |       0 B |
 TwoArguments_DefineMessage_FilteredByLevel |  19.26 ns | 0.1237 ns | 0.1097 ns | 51,914,239.5 |   1.02 |     0.02 |      - |       0 B |

       Method |      Mean |     Error |    StdDev |         Op/s | Scaled | ScaledSD | Allocated |
------------- |----------:|----------:|----------:|-------------:|-------:|---------:|----------:|
 TwoArguments | 330.32 ns | 4.2330 ns | 3.9595 ns |  3,027,379.5 |   5.94 |     0.10 |      88 B |
  NoArguments |  55.58 ns | 0.8043 ns | 0.7523 ns | 17,992,955.1 |   1.00 |     0.00 |       0 B |
```

This doesn't address scope impact but adds benchmarks to be able to iterate on perf further.

### DIFF
--- a/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
+++ b/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
 using BenchmarkDotNet.Attributes;
 using Microsoft.Extensions.DependencyInjection;
 
--- a/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
+++ b/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using BenchmarkDotNet.Attributes;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Extensions.Logging.Performance
+{
+    [AspNetCoreBenchmark]
+    public class FormattingBenchmarks : LoggingBenchmarkBase
+    {
+        private ILogger _logger;
+
+        [Benchmark]
+        public void TwoArguments()
+        {
+            TwoArgumentErrorMessage(_logger, 1, "string", Exception);
+        }
+
+        //[Benchmark(Baseline = true)]
+        //public void NoArguments()
+        //{
+        //    NoArgumentErrorMessage(_logger, Exception);
+        //}
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var services = new ServiceCollection();
+            services.AddLogging();
+            services.AddSingleton<ILoggerProvider, LoggerProvider<MessageFormattingLogger>>();
+
+            _logger = services.BuildServiceProvider().GetService<ILoggerFactory>().CreateLogger("Logger");
+        }
+
+        public class MessageFormattingLogger: ILogger
+        {
+            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+            {
+                formatter(state, exception);
+            }
+
+            public bool IsEnabled(LogLevel logLevel)
+            {
+                return true;
+            }
+
+            public IDisposable BeginScope<TState>(TState state)
+            {
+                return null;
+            }
+        }
+    }
+}
--- a/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
+++ b/benchmarks/Logging.Performance/LogValuesBenchmarks.cs
@@ -15,11 +15,11 @@ namespace Microsoft.Extensions.Logging.Performance
             TwoArgumentErrorMessage(_logger, 1, "string", Exception);
         }
 
-        //[Benchmark(Baseline = true)]
-        //public void NoArguments()
-        //{
-        //    NoArgumentErrorMessage(_logger, Exception);
-        //}
+        [Benchmark(Baseline = true)]
+        public void NoArguments()
+        {
+            NoArgumentErrorMessage(_logger, Exception);
+        }
 
         [GlobalSetup]
         public void Setup()
--- a/benchmarks/Logging.Performance/Logging.Performance.csproj
+++ b/benchmarks/Logging.Performance/Logging.Performance.csproj
@@ -1,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <RootNamespace>Microsoft.Extensions.DependencyInjection.Performance</RootNamespace>
+    <OutputType>Exe</OutputType>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Extensions.Logging\Microsoft.Extensions.Logging.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="$(BenchmarkDotNetPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.BenchmarkRunner.Sources" Version="$(MicrosoftAspNetCoreBenchmarkRunnerSourcesPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(MicrosoftExtensionsDependencyInjectionPackageVersion)" />
+  </ItemGroup>
+
+</Project>
--- a/benchmarks/Logging.Performance/Logging.Performance.csproj
+++ b/benchmarks/Logging.Performance/Logging.Performance.csproj
@@ -1,10 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RootNamespace>Microsoft.Extensions.DependencyInjection.Performance</RootNamespace>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <RootNamespace>Microsoft.Extensions.Logging.Performance</RootNamespace>
     <OutputType>Exe</OutputType>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
--- a/benchmarks/Logging.Performance/LoggingBenchmarkBase.cs
+++ b/benchmarks/Logging.Performance/LoggingBenchmarkBase.cs
@@ -10,8 +10,9 @@ namespace Microsoft.Extensions.Logging.Performance
         protected static readonly Action<ILogger, Exception> NoArgumentTraceMessage = LoggerMessage.Define(LogLevel.Trace, 0, "Message");
         protected static readonly Action<ILogger, Exception> NoArgumentErrorMessage = LoggerMessage.Define(LogLevel.Error, 0, "Message");
 
-        protected static readonly Action<ILogger,  int, string, Exception> TwoArgumentTraceMessage = LoggerMessage.Define<int, string>(LogLevel.Trace, 0, "Message {Argument1} {Argument2}");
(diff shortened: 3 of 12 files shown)


src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -245,11 +245,9 @@ private static LogValuesFormatter CreateLogValuesFormatter(string formatString,
             return logValuesFormatter;
         }
 
-        private class LogValues : IReadOnlyList<KeyValuePair<string, object>>
+        private readonly struct LogValues : IReadOnlyList<KeyValuePair<string, object>>
         {
-            public static Func<object, Exception, string> Callback = (state, exception) => ((LogValues)state)._formatter.Format(((LogValues)state).ToArray());
-
-            private static object[] _valueArray = new object[0];
+            public static readonly Func<LogValues, Exception, string> Callback = (state, exception) => state.ToString();
```
benaadams: Use delegate and `in` rather than `Func`?
```csharp
public delegate string Formatter(in LogValues state, Exception exception);
public static readonly Formatter Callback = (in LogValues state, Exception exception) => state.ToString();
```

src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
```diff
@@ -425,7 +425,7 @@ public static void Log(this ILogger logger, LogLevel logLevel, EventId eventId,
 
         //------------------------------------------HELPERS------------------------------------------//
 
-        private static string MessageFormatter(object state, Exception error)
+        private static string MessageFormatter(FormattedLogValues state, Exception error)
```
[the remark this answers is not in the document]
pakrym: `ILogger.Log` takes `Func<TState, Exception, string> formatter` so unfortunately, it won't help.

pakrym:
I'm looking into TwoArguments_DefineMessage_FilteredByLevel slowdown, seems like a side-effect of having to allocate stack space for structs.

src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs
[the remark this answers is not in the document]
benaadams: Shame, currently it will make a copy to stack (>8bytes) to get value semantics, then pass the pointer to that stack as the actual argument (due to [ABI](https://github.com/dotnet/coreclr/blob/master/Documentation/botr/clr-abi.md)), then it might make a second copy from that stack at the start of the called method (not sure about this bit)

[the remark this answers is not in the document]
pakrym: Wait, `Func<T>` is `<in T>`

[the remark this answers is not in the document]
pakrym: That's variance :(

pakrym:
Fixed regression, updated numbers.

AndyAyersMS:
Trying to parse what you wrote there ... if you new up a struct conditionally and that struct contains GC types you may pay costs in the method prolog to have the GC parts of the struct zeroed even if you never end up using the struct.

That is if S is a struct with GC type fields, then

```C#
if (oftenNotTrue())
{
     use(new S(...));
}
```
will pay some per-call cost for the potential `new S` even if `oftenNotTrue()` is indeed mostly false. As you note splitting off the `new S` so it happens in a child frame will avoid this.

This is one of those "currently" true things because the jit isn't very smart about structs passed to calls (even if there's a behind the scenes copy)....

```asm
G_M63142_IG01:
       sub      rsp, 56
       vzeroupper 
       xor      rax, rax                    // prolog cost (always paid)
       mov      qword ptr [rsp+28H], rax    //  ... cont
       mov      qword ptr [rsp+30H], rax    // nulling 2 gc fields

G_M63142_IG02:
       call     F:OftenNotTrue():bool
       test     eax, eax
       je       SHORT G_M63142_IG03
       lea      rdx, bword ptr [rsp+28H]    // "new" S cost (conditionally paid)
       vxorps   xmm0, xmm0                  // ... cont
       vmovdqu  qword ptr [rdx], xmm0       // .. .cont
       mov      rcx, 0xD1FFAB1E
       call     F:Use(byref)

G_M63142_IG03:
       nop      

G_M63142_IG04:
       add      rsp, 56
       ret      
```

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -608,25 +583,19 @@ IEnumerator IEnumerable.GetEnumerator()
             }
         }
 
-        private class LogValues<T0, T1, T2, T3, T4, T5> : IReadOnlyList<KeyValuePair<string, object>>
+        private readonly struct LogValues<T0, T1, T2, T3, T4, T5> : IReadOnlyList<KeyValuePair<string, object>>
         {
-            public static Func<object, Exception, string> Callback = (state, exception) => ((LogValues<T0, T1, T2, T3, T4, T5>)state)._formatter.Format(((LogValues<T0, T1, T2, T3, T4, T5>)state).ToArray());
+            public static readonly Func<LogValues<T0, T1, T2, T3, T4, T5>, Exception, string> Callback = (state, exception) => state._formatter.Format(state.ToArray());
```
davidfowl: Why not call _state.ToString() here?

src/Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs
```diff
@@ -470,23 +457,17 @@ IEnumerator IEnumerable.GetEnumerator()
             }
         }
 
-        private class LogValues<T0, T1, T2, T3> : IReadOnlyList<KeyValuePair<string, object>>
+        private readonly struct LogValues<T0, T1, T2, T3> : IReadOnlyList<KeyValuePair<string, object>>
         {
-            public static Func<object, Exception, string> Callback = (state, exception) => ((LogValues<T0, T1, T2, T3>)state)._formatter.Format(((LogValues<T0, T1, T2, T3>)state).ToArray());
+            public static readonly Func<LogValues<T0, T1, T2, T3>, Exception, string> Callback = (state, exception) => state._formatter.Format(state.ToArray());
```
davidfowl: Why not call state.ToString() here?

benchmarks/Logging.Performance/Logging.Performance.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
BrennanConroy: 2.2?

benchmarks/Logging.Performance/Logging.Performance.csproj
```diff
@@ -0,0 +1,21 @@
...
+    <RootNamespace>Microsoft.Extensions.DependencyInjection.Performance</RootNamespace>
```
BrennanConroy: Copy-pasta

benchmarks/Logging.Performance/LogValuesBenchmarks.cs
```diff
@@ -0,0 +1,52 @@
+﻿using System;
```
BrennanConroy: nit: copyright

benchmarks/Logging.Performance/LoggingBenchmarkBase.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+
+namespace Microsoft.Extensions.Logging.Performance
+{
+    public class LoggingBenchmarkBase
+    {
+        protected static readonly Action<ILogger, Exception> NoArgumentTraceMessage = LoggerMessage.Define(LogLevel.Trace, 0, "Message");
+        protected static readonly Action<ILogger, Exception> NoArgumentErrorMessage = LoggerMessage.Define(LogLevel.Error, 0, "Message");
+
+        protected static readonly Action<ILogger,  int, string, Exception> TwoArgumentTraceMessage = LoggerMessage.Define<int, string>(LogLevel.Trace, 0, "Message {Argument1} {Argument2}");
```
BrennanConroy: nit: extra spaces before `int`

build/dependencies.props
```diff
@@ -33,7 +35,7 @@
     <XunitAbstractionsPackageVersion>2.0.1</XunitAbstractionsPackageVersion>
     <XunitAssertPackageVersion>2.3.1</XunitAssertPackageVersion>
     <XunitExtensibilityExecutionPackageVersion>2.3.1</XunitExtensibilityExecutionPackageVersion>
-    <XunitPackageVersion>2.3.1</XunitPackageVersion>
+    <XunitPackageVersion>2.4.0</XunitPackageVersion>
```
BrennanConroy: Either revert this, or update it in `AspNetCore` and tell Eilon

Issue #918, opened 2018-10-29
LoggedTest and its siblings should have a consistent default log level


JunTaoLuo:
It's set to:
Information: https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Testing/LoggedTest/LoggedTestBase.cs#L41
Trace: https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Testing/LoggedTest/LoggedTestBase.cs#L68
Debug: https://github.com/aspnet/Logging/blob/master/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs#L61

We should just make these consistent and set the default to Debug everywhere.

BrennanConroy:
Done

PR #922, opened 2018-10-31
Default test log level to Debug

BrennanConroy:
We default to Debug in AssemblyTestLog https://github.com/aspnet/Logging/blob/2e30d867a412cd7b543f7ec0ed3932be0d08b8e3/src/Microsoft.Extensions.Logging.Testing/AssemblyTestLog.cs#L61 we should do so here too.


JunTaoLuo:
I've pointed it out here: https://github.com/aspnet/Logging/issues/918. Can you make it consistent everywhere?

JunTaoLuo (approved):

halter73:
I understand the LogLevels were inconsistent previously, but I don't see any explanation for why Debug was chosen as the new standardized default. Can we change the default back to Trace?

@BrennanConroy 	@JunTaoLuo

JunTaoLuo:
@tratcher was passionate about not setting it to Trace to avoid large log files, e.g. applicationhost.config files.

Issue #779, opened 2018-03-16
Consider adding support for wildcarding interior segments of log sources


rynowak:
Some discussion here: https://github.com/aspnet/HttpClientFactory/issues/51

The HttpClient factory creates log sources like:
```
Microsoft.Net.Http.HttpClient.foo.ClientHandler
Microsoft.Net.Http.HttpClient.foo.Polly
Microsoft.Net.Http.HttpClient.bar.ClientHandler
Microsoft.Net.Http.HttpClient.bar.Polly
```

The 5th segment represents the 'client name' - a logical identifier assigned by the application author. The 6th segment represents a specific source associated with that 'client name'.

Configuring the log level for a given client *by name* is the most primary scenario, so we like this design.... 

**BUT** 

It would be really handy to be able to configure.... say, the log level of Polly across all of your logical clients. That would require the ability to write a logging filter like:
```
Microsoft.Net.Http.HttpClient.*.Polly
```

pakrym:
Is supporting a single * enough or would you like to be able to have multiple wildcarding interior segments (e.g., `*.HttpClient.*.Polly`)?

I prefer supporting only a single wildcard because it would allow us to avoid writing complicated matching logic or using regexes.

rynowak:
Yeah I think there's a real rabbit hole of complexity here if you try to support everything that could be imagined. The scenario that I care about is the one in the original post.

I'd also be OK with us closing this, no one has actually asked for this, I just thought it would be nice.

PR #924, opened 2018-10-31
Add support for category wildcards

pakrym:
Only a single wildcard character is supported per category name.

Fixes: https://github.com/aspnet/Logging/issues/779

### DIFF
--- a/src/Microsoft.Extensions.Logging/LoggerRuleSelector.cs
+++ b/src/Microsoft.Extensions.Logging/LoggerRuleSelector.cs
@@ -7,6 +7,8 @@ namespace Microsoft.Extensions.Logging
 {
     internal class LoggerRuleSelector
     {
+        private static readonly char[] WildcardChar = { '*' };
+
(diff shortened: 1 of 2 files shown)


src/Microsoft.Extensions.Logging/LoggerRuleSelector.cs
```diff
@@ -47,9 +49,22 @@ private static bool IsBetter(LoggerFilterRule rule, LoggerFilterRule current, st
...
             {
-                return false;
+                var categoryParts = rule.CategoryName.Split(WildcardChar);
+                if (categoryParts.Length > 2)
+                {
+                    throw new InvalidOperationException("Only one wildcard character is allowed in category name.");
+                }
+
+                var prefix = categoryParts[0];
+                var suffix = categoryParts.Length > 1 ? categoryParts[1] : string.Empty;
```
rynowak: Should we enforce the invariant that a wildcard needs to replace whole segments?

Right now this supports things like `Micro*` or `*soft` which just doesn't make sense.

PR #925, opened 2018-10-31
Update Roslyn version

rynowak:
This was failing to auto-update because of the PrivateAssets and the
conflicts with the shared framework which has 2.8.0 inside.


rynowak:
Merging this to unblock the build.

PR #926, opened 2018-11-01
Adjust Roslyn package versions

rynowak:


rynowak:
Merging to unblock the build

Issue #818, opened 2018-04-27
Add support for logging scope information in Azure App Service Logger


anurse:
Currently, scope data is not logged when an event is logged via the App Service Logger. We should include scope data so that Request IDs and SignalR Connection IDs are available in the log messages

PR #928, opened 2018-11-01
Add scopes to azure loggers

pakrym:
Fixes: https://github.com/aspnet/Logging/issues/818

### DIFF
--- a/src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
+++ b/src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
@@ -40,7 +40,22 @@ namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
             builder.Append(logLevel.ToString());
             builder.Append("] ");
             builder.Append(_category);
-            builder.Append(": ");
+
+            var scopeProvider = _provider.ScopeProvider;
+            if (scopeProvider != null)
+            {
+                scopeProvider.ForEachScope((scope, stringBuilder) =>
+                {
+                    builder.Append(" => ").Append(scope);
+                }, builder);
+
+                builder.AppendLine(":");
+            }
+            else
+            {
+                builder.Append(": ");
+            }
+
             builder.AppendLine(formatter(state, exception));
 
             if (exception != null)
--- a/src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
+++ b/src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Extensions.Logging.AzureAppServices.Internal
             {
                 scopeProvider.ForEachScope((scope, stringBuilder) =>
                 {
-                    builder.Append(" => ").Append(scope);
+                    stringBuilder.Append(" => ").Append(scope);
                 }, builder);
 
                 builder.AppendLine(":");
--- a/src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationExtensions.cs
+++ b/src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationExtensions.cs
@@ -1,0 +1,24 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging.Configuration;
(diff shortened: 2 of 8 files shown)


src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
```diff
@@ -40,7 +40,22 @@ public void Log<TState>(DateTimeOffset timestamp, LogLevel logLevel, EventId eve
...
+
+            var scopeProvider = _provider.ScopeProvider;
+            if (scopeProvider != null)
+            {
+                scopeProvider.ForEachScope((scope, stringBuilder) =>
+                {
+                    builder.Append(" => ").Append(scope);
+                }, builder);
+
+                builder.AppendLine(":");
```
BrennanConroy: If we `AppendLine` here should we also add a few spaces like we do in `ConsoleLogger`?

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
```diff
@@ -40,7 +40,22 @@ public void Log<TState>(DateTimeOffset timestamp, LogLevel logLevel, EventId eve
...
             builder.Append("] ");
             builder.Append(_category);
-            builder.Append(": ");
+
+            var scopeProvider = _provider.ScopeProvider;
+            if (scopeProvider != null)
+            {
+                scopeProvider.ForEachScope((scope, stringBuilder) =>
+                {
+                    builder.Append(" => ").Append(scope);
```
BrennanConroy: Use `stringBuilder` to avoid the capture

src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationExtensions.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.Extensions.DependencyInjection;
```
BrennanConroy: nit: copyright

src/Microsoft.Extensions.Logging.Configuration/LoggerProviderConfigurationExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Logging.Configuration;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Extensions.Logging.Configuration
+{
+    /// <summary>
+    /// Provides set of helpers to initialize options objects from logger provider configuration.
```
BrennanConroy: "Provides **a** set of helpers"

test/Microsoft.Extensions.Logging.AzureAppServices.Test/BatchingLoggerProviderTests.cs
```diff
@@ -13,8 +14,9 @@ namespace Microsoft.Extensions.Logging.AzureAppServices.Test
 {
     public class BatchingLoggerProviderTests
     {
-        DateTimeOffset _timestampOne = new DateTimeOffset(2016, 05, 04, 03, 02, 01, TimeSpan.Zero);
-        string _nl = Environment.NewLine;
+        private DateTimeOffset _timestampOne = new DateTimeOffset(2016, 05, 04, 03, 02, 01, TimeSpan.Zero);
+        private string _nl = Environment.NewLine;
+        private Regex _timeStampRegex = new Regex(@"^\d\d\d\d-\d\d-\d\d \d\d:\d\d:\d\d.\d\d\d .\d\d:\d\d ");
```
BrennanConroy: super nit: this regex can be a bit cleaner, i.e. `^\d{4}-\d{2}-\d{2}...`

BrennanConroy: nit: add a timeout to the regex just in case

src/Microsoft.Extensions.Logging.AzureAppServices/Internal/BatchingLogger.cs
pakrym: It didn't append spaces in case of multiline messages before so I'm reluctant to add it here.

BrennanConroy: Fair enough

PR #929, opened 2018-11-01
Optimize BeginScope performance

pakrym:
By unloading more logic from the hot path to logger creation time.

Fixes: https://github.com/aspnet/Logging/issues/894
Fixes: https://github.com/aspnet/Logging/issues/714

Also `LoggerFactory` is much more readable now.

Before:
```

               Method | CaptureScopes |     Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
--------------------- |-------------- |---------:|---------:|---------:|------------:|-------:|----------:|
 MultipleMixedLoggers |         False | 185.4 ns | 1.233 ns | 1.093 ns | 5,393,087.5 |      - |       0 B |
 MultipleMixedLoggers |          True | 310.3 ns | 3.418 ns | 2.854 ns | 3,222,362.8 | 0.0005 |     168 B |
              
```
After:

```
 Method | CaptureScopes |     Mean |    Error |   StdDev |        Op/s |  Gen 0 | Allocated |
--------------------- |-------------- |---------:|---------:|---------:|------------:|-------:|----------:|
 MultipleMixedLoggers |         False | 173.6 ns | 2.438 ns | 2.281 ns | 5,761,550.5 |      - |       0 B |
 MultipleMixedLoggers |          True | 280.3 ns | 5.552 ns | 6.818 ns | 3,567,640.0 | 0.0005 |     168 B |
```

### DIFF
--- a/src/Microsoft.Extensions.Logging/Logger.cs
+++ b/src/Microsoft.Extensions.Logging/Logger.cs
@@ -9,40 +9,13 @@ namespace Microsoft.Extensions.Logging
 {
     internal class Logger : ILogger
     {
-        private readonly LoggerFactory _loggerFactory;
-
-        private LoggerInformation[] _loggers;
-
-        private int _scopeCount;
-
-        public Logger(LoggerFactory loggerFactory)
-        {
-            _loggerFactory = loggerFactory;
-        }
-
-        public LoggerInformation[] Loggers
-        {
-            get { return _loggers; }
-            set
-            {
-                var scopeSize = 0;
-                foreach (var loggerInformation in value)
-                {
-                    if (loggerInformation.CreateScopes)
-                    {
-                        scopeSize++;
-                    }
-                }
-                _scopeCount = scopeSize;
-                _loggers = value;
-            }
-        }
-
-        public bool CaptureScopes { get; set; }
+        public LoggerInformation[] Loggers { get;set; }
+        public MessageLogger[] MessageLoggers { get; set; }
+        public ScopeLogger[] ScopeLoggers { get; set; }
 
         public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
         {
-            var loggers = Loggers;
+            var loggers = MessageLoggers;
             if (loggers == null)
             {
                 return;
@@ -80,7 +53,7 @@ namespace Microsoft.Extensions.Logging
 
         public bool IsEnabled(LogLevel logLevel)
         {
-            var loggers = Loggers;
+            var loggers = MessageLoggers;
             if (loggers == null)
             {
                 return false;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -97,18 +92,17 @@ public void AddProvider(ILoggerProvider provider)
             {
                 AddProviderRegistration(provider, dispose: true);
 
-                foreach (var logger in _loggers)
+                foreach (var existingLogger in _loggers)
                 {
-                    var loggerInformation = logger.Value.Loggers;
-                    var categoryName = logger.Key;
+                    var logger = existingLogger.Value;
+                    var loggerInformation = existingLogger.Value.Loggers;
```
BrennanConroy:
```suggestion
                    var loggerInformation = logger.Loggers;
```

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -123,50 +117,70 @@ private void AddProviderRegistration(ILoggerProvider provider, bool dispose)
...
         }
 
-        private void ApplyRules(LoggerInformation[] loggers, string categoryName, int start, int count)
+        private (MessageLogger[] MessageLoggers, ScopeLogger[] ScopeLoggers) ApplyFilters(LoggerInformation[] loggers)
         {
-            for (var index = start; index < start + count; index++)
-            {
-                ref var loggerInformation = ref loggers[index];
+            var messageLoggers = new List<MessageLogger>();
+            var scopeLoggers = _filterOptions.CaptureScopes? new List<ScopeLogger>() : null;
```
BrennanConroy: nit: space after `CaptureScopes`

src/Microsoft.Extensions.Logging/LoggerInformation.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Extensions.Logging
 {
-    internal struct LoggerInformation
+    internal struct MessageLogger
```
BrennanConroy: Can these be `readonly`?

src/Microsoft.Extensions.Logging/LoggerFactory.cs
pakrym:
```suggestion
            var scopeLoggers = _filterOptions.CaptureScopes ? new List<ScopeLogger>() : null;
```

src/Microsoft.Extensions.Logging/Logger.cs
```diff
@@ -9,40 +9,13 @@ namespace Microsoft.Extensions.Logging
...
-                        scopeSize++;
-                    }
-                }
-                _scopeCount = scopeSize;
-                _loggers = value;
-            }
-        }
-
-        public bool CaptureScopes { get; set; }
+        public LoggerInformation[] Loggers { get;set; }
```
BrennanConroy: nit: space after `;`

BrennanConroy: Is there a good reason these are array instead of `List<...>`? We could save calling `.ToArray()` if they were `List<...>`

pakrym: Most of the time (when ILoggerFactory,AddProvider is no used) this would get set only once on startup, seemed wasteful to keep the list around.

src/Microsoft.Extensions.Logging/LoggerFactory.cs
```diff
@@ -50,15 +49,9 @@ private void RefreshFilters(LoggerFilterOptions filterOptions)
             lock (_sync)
             {
                 _filterOptions = filterOptions;
-                foreach (var logger in _loggers)
+                foreach (var logger in _loggers.Values)
```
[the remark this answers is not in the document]
pakrym: Doesn't matter much, this is executed only on configuration reloads.

